PR #1, opened 2018-10-01
Fork ripple plugin into edge-currency-accountbased

paullinator:
Port over Ripple plugin and add XLM and EOS support

Major refactor to pull common code into `./common/` and separate currency specific routines into their own directory

XRP working as before from edge-currency-ripple repo.
XLM working completely but using both stellar.org server and edge.app server
EOS support working but only able to show outgoing transactions and unable to activate new account.


src/stellar/stellarEngine.js
```diff
@@ -0,0 +1,570 @@
...
+  // synchronous
+  async makeSpend (edgeSpendInfo: EdgeSpendInfo) {
+    // Validate the spendInfo
+    const valid = validateObject(edgeSpendInfo, MakeSpendSchema)
+
+    if (!valid) {
+      throw new Error('Error: invalid EdgeSpendInfo')
+    }
+
+    if (edgeSpendInfo.spendTargets.length !== 1) {
```
paullinator: make validation common

PR #5, opened 2018-11-11
Paul/lazy load transactions

paullinator:


paullinator:
Approved by @thehobbit85

PR #8, opened 2019-01-08
Paul/eos fixes3

paullinator:


swansontec:
Waiiiitttt... how does our app have EOS if this isn't merged? [...checks edge-react-gui...] Wow! We shipped with a git branch for this particular plugin. Yeah, we need to get this in ASAP.

src/common/engine.js
```diff
@@ -279,7 +286,7 @@ class CurrencyEngine {
         if (!this.txIdMap[currencyCode]) {
           this.txIdMap[currencyCode] = {}
         }
-        this.txIdMap[currencyCode][tx.txid] = i
+        this.txIdMap[currencyCode][normalizeAddress(tx.txid)] = i
```
swansontec: This isn't an address. Maybe the helper should be called 'normalizeHex' instead.

src/common/engine.js
```diff
@@ -697,7 +729,39 @@ class CurrencyEngine {
...
+      currencyCode = edgeSpendInfo.currencyCode
+    } else {
+      currencyCode = this.currencyInfo.currencyCode
+    }
+    const nativeBalance = this.walletLocalData.totalBalances[currencyCode]
+    if (!nativeBalance || bns.eq(nativeBalance, '0')) {
+      throw new error.InsufficientFundsError()
+    }
+
+    edgeSpendInfo.currencyCode = currencyCode
```
swansontec: Bold move, mutating an object that doesn't belong to you. I think it's OK with the way the core and bridge work, though, so no need to fix it. If it were another data structure I might worry more about it.

src/stellar/stellarEngine.js
```diff
@@ -475,8 +452,8 @@ export class StellarEngine extends CurrencyEngine {
 
     const networkFee = transaction.fee.toString()
     nativeAmount = bns.add(networkFee, nativeAmount) // Add fee to total
-    nativeBalance = bns.sub(nativeBalance, '10000000') // Subtract the 1 min XLM
-    if (bns.gt(nativeAmount, nativeBalance)) {
+    const nativeBalance2 = bns.sub(nativeBalance, '10000000') // Subtract the 1 min XLM
```
swansontec: Maybe `realNativeBalance` would have been a better name, but I get what this is doing.

src/stellar/stellarEngine.js
```diff
@@ -242,7 +242,16 @@ export class StellarEngine extends CurrencyEngine {
           pagingToken = await this.processTransaction(tx)
         }
       } catch (e) {
-        pagingToken = undefined
+        if (e.response && e.response.title === 'Resource Missing') {
+          this.log('Account not found. Probably not activated w/minimum XLM')
```
swansontec: Only QA / support sees these logs. I presume this is for their benefit. Does the core need a way to surface this through to the GUI as a proper error / plugin status flag?

src/eos/eosEngine.js
```diff
@@ -713,17 +713,35 @@ export class EosEngine extends CurrencyEngine {
...
+      try {
+        err = JSON.parse(e)
+      } catch (e2) {
+        throw e
       }
-    )
-    edgeTransaction.date = Date.now() / 1000
-    edgeTransaction.txid = signedTx.transaction_id
-    return edgeTransaction
+      if (err.error && err.error.name === 'tx_net_usage_exceeded') {
```
swansontec: Do we want to make these errors official in the Core? They obviously work fine like this for now.

src/eos/eosEngine.js
```diff
@@ -12,74 +14,99 @@ import type {
...
-  transactionsChecked: number
+  eosPlugin: EosPlugin
   activatedAccountsCache: { [publicAddress: string]: boolean }
   otherData: EosWalletOtherData
+  otherMethods: Object
 
   constructor (
-    currencyPlugin: CurrencyPlugin,
+    currencyPlugin: EosPlugin,
     io_: any,
```
swansontec: Should use `EdgeIo` here.

src/eos/eosEngine.js
```diff
@@ -94,22 +121,82 @@ export class EosEngine extends CurrencyEngine {
       }
     } catch (e) {
       this.log(`Error fetching height: ${JSON.stringify(e)}`)
```
swansontec: You can't use `JSON.stringify` on errors, since that just prints "{}" no matter what the error. You want to use `String(e)` instead, which gives you a nice description like "FooError: some message"

src/eos/eosEngine.js
```diff
@@ -420,33 +512,13 @@ export class EosEngine extends CurrencyEngine {
   }
 
   // synchronous
```
swansontec: We should ditch these incorrect synchronous / asynchronous comments, now that the `async` keyword exits to convey the same info in a computer-enforced manner.

paullinator:
Thanks. I've made a task to address these smaller issues after merge

PR #9, opened 2019-01-20
Add Ethereum

paullinator:
This is built on top of PR https://github.com/EdgeApp/edge-currency-accountbased/pull/8 which adds EOS functionality. Merge that PR first.

### DIFF
--- a/src/common/utils.js
+++ b/src/common/utils.js
@@ -4,8 +4,9 @@
  */
 
 import { validate } from 'jsonschema'
-import { type EdgeCurrencyInfo } from 'edge-core-js'
+import { type EdgeCurrencyInfo, type EdgeMetaToken } from 'edge-core-js'
 import { BN } from 'bn.js'
+import { bns } from 'biggystring'
 const Buffer = require('buffer/').Buffer
 
 function normalizeAddress (address: string) {
@@ -48,6 +49,10 @@ export function isHex (h: string) {
   return out
 }
 
+export function toHex (num: string) {
+  return bns.add(num, '0', 16)
+}
+
 export function hexToBuf (hex: string) {
   const noHexPrefix = hex.replace('0x', '')
   const noHexPrefixBN = new BN(noHexPrefix, 16)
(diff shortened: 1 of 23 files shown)


swansontec:
This still has the commits from the EOS PR mixed in for some reason. Could you rebase it on top of master for review purposes?

src/ethereum/ethEngine.js
```diff
@@ -0,0 +1,947 @@
...
+  decrementNonce: boolean
+}
+
+export class EthereumEngine extends CurrencyEngine {
+  ethereumPlugin: EthereumPlugin
+  otherData: EthereumWalletOtherData
+
+  constructor (
+    currencyPlugin: EthereumPlugin,
+    io_: any,
```
swansontec: `EdgeIo` type here.

src/ethereum/ethEngine.js
```diff
@@ -0,0 +1,947 @@
...
+      if (walletInfo.keys.keys && walletInfo.keys.keys.ethereumKey) {
+        this.walletInfo.keys.ethereumKey = walletInfo.keys.keys.ethereumKey
+      }
+    }
+    this.currencyPlugin = currencyPlugin
+  }
+
+  async fetchGetEtherscan (server: string, cmd: string) {
+    let apiKey = ''
+    if (global.etherscanApiKey && global.etherscanApiKey.length > 5) {
```
swansontec: Note to self: s/global/initOptions/ in the upcoming refactor.

src/ethereum/ethEngine.js
```diff
@@ -0,0 +1,947 @@
...
+    const url = `${server}/api${cmd}${apiKey}`
+    return this.fetchGet(url)
+  }
+
+  async fetchGet (url: string) {
+    const response = await this.io.fetch(url, {
+      method: 'GET'
+    })
+    if (!response.ok) {
+      const cleanUrl = url.replace(global.etherscanApiKey, 'private')
```
swansontec: Do we need to sanitize the blockcypher API key here too?

src/ethereum/ethEngine.js
```diff
@@ -0,0 +1,947 @@
...
+    this.addToLoop('checkTransactionsInnerLoop', TRANSACTION_POLL_MILLISECONDS)
+    this.addToLoop('checkUnconfirmedTransactionsInnerLoop', UNCONFIRMED_TRANSACTION_POLL_MILLISECONDS)
+    super.startEngine()
+  }
+
+  async killEngine () {
+    // Set status flag to false
+    this.engineOn = false
+    // Clear Inner loops timers
+    for (const timer in this.timers) {
```
swansontec: The logic for killing timers seems pretty similar between plugins. Might make sense to put in the base class.

src/ethereum/ethEngine.js
```diff
@@ -0,0 +1,947 @@
...
+    // this.leavePool()
+  }
+
+  async resyncBlockchain (): Promise<void> {
+    await this.killEngine()
+    await this.clearBlockchainCache()
+    await this.startEngine()
+  }
+
+  // synchronous
```
swansontec: The comment lies. It's async.

src/common/utils.js
```diff
@@ -12,6 +12,14 @@ function normalizeAddress (address: string) {
   return address.toLowerCase().replace('0x', '')
 }
 
+function addHexPrefix (value: string) {
+  if (value.startsWith('0x')) {
```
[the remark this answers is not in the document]
paullinator: Updated to use indexOf

src/ethereum/ethEngine.js
paullinator: Good point. We were never doing that even in the original ethereum plugin.

src/ethereum/ethPlugin.js
```diff
@@ -0,0 +1,177 @@
...
+    if (!currencyEngine.otherData.recommendedFee) {
+      currencyEngine.otherData.recommendedFee = '0'
+    }
+
+    const out: EdgeCurrencyEngine = currencyEngine
+    return out
+  }
+
+  async parseUri (uri: string): Promise<EdgeParsedUri> {
+    const networks = { ethereum: true, ether: true }
```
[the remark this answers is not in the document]
paullinator: Hmm. are you referring to `parseUriCommon()`. That it does use.

src/ethereum/ethEngine.js
paullinator: Good catch. moved killEngine to the base class

src/ethereum/ethEngine.js
paullinator: Removed all these obsolete comments

PR #16, opened 2019-02-08
Fix tx amount for XRP

paullinator:
Make sure fee is included in nativeAmount for outgoing transactions.


src/xrp/xrpEngine.js
```diff
@@ -144,10 +144,8 @@ export class XrpEngine extends CurrencyEngine {
         const date: number = Date.parse(tx.outcome.timestamp) / 1000
         const blockHeight: number = tx.outcome.ledgerVersion
```
paullinator: In the below code, `bc.value` is negative if this transaction is a spend. tx.outcome.fee is always positive. Adding the two together effectively removed the fee from the total amount spent which is NOT what we want. Removing line 149 effectively keeps exchangeAmount as the total of spend + fee.

PR #28, opened 2019-03-04
Multicast ether APIs

paullinator:
Utilize multiple ether APIs for blockchain queries

Fixes stuck loading of ether wallets due to rate limiting from APIs. This PR adds use of Blockscout and Infura in addition to etherscan

### DIFF
--- a/src/ethereum/ethEngine.js
+++ b/src/ethereum/ethEngine.js
@@ -59,7 +59,7 @@ const ADDRESS_QUERY_LOOKBACK_BLOCKS = 4 * 60 * 24 * 7 // ~ one week
 const NUM_TRANSACTIONS_TO_QUERY = 50
 const WEI_MULTIPLIER = 100000000
 
-type EthFunction = 'broadcastTx' | 'eth_blockNumber' | 'eth_getTransactionCount'
+type EthFunction = 'broadcastTx' | 'eth_blockNumber' | 'eth_getTransactionCount' | 'eth_getBalance' | 'getTokenBalance'
 
 type BroadcastResults = {
   incrementNonce: boolean,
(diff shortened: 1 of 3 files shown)


src/ethereum/ethEngine.js
```diff
@@ -96,7 +104,11 @@ export class EthereumEngine extends CurrencyEngine {
   async fetchGetEtherscan (server: string, cmd: string) {
     const { etherscanApiKey } = this.initOptions
     let apiKey = ''
-    if (etherscanApiKey && etherscanApiKey.length > 5) {
+    if (
+      etherscanApiKey &&
+      etherscanApiKey.length > 5 &&
+      server.includes('etherscan')
```
swansontec: If this function is used for more than just Etherscan, should it have a different name?

src/ethereum/ethEngine.js
```diff
@@ -166,41 +175,64 @@ export class EthereumEngine extends CurrencyEngine {
     }
   }
 
-  async checkAccountFetch (tk: string, url: string) {
+  updateBalance (tk: string, balance: string) {
+    if (typeof this.walletLocalData.totalBalances[tk] === 'undefined') {
```
swansontec: Please consider `x == null` instead of `typeof x === 'undefined'`. It's the same thing, but a lot shorter. If you really mean exactly "`undefined` and not `null`", you can just say that, like `x === undefined`.

The `typeof` thing is only useful in exactly one case: When you have a top-level variable like `window` that may or may not even exist at all, then you really do need `typeof`, since touching a non-existing variable any other way is an error.

src/ethereum/ethEngine.js
```diff
@@ -591,6 +625,7 @@ export class EthereumEngine extends CurrencyEngine {
 
   async multicastServers (func: EthFunction, ...params: any): Promise<any> {
     let out = { result: '', server: 'no server' }
+    let funcs, funcs2, url
```
swansontec: These seem to be `const` within each respective case statement. You can use curly braces within case statements to declare local variables and make this clearer:

```js
switch (x) {
  case 1: {
    const local = stuff
    // ...
    break
  }
  case 2: {
    const local = otherStuff
    // ...
    break
  }
}

src/ethereum/ethEngine.js
swansontec: Ah, I get it. It's the Etherscan API, but not necessarily hosted by Etherscan the company. Seems fine, then.

src/ethereum/ethEngine.js
paullinator: This was just moving old existing code that admittedly could have just been
`    if (this.walletLocalData.totalBalances[tk]) {`
 I'll make note for future refactors

PR #29, opened 2019-03-06
Convert Infura eth_getBalance to decimal string

paullinator:
Valid return value of getBalance should be a decimal string. Unlike etherscan, infura returns hex which was causing a crash in GUI. Validate that the return value is hex then convert it using biggystring.

### DIFF
--- a/src/ethereum/ethEngine.js
+++ b/src/ethereum/ethEngine.js
@@ -25,7 +25,8 @@ import {
   promiseAny,
   shuffleArray,
   toHex,
-  validateObject
+  validateObject,
+  isHex
 } from '../common/utils.js'
 import { currencyInfo } from './ethInfo.js'
 import { calcMiningFee } from './ethMiningFees.js'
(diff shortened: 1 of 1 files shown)


src/ethereum/ethEngine.js
```diff
@@ -722,6 +723,12 @@ export class EthereumEngine extends CurrencyEngine {
             params[0],
             'latest'
           ])
+          // Convert hex
+          if (!isHex(result.result)) {
+            throw new Error('Infura eth_getBalance not hex')
```
swansontec: Since `bns.add` apparently works fine with both decimal and hex strings, is the check really needed?

paullinator: My concern is if we don’t get a hex string, then the API is broken or we have a malicious or broken server. It’s not much of a check but at least we won’t barf on a value like “123.456” which is a valid bns value but not valid hex.

PR #31, opened 2019-03-09
Check if makeSpend is from Totle plugin and set data if exists.

passabilities:


paullinator:
This looks very hacky. There should be no code in the currency plugins that is specific to an exchange provider. What’s the purpose of this PR?

passabilities:
@paullinator I agree this isn't the best implementation but without knowing the code base very well I was not sure of the best way to add our plugin without breaking support for others. The main goal is that if a token is the source of the swap then we (as in **Totle**) need the `to` address to be our contract, not the token's contract.

passabilities:
@swansontec The only thing would be I guess if `data` exists, then we would need the `publicAddress` to be used as the `to` in the transaction. Also note that the transaction will fail at this time as we need to submit another transaction before hand to set an allowance before a user sells a token. We are working on this now and will keep you updated when it's completed.

swansontec:
I've been playing with the Totle plugin over the past week, and another thing we need to fix is the gas price. The Ethereum plugin currently takes the gas price in GWei, not Wei like it ought to. We would need to fix the Edge GUI as well to make this transition, so maybe the Totle plugin can just divide by a billion for the time being.

I also noticed the need for the second token approval transaction, yes.

kylanhurt:
PR superceded by: https://github.com/EdgeApp/edge-currency-accountbased/pull/45

PR #37, opened 2019-04-17
Fix Issue with Token Denominations for QR Code Parsing and Encoding

kylanhurt:
The purpose of this task is to get the Ethereum plugin to properly determine denomination info for customTokens, specifically when working with encoded and parsed URI's.

Asana Task: https://app.asana.com/0/361770107085503/1118536468069439/f

### DIFF
--- a/src/ethereum/ethPlugin.js
+++ b/src/ethereum/ethPlugin.js
@@ -13,7 +13,9 @@ import {
   type EdgeCurrencyPlugin,
   type EdgeEncodeUri,
   type EdgeIo,
+  type EdgeMetaToken,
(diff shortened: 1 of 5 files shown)


src/ethereum/ethPlugin.js
```diff
@@ -164,6 +171,7 @@ export class EthereumPlugin extends CurrencyPlugin {
           contractAddress,
           currencyName,
           multiplier,
+          denominations: [{ name: currencyCode, multiplier }],
```
swansontec: Does the GUI need denominations? I don't think it makes sense to return this, now that I'm thinking about it more.

PR #39, opened 2019-05-15
Guillaume/unconfirmed nonce

westerly:
Re created @paullinator PR so that it is possible to spend more than one unconfirmed tx with the ETH plugin.

### DIFF
--- a/src/common/engine.js
+++ b/src/common/engine.js
@@ -121,6 +121,29 @@ class CurrencyEngine {
     )
   }
 
+  isSpendTx (edgeTransaction: EdgeTransaction): boolean {
+    if (edgeTransaction.nativeAmount) {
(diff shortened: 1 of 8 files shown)


src/common/engine.js
```diff
@@ -121,6 +121,29 @@ class CurrencyEngine {
...
+  isSpendTx (edgeTransaction: EdgeTransaction): boolean {
+    if (edgeTransaction.nativeAmount) {
+      if (edgeTransaction.nativeAmount.slice(0, 1) === '-') {
+        return true
+      }
+      if (bns.gt(edgeTransaction.nativeAmount, '0')) {
+        return false
+      }
+    }
+    let out = true
```
swansontec: The `out` variable isn't really needed if you put `return false` inside the loop, and then `return true` at the end of the method. Still, no biggie - it obviously works fine.

PR #44, opened 2019-07-07
Prevent empty strings to be a valid eth / token balance

westerly:
* Prevent empty strings to be a valid eth / token balance.


westerly:
@swansontec do you think that can be merged?

PR #47, opened 2019-07-12
Fix missing otherParams.data bug

kylanhurt:
Put in check for otherParams before attempting to assign to data property (prevents 'property of undefined' error)


kylanhurt:
Verbal approval from @swansontec

PR #50, opened 2019-07-23
FIO Plug-in added for FIO key generation

shawnarneygit:
The edge-currency-accountbased plug-in has been modified to add 'fio' as a plug in.  
The methods for the 'fio' plug-in, have been modifed to support FIO private and public key generation.

The edge-currency-accountbased plug-in, now has a dependency on the FIO typescript SDK.  
Which has a dependency on the FIOJS SDK.  These dependencies are referenced via npmjs.com.

The edge-currency-accountbased plug-in, is referencing the 'fiosdk_typescript' SDK in the 'package.json' file here:
"@dapix/react-native-fio": "0.1.0"


swansontec:
I went ahead and resolved the merge conflicts, so your code is now on this branch: https://github.com/EdgeApp/edge-currency-accountbased/tree/dapix/fio

Reviewing now.

src/fio/fioPlugin.js
```diff
@@ -0,0 +1,217 @@
...
+      if (!denom) {
+        throw new Error('InternalErrorInvalidCurrencyCode')
+      }
+      amount = bns.div(nativeAmount, denom.multiplier, 16)
+    }
+    const encodedUri = this.encodeUriCommon(obj, 'fio', amount)
+    return encodedUri
+  }
+
+  async getAccSystemStats (account: string) {
```
swansontec: This looks like leftover EOS code. The `eosServer` is never set or accessed anywhere else.

src/fio/fioPlugin.js
```diff
@@ -0,0 +1,217 @@
...
+
+export function checkAddress (address: string): boolean {
+  const start = address.startsWith('FIO')
+  const lenght = address.length === 53
+  return start && lenght
+}
+
+export class FioPlugin extends CurrencyPlugin {
+  otherMethods: Object
+  eosServer: Object
```
swansontec: Leftover EOS code?

src/fio/fioPlugin.js
```diff
@@ -0,0 +1,217 @@
...
+
+    currencyEngine.otherData = currencyEngine.walletLocalData.otherData
+    // currencyEngine.otherData is an opaque utility object for use for currency
+    // specific data that will be persisted to disk on this one device.
+    // Commonly stored data would be last queried block height or nonce values for accounts
+    // Edit the flow type EosWalletOtherData and initialize those values here if they are
+    // undefined
+    // TODO: Initialize anything specific to this currency
+    // if (!currencyEngine.otherData.nonce) currencyEngine.otherData.nonce = 0
+    if (!currencyEngine.otherData.accountName) {
```
swansontec: Since these are properties of the currency engine, it seems like they should be set up in the currency engine constructor. Assuming they are even needed (they look like more EOS leftovers)?

src/fio/fioPlugin.js
```diff
@@ -0,0 +1,217 @@
...
+
+  constructor (io: EdgeIo) {
+    super(io, 'fio', currencyInfo)
+  }
+
+  async createPrivateKey (walletType: string): Promise<Object> {
+    const type = walletType.replace('wallet:', '')
+    if (type === 'fio') {
+      const buffer = this.io.random(32)
+      return FIOSDK.createPrivateKey(buffer)
```
swansontec: I see that the format is:

```typescript
{
  fioKey: string // base58?,
  mnemonic: string
}
```

If the `fioKey` is missing, will the FIO SDK still work correctly?

shawnarneygit: For this, I don't know the implications of this. Or how the fioKey may be missing in this scenario.

FIOSDK.createPrivateKey(buffer) does take a buffer to generate the private key.  And the FIO SDK as is, is just calling static methods.

swansontec:
See #62 as the replacement.

PR #75, opened 2019-09-12
Implement import key and mnemonic for RSK

kylanhurt:
Implement import key and mnemonic (RSK)

Remove unnecessary comments


swansontec:
Instead of leaving yet another review, I decided to just make a commit with what I would change. There shouldn't be any logic changes at this point, just cleaning up inconsistencies  & redundant logic.

PR #79, opened 2019-09-17
Allow Tezos Mnemonic Import

kylanhurt:


src/tezos/tezosPlugin.js
```diff
@@ -43,6 +43,23 @@ export class TezosPlugin extends CurrencyPlugin {
...
+        type: 'wallet:tezos',
+        id: 'fake',
+        keys
+      })
+      return {
+        mnemonic: keys.mnemonic,
+        privateKey: keys.sk
+      }
+    } catch (e) {
+      throw new Error('Invalid input')
```
swansontec: `catch () { throw ... }` is completely useless - it throws away the *actual* error and replaces it with something that has less information. Just let the original eztz method throw whatever it wants.

src/tezos/tezosPlugin.js
```diff
@@ -56,7 +65,7 @@ export class TezosPlugin extends CurrencyPlugin {
         privateKey: keys.sk
       }
     } catch (e) {
-      throw new Error('Invalid input')
+      throw new Error(e)
```
swansontec: This code is wrong. `throw new Error(e)` doesn't even make sense - `e` is an `Error` object, not a string. Just get rid of the entire try/catch block.

src/tezos/tezosPlugin.js
```diff
@@ -76,8 +85,12 @@ export class TezosPlugin extends CurrencyPlugin {
   async derivePublicKey (walletInfo: EdgeWalletInfo): Promise<Object> {
     const type = walletInfo.type.replace('wallet:', '')
     if (type === 'tezos') {
-      const keypair = eztz.crypto.generateKeys(walletInfo.keys.mnemonic, '')
-      return { publicKey: keypair.pkh, publicKeyEd: keypair.pk }
+      try {
+        const keypair = eztz.crypto.generateKeys(walletInfo.keys.mnemonic, '')
```
swansontec: I don't think we need a try/catch block here either. If something goes wrong, eztz will presumably throw its own error with its own useful message.

PR #80, opened 2019-10-01
implement Hedera support

abonander:
This implements currency support for HBAR, the currency of Hedera Hashgraph: https://www.hedera.com/

I will be opening a companion PR in `edge-react-gui`.

Hedera is a bit different from other cryptos which is probably going to need some adjustments in edge-react-gui for a nice UX and maybe an additional API in `EdgeIo`:

* Hedera doesn't use public keys to address wallets; instead, it uses an account ID which is represented as a 3-tuple of unsigned ints `{ shard, realm, account }` with a de-facto string format of `shard.realm.account`. To accommodate this in `HederaPlugin.importPrivateKey()` I expect the string to be the stringified account ID and the private key string concatenated together (currently with `:` as the separator).
* Hedera accounts cost crypto to create and maintain; exchanges are intended to create accounts for users who buy HBAR through them, and Hedera has also created a number of accounts for initial users itself. Thus the current "Create Wallet" flow isn't really suitable but there's no way to disable it in `edge-react-gui`. This is why `HederaPlugin.createPrivateKey()` throws an exception. (We can generate private keys for the user but it doesn't make sense to show them a wallet until their account has been created on the network.)
* Transaction signing is implemented with ed25519 which is not provided in `EdgeIo` currently; the SDK itself uses `tweetnacl` internally.
* Transactions are not broadcast to the network but submitted to a single node over gRPC; in the SDK we support `grpc-web` which uses global `fetch()`, so it functions as-is but the SDK may need to be modified to take `EdgeIo.fetch` instead if that's desired.
* Transactions reach consensus in less than a couple seconds usually so it doesn't really make sense to submit and then poll for results; instead `broadcastTx` just directly waits for consensus.
* Transaction fees are assessed in addition to the transfer amount, so the current UI behavior of subtracting the fee from the total amount to show the amount transferred doesn't really make sense.

I should also note that I tried to run `npm run format` but it failed with the following message:

```
> npm run format

> edge-currency-accountbased@0.4.5 format /home/austin/edge/edge-currency-accountbased
> import-sort -l --write '*.js' 'src/**/*.js' 'test/**/*.js'; prettier-eslint --write '*.js' 'src/**/*.js' 'test/**/*.js'

prettier-eslint-cli [ERROR]: There has been an unknown error when running the prettier-eslint CLI. If it's unclear to you what went wrong, then try this: 
  ✅  Run the script again with the LOG_LEVEL environment variable set to "trace"
  2. Search existing issues on GitHub: https://github.com/prettier/prettier-eslint-cli/issues?utf8=%E2%9C%93&q=_rxjs.Observable.bindNodeCallback%20is%20not%20a%20function
  3. Make a minimal reproduction in a totally separate repository. You can fork this one: https://github.com/kentcdodds/prettier-eslint-cli-repro
  4. Post an issue with a link to your reproduction to the issues on GitHub: https://github.com/prettier/prettier-eslint-cli/issues/new
/home/austin/edge/edge-currency-accountbased/node_modules/prettier-eslint-cli/dist/uncaught-exception-handler.js:45
  throw err;
  ^

TypeError: _rxjs.Observable.bindNodeCallback is not a function
    at Object.<anonymous> (/home/austin/edge/edge-currency-accountbased/node_modules/prettier-eslint-cli/dist/format-files.js:143:31)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/home/austin/edge/edge-currency-accountbased/node_modules/prettier-eslint-cli/dist/index.js:14:20)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
npm ERR! code ELIFECYCLE
npm ERR! errno 7
npm ERR! edge-currency-accountbased@0.4.5 format: `import-sort -l --write '*.js' 'src/**/*.js' 'test/**/*.js'; prettier-eslint --write '*.js' 'src/**/*.js' 'test/**/*.js'`
npm ERR! Exit status 7
npm ERR! 
npm ERR! Failed at the edge-currency-accountbased@0.4.5 format script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/austin/.npm/_logs/2019-10-01T02_10_25_965Z-debug.log
```

(Disclosure: I am the principal developer of the [official JavaScript SDK](https://github.com/hashgraph/hedera-sdk-js) used in this integration but not a direct employee of Hedera. This is, however, an officially sanctioned effort.)


swansontec:
Thanks for your pull request!

Let me speak to a few of your issues:

- Our "public key" nomenclature is actually a mistake. Really, it's just any on-blockchain unique account identifier, and not necessarily a public key as such. For Ethereum, the "public key" is actually an address, even though addresses are hashes, not keys. Also, I am not happy with Paul's architecture around key management, since it forces this bad nomenclature on all plugins. From what you have said, it sounds like the following key format would be ideal for your chain:

```
type ReadOnlyHederaKeys = {
  shard: string,
  realm: string,
  account: string
}
type SpendingHederaKeys = ReadOnlyHederaKeys & {
  hederaKey: string // ed25519 spending key
}
```

Having the user cram this into a single string for our import feature is bit unfortunate.

- The EOS chain also costs money to open accounts, and we have a UX workflow to handle this. However, it's bespoke for the EOS chain. We could probably elements of this workflow for Hedera as well.
- We aren't too concerned about using `EdgeIo` for everything. The currency plugins run in a WebView, but the `EdgeIo` provides a link back to the React Native side of things. This provides an "escape hatch" for accessing native functionality like accelerated crypto or CORS-less networking. If you library works with what the WebView provides, you don't need to do anything extra.
- The amount of a transaction is the amount that leaves the user's wallet, for any reason (the "bottom line" from a wallet balance perspective). It's not the amount that goes to the recipient. So, if the user pays fees to send funds, those fees should be added to the transaction total. I know this is weird, but Bitcoin lets you send to multiple recipients, and Ethereum lets you send to zero recipients (like smart contract stuff), so this was the only way to unify the different coins under a common structure.

Regarding the `yarn format` woes, what version of Node.js are you running? Also, did you install the library using `yarn` or `npm`? It looks like some kind of version mismatch. Another thing you could try is deleting the `rxjs` line in the package.json `resolutions` section and then reinstalling, since it looks like that may be related. If all else fails, it looks like this repo is using an older linting setup, so I could go ahead and upgrade that to see if it helps.

abonander:
Thanks for the quick response!

I did figure out that "public key" was an alias for "public address" which we can use for account IDs. 

I do agree that requiring the account ID and private key in the same text field is sub-optimal, but I don't think a specialized UI is strictly necessary; we have the de-facto string format for IDs as mentioned above. Hedera uses that in many places already, like on https://portal.hedera.com (which enabled users to make accounts on the network before open-access) and we support to/from string conversions in various SDKs. For example, `{ shard: 0, realm: 0, account: 1001 }` would stringify to `0.0.1001`. We could also just accept the account number itself as non-zero `shard` and `realm` are currently not used/supported by the network, so a form like `0.0.____` might work (but wouldn't exactly be future-proof). In the SDKs we accept both forms for maximum flexibility.

I'll need to ask about getting some sort of exchange or backend support for creating an account for a user, but I can't make any promises; we can trivially support creating another Hedera wallet if the user has one set up already.

We could also implement a specialized `makeSend()` flow that generates a URI to request an account creation from another Edge user. That would allow both the requesting user and the creating user to keep their private keys confidential (the requesting user only needs to supply their Ed25519 public key for creation anyway). The user that creates the account would need some sort of side-channel for communicating the ID of the created account back to the requesting user, though.

Unfortunately, we don't have a good way to calculate transaction fees up-front; the user only specifies a maximum fee they want to pay and the transaction will fail if the calculated fee exceeds that so the user doesn't overspend. After the transaction is processed, the assessed fee is in the generated transaction record which has an associated cost to query from the network. Theoretically a transaction explorer could provide this information retroactively but we don't have one yet.

I did install Node 10.15.1 as per the developer instructions; I installed via yarn. The installed version of `rxjs` is the same one as under the `resolutions` key. However, if the build is passing is this step necessary? I fixed all the lint errors which included style lints.

swansontec:
Edge actually uses a small server to handle EOS account creation for our users. The server has a wallet loaded with EOS, and accepts payment in the form of BTC, LTC, and a few other coins. I wonder if we could re-purpose this server for creating Hedera accounts as well. Since the user only provides an ed25519 public key, this should be safe to operate in a non-custodial way.

The transaction fee thing is rather weird, but Ethereum is the same way - you specify a gas limit, but the actual fee may be less. In that case, we get the fee back from the network after we perform the spend. Without this information, how can a user ever know what their balance is? If I have $5, and then spend $4 + some unknown fee, my remaining balance could be anywhere between $0 or $1.

The code formatting isn't critical, but I don't like having things broken. I need to investigate this problem further.

abonander:
Having a server to create Hedera accounts sounds good to me, although the reference currency for HBAR is USD so that may involve some conversion. I assume Edge would purchase their own HBAR to support this? I don't have the authority to work out a different arrangement than that, but I can get a hold of someone who does.

Fortunately, Hedera does support free queries for account balance so that part we can give an exact and up-to-date answer on. We could also fetch records with every transaction to get the fee assessed and add the cost of that query to the fee that's displayed. I'm not sure how much that costs right now so I'll have to ask around.

swansontec:
That's what I was thinking - we buy EOS on an exchange to keep our server loaded, and then our payment software figures out the current BTC price to keep the purchase amount stable. It's basically like running a gumball machine, except that the gumballs are crypto accounts.

On the flip side, we don't actually have the resources to build & operate this server. Ideally, this would be a piece of infrastructure somebody should build & operate for the benefit of the Hedera network (& for profit).

abonander:
I've asked about what we can do there. In the meantime is there anything else blocking this PR as it is?

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,325 @@
...
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json'
+          },
+          body: JSON.stringify({
+            txn_id: txn.txid
+          })
+        }
+        const response = await this.io.fetch(
+          `http://10.0.2.2:8080/v1/account/request/${requestId}/payment`,
```
abonander: TODO: swap this URL with that of our deployed creator API

abonander:
@swansontec I have added integration with an account creator API that we are standing up, the URLs in the code right now are placeholders.

swansontec:
Awesome. We are pretty busy at the moment, but I will see if I can find some time to take a look at this PR. Let me know when the URI's are finalized, so we can test this live.

src/hedera/hederaInfo.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    {
+      name: 'tHBAR',
+      multiplier: '1',
+      symbol: hbarUnitSymbols.tinybar
+    }
+  ],
+  // FIXME: upload to EDGE's image server
+  symbolImage: `http://10.0.2.2:8000/hedera-white.png`,
+  symbolImageDarkMono: `http://10.0.2.2:8000/hedera-black.png`,
```
abonander: @swansontec we have images for these but where should they go? In the `edge-react-gui` we have these same images in `assets/` but I don't think we can link those here.

abonander:
All right, swapped the endpoints.

abonander:
@swansontec this needs a rebase but should otherwise be ready to review.

src/hedera/hederaPlugin.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  importPrivateKey(
+    privateKey: string
+  ): Promise<{
+    hederaAccount: string,
+    hederaPrivateKey: hedera.Ed25519PrivateKey
+  }> {
+    console.log('hederaPlugin/importPrivateKey', privateKey)
```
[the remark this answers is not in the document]
abonander: There is mnemonic support in the JS SDK but it only supports 24-word mnemonics with the BIP-39 English word list. Is that sufficient?

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,347 @@
...
+            txn_id: txn.txid
+          })
+        }
+        const response = await this.io.fetch(
+          `https://api.kabuto.sh/v1/account/request/${requestId}/payment`,
+          options
+        )
+
+        if (response.ok) {
+          this.addToLoop('checkAccountCreationStatus', 5000)
```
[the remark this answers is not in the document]
abonander: I actually have that already on line 200.

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,347 @@
...
+
+    if (!this.client) {
+      await this.checkAccountCreationStatus()
+    }
+  }
+
+  async startEngine() {
+    this.log('hedera start engine called')
+    this.engineOn = true
+    await this.updateBalance()
```
paullinator: this should be part of the loop only

package.json
```diff
@@ -44,6 +44,7 @@
   "dependencies": {
     "@binance-chain/javascript-sdk": "^2.14.4",
     "@dapix/react-native-fio": "0.1.0",
+    "@hashgraph/sdk": "^0.7.1",
```
abonander: This needs to be updated to 1.1

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,347 @@
...
+          method: 'POST',
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json'
+          },
+          body: JSON.stringify({
+            txn_id: txn.txid
+          })
+        }
+        const response = await this.io.fetch(
```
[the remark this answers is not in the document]
abonander: We talked about it internally and we're going to revisit the implementation of the creator so it accepts signed transactions and validates then submits them itself so that part won't be the client's responsibility anymore.

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,347 @@
...
+      this.client,
+      edgeTransaction.otherParams.transferTx
+    )
+    await transferTx.sign(this.privateKey)
+
+    return {
+      ...edgeTransaction,
+      otherParams: {
+        ...edgeTransaction.otherParams,
+        signedTx: transferTx.toBytes()
```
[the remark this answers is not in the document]
abonander: Any preference on format? I'll probably just convert to Base64 if not.

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,347 @@
...
+      await hedera.Transaction.fromBytes(
+        this.client,
+        signedTx
+      ).executeForReceipt()
+    } catch (e) {
+      this.log(e)
+      throw e
+    }
+    // must be > 0 to not show "Synchronizing"
+    edgeTransaction.blockHeight = 1
```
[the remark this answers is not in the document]
abonander: Since Hedera doesn't have a concept of block height, should I override `getBlockHeight()` to return a value greater than 1 or set it in `walletLocalData`?

src/hedera/hederaEngine.js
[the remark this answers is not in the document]
paullinator: base64 is fine

src/hedera/hederaPlugin.js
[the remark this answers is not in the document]
abonander: @paullinator in this context how do I know whether the private key string is a mnemonic or our encoded account ID / private key pair?

I believe there's some refactoring to be done so the account ID isn't part of the private key but I wasn't sure what screen in `edge-react-gui` the add-wallet flow used to get the account ID (which I think is a feature of either EOS or Stellar as well?)

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,347 @@
...
+    currencyPlugin: HederaPlugin,
+    walletInfo: EdgeWalletInfo,
+    opts: EdgeCurrencyEngineOptions,
+    io: EdgeIo
+  ) {
+    super(currencyPlugin, walletInfo, opts)
+    this.log('hedera constructor called', walletInfo)
+    this.hederaPlugin = currencyPlugin
+
+    const { hederaPrivateKey } = walletInfo.keys
```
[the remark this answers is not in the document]
abonander: Is the usage correct now?

src/hedera/hederaEngine.js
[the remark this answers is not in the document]
abonander: I overrode `getBlockHeight()` to always return `2`.

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,404 @@
...
+    } else {
+      return {
+        publicAddress: '',
+        publicKey: this.walletInfo.keys.publicKey
+      }
+    }
+  }
+
+  getBlockHeight(): number {
+    return 2
```
abonander, line 392: Even returning this value, the wallet still shows as synchronizing for some reason.

src/hedera/hederaInfo.js
```diff
@@ -0,0 +1,57 @@
...
+import { type EdgeCurrencyInfo } from 'edge-core-js/types'
+
+// import { imageServerUrl } from '../common/utils'
+import { type HederaSettings } from './hederaTypes.js'
+
+export const useTestnet = true
+
+const otherSettings: HederaSettings = {
+  creatorApiServers: useTestnet
+    ? ['http://10.0.2.2:8080'] // ? ['https://creator.testnet.myhbarwallet.com']
```
abonander, line 14: Will fix, had to make changes to these services during development.

abonander: On this note, I didn't see much of a convention for switching e.g. the Ethereum plugin to use testnets, besides a comment on the `chainId` parameter here: https://github.com/EdgeApp/edge-currency-accountbased/blob/a7e3b9f4162b42e33cb01d355b932174b9c154f3/src/ethereum/ethEngine.js#L532

Would you prefer if I remove the ternary and leave it on the mainnet values for merge?

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,404 @@
...
+            currencyCode,
+            amount: bns.toFixed(response.amount, 3, 9),
+            exchangeAmount: '0'
+          }
+        }
+
+        throw new Error({
+          message: 'ErrorActivationRequest',
+          detail: response.message
+        })
```
abonander, line 94: Do you have a convention for passing additional context with an error? I could have sworn I saw this pattern elsewhere in the codebase but I could be misremembering.

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,404 @@
...
+        this.walletLocalData.otherData.hederaAccount = accountId.toString()
+        this.walletLocalDataDirty = true
+      } else {
+        this.addToLoop('checkAccountCreationStatus', 5000)
+      }
+    } else {
+      this.accountId = new hedera.AccountId(
+        this.walletLocalData.otherData.hederaAccount
+      )
+    }
```
abonander, line 217: This is duplicated with the above because I'm not sure where the most appropriate place to do this check is.

src/hedera/hederaPlugin.js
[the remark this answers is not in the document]
abonander: We solved the account ID problem by looking up by public key but I'm still not sure how you'd prefer to tell the difference between a mnemonic and a private seed here.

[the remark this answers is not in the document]
abonander: Also, any policy on logging confidential data like this? It was mostly for debugging, I can remove it now if you prefer.

src/hedera/hederaInfo.js
abonander: Are those images I sent on the server now?

abonander:
@paullinator @swansontec Okay, I have refactored wallet import to take just the private key and look up the account by public key like EOS does, and updated the [companion PR in edge-react-gui](https://github.com/EdgeApp/edge-react-gui/pull/1644) to match.

Account activation works with the creator service we're spinning up, though right now it requires a manual resync of the wallet to get the account ID showing in the UI, and there's nothing that shows right now when the account is still pending activation. I'm not sure how the EOS plugin notifies the frontend when the account is activated, nor how the frontend knows if the account has not been activated yet (without having to call `CurrencyEngine.makeSpend()` to get `ErrorAccountNotActivated`).

I have a handful of other questions as comments on this PR if you would like to give it another preliminary review, and I think the edge-react-gui PR is ready for a look-over as well: https://github.com/EdgeApp/edge-react-gui/pull/1644

src/hedera/hederaInfo.js
paullinator: If you look at the recent architecture of the Ethereum plugin, it can export multiple currency types with the only difference being the API servers and chain ID. IT would show up as a different wallet type that a user can create, but this is one way to expose testnet.

src/hedera/hederaEngine.js
paullinator: @swansontec Can you comment on the best method of passing errors. I know you thought about this one a lot.

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,404 @@
...
+  }
+
+  async startEngine() {
+    console.log('hederaEngine/startEngine', this)
```
paullinator, line 197: Do not console.log whole objects. This gets messy and could be a security issue.

src/hedera/hederaEngine.js
paullinator: Sorry, I dont quite understand what you are referring to here. It might be good to get on a call to discuss further.

src/hedera/hederaPlugin.js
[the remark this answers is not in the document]
paullinator: Regarding the mnemonic versus private key, for other currencies we try to decode the string as a private key first, and if that fails then try to decode it as a mnemonic. There are no cases where a string could be both. 

Definitely remove any logging of sensitive information.

src/hedera/hederaEngine.js
abonander: This was for debugging during development, I'll be removing most of these log calls.

src/hedera/hederaInfo.js
abonander: That would be pretty interesting, although on testnet we still have the issue of account activation. We could support imports only for testnet wallets (since [the Hedera portal](https://portal.hedera.com/) automatically creates testnet accounts for registered users and tops them up daily with HBAR).

src/hedera/hederaEngine.js
abonander: This block at lines 203-217 is duplicated between `startEngine()` and `loadEngine()` because I wasn't sure where the best place to have this logic is and opted for redundancy rather than making the wrong choice.

abonander:
Removed all logging calls that might reveal confidential information.

src/hedera/hederaEngine.js
abonander: Needs to call `this.currencyEngineCallbacks.onAddressesChecked` with 1 when syncing is done.

src/hedera/hederaEngine.js
abonander: `startEngine()` only

abonander:
* Don't wait for receipt

src/hedera/hederaEngine.js
```diff
@@ -0,0 +1,512 @@
...
+
+      if (txs) {
+        this.processTxs(txs)
+      } else {
+        this.earliestReached = true
+      }
+    }
+
+    // tell the UI that we're done syncing
+    this.currencyEngineCallbacks.onAddressesChecked(1)
```
abonander, line 291: We can't really give more granular updates right now since Kabuto only gives transactions in reverse chronological order with a hard page limit, so right now we start by fetching the latest transactions and then going backwards in time.

If it's preferable to start at the beginning and work forward, that's possible but will need to be implemented in Kabuto; however, this can then be updated in terms of `latestTx.consensusAt / Date.now()` and set to `1` when finished.

abonander: I figured out a way to give a decent progress indicator that's basically the inverse of the above.

peachbits:
@abonander Please run yarn precommit and resolve the flow errors. Once done we can pass this to QA

abonander:
@peachbits I have addressed the Flow and lint errors for `src/hedera` but there are errors in parts of the code which this PR does not touch even after rebasing on current `master`; please advise.

abonander:
@peachbits I ran `yarn` after the rebase but as a sanity check I ran `rm -rf node_modules && yarn` and after fixing one error in `hederaInfo.js` now `yarn run precommit` passes.

PR #88, opened 2019-11-10
Tron Integration

MarcelBlockchain:
Notes to the updated PR:

The biggest change are the transaction fees. Every Tron user has a daily amount of free transactions. These are pulled using `trx_getAccountResource`. As long as the users `resources.freeNetLimit` is higher than `268`, the transaction is for free.

In case the account, the user sends tokens to is not yet activated, an activation fee of 0.1 TRX will be added. See more details below.

The costs for account activation and transaction per byte are pulled from the network using `async checkUpdateNetworkFees`. It’s unit is SUN (1 SUN = 0.000001 TRX) and are converted to TRX using `tronWeb.fromSun(myvalue)`. 

Acitivation fee for not activated receiving accounts: 100000 SUN

Transaction cost per byte: 10 SUN

For a tron transfer you need to pay for 268 bytes (`raw_data_hex.length`)

These base values won’t be changed anytime soon.

From the docs:

"The TRON network grants every account a free pool of Bandwidth Points for free transactions every 24 hours. To engage in transactions more frequently requires freezing TRX for additional bandwidth points, or paying the fee in TRX.

There are 5000 bandwidth points for free per account per day. When an account hasn’t frozen any balance, or when its bandwidth points have run out, complimentary bandwidth points can be used”

https://developers.tron.network/docs/bandwith

---

In case you want to start testing on shasta test net with free money, prepend 'shasta' to the link in tronEngine.js, tronPlugin.js and tronInfo.js

const tronWeb = new TronWeb({
  fullHost: 'https://api.trongrid.io'
})

--> api.shasta.trongrid.io

free shasta money: https://www.trongrid.io/shasta

Tron Metamask Alternative: https://github.com/TronLink/tronlink-extension

### DIFF
--- a/src/tron/tronEngine.js
+++ b/src/tron/tronEngine.js
@@ -12,13 +12,14 @@ import TronWeb from 'tronweb'
 import { CurrencyEngine } from '../common/engine.js'
 import {
   asyncWaterfall,
-  getDenomInfo,
   shuffleArray,
   validateObject
 } from '../common/utils.js'
 import { currencyInfo } from './tronInfo.js'
 import { TronPlugin } from './tronPlugin.js'
 import {
+  NetworkFeesSchema,
+  TronAccountResources,
   TronApiAccountBalance,
   TronApiGetTransactions,
   TronApiNodeInfo,
@@ -33,33 +34,38 @@ const tronWeb = new TronWeb({
 const PRIMARY_CURRENCY = currencyInfo.currencyCode
 const ACCOUNT_POLL_MILLISECONDS = 20000
 const BLOCKCHAIN_POLL_MILLISECONDS = 20000
-const TRANSACTION_POLL_MILLISECONDS = 10000 // was 3000
+const TRANSACTION_POLL_MILLISECONDS = 3000 // was 3000
 const ADDRESS_QUERY_LOOKBACK_BLOCKS = 15 * 60 * 24 // ~ one day. 3-5 sec block time
+const NETWORKFEES_POLL_MILLISECONDS = 60 * 10 * 1000
 const NUM_TRANSACTIONS_TO_QUERY = 200
 const TIMESTAMP_BEFORE_TRX_LAUNCH = 1529920269000 // 2019-04-01, TRX launched on 6/25/2018 09:51:09
 const TRANSACTION_QUERY_TIME_WINDOW = 1000 * 60 * 60 * 24 * 3 * 28 // 3 months
-// const NATIVE_UNIT_MULTIPLIER = '1000000'
+const TX_SIZE = '268' // in bytes, raw_data_hex.length of any "TransferContract" transaction
+const NATIVE_UNIT_MULTIPLIER = '1000000'
 
 type TronFunction =
   | 'trx_txBlockNumber'
   | 'trx_blockNumber'
   | 'trx_getBalance'
   | 'trx_getTransactions'
+  | 'trx_getAccountResource'
 
 export class TronEngine extends CurrencyEngine {
   constructor (
     currencyPlugin: TronPlugin,
     walletInfo: EdgeWalletInfo,
-    initOptions: any,
     opts: any, // EdgeCurrencyEngineOptions
     fetchCors: Function
   ) {
+    // $FlowFixMe we need fetchCors here
     super(currencyPlugin, walletInfo, opts, fetchCors)
+    // $FlowFixMe fetchCors is not missing in TronEngine
     this.fetchCors = fetchCors
   }
 
   async fetchGet (url: string) {
     const options = { method: 'GET' }
+    // $FlowFixMe fetchCors is not missing in TronEngine
     const response = await this.fetchCors(url, options)
     if (!response.ok) {
       throw new Error(
@@ -78,6 +84,7 @@ export class TronEngine extends CurrencyEngine {
       },
       body: JSON.stringify(params)
     }
+    // $FlowFixMe fetchCors is not missing in TronEngine
     const response = await this.fetchCors(url, options)
     if (!response.ok) {
       this.log(`The server returned error code ${response.status} for ${url}`)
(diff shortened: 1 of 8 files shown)


src/tron/tronEngine.js
```diff
@@ -0,0 +1,548 @@
...
+    }
+
+    const otherParams: Object = {}
+
+    let blockHeight = tx.blockNumber
+    if (blockHeight < 0) blockHeight = 0
+    const unixTimestamp = new Date(tx.block_timestamp)
+    const edgeTransaction: EdgeTransaction = {
+      txid: tx.txID,
+      date: unixTimestamp.getTime(),
```
paullinator: This is setting a JavaScript timestamp which is incorrect as the Edge transaction object is expecting a UNIX timestamp. You need to divide by 1000 to get a UNIX timestamp.I think the problem is being hidden by the fact that you are passing a UNIX timestamp to the new date constructor which is expecting a JavaScript timestamp. I think this bug also exists in BNB so I see where you carried over the issue. If the timestamp from the API is already in a UNIX number format, then you can simply assign that directly to the date with no additional conversion

src/tron/tronEngine.js
```diff
@@ -0,0 +1,548 @@
...
+
+      // **********************************
+      // Create the unsigned EdgeTransaction
+
+      const edgeTransaction: EdgeTransaction = {
+        txid: '', // txid
+        date: 0, // date
+        currencyCode, // currencyCode
+        blockHeight: 0, // blockHeight
+        nativeAmount: edgeNativeAmount, // nativeAmount
```
paullinator: Native amount should include the network fee. Network fee is a constant above that is set to zero. Is this really true of Tron? I was not aware that Tron has zero fees.

src/tron/tronEngine.js
```diff
@@ -0,0 +1,548 @@
...
+
+  async signTx (edgeTransaction: EdgeTransaction): Promise<EdgeTransaction> {
+    const privKey = this.walletInfo.keys.tronKey
+
+    const signedTx = await tronWeb.trx.sign(
+      edgeTransaction.otherParams,
+      privKey
+    )
+    // this.log(`SUCCESS TRX broadcastTx\n${JSON.stringify(signedTx, null, 2)}`)
+    edgeTransaction.signedTx = signedTx
```
paullinator: Please make sure that this is a string and not an object. The user interface will show this value if it exists and the user taps on the advanced details

src/tron/tronEngine.js
```diff
@@ -0,0 +1,523 @@
...
+          } else {
+            checkAddressSuccess = false
+            this.log('checkTransactionsFetch inner loop invalid query results.')
+            break
+          }
+        }
+        start = end
+      }
+    } catch (e) {
+      this.log(
```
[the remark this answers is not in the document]
MarcelBlockchain: I'm not following what you mean by 'needs checkAddressSuccess = false'
You mean for this.log() ?
I tried comparing with the other Engines like bnbEngine, and it seems it's the same

src/tron/tronEngine.js
MarcelBlockchain: Should be fine. If you need any code changes, let me know. From the docs:

The TRON network grants every account a free pool of Bandwidth Points for free transactions every 24 hours. To engage in transactions more frequently requires freezing TRX for additional bandwidth points, or paying the fee in TRX.

There are 5000 bandwidth points for free per account per day. When an account hasn’t frozen any balance, or when its bandwidth points have run out, complimentary bandwidth points can be used. Each transaction in the TRON network is about 200 bytes, so each account enjoys about 25 transactions for free each day.

If ran out of free transactions costs are:
For a normal transfer, it costs about 0.002 TRX (USD 0.000028)

Activating an account costs 0.1TRX (USD 0.0014)

1 TRX = 0.014$

https://developers.tron.network/docs/bandwith

src/tron/tronEngine.js
MarcelBlockchain: When changing from Object to String using JSON.stringify(signedTx, null,2), it fails to send a tx:
makeSpend error:  Invalid amount provided

I could successfully copy the 'signed raw hex' when clicking on 'view advanced transaction data' in the transaction details screen when it's an Object, if you meant that.

    edgeTransaction.signedTx = signedTx
VS
    edgeTransaction.signedTx = JSON.stringify(signedTx, null, 2)

Maybe I misunderstood something

src/tron/tronPlugin.js
```diff
@@ -0,0 +1,148 @@
...
+        throw new Error('InternalErrorInvalidCurrencyCode')
+      }
+      amount = bns.div(nativeAmount, denom.multiplier, 18)
+    }
+    const encodedUri = this.encodeUriCommon(obj, 'tron', amount)
+    return encodedUri
+  }
+}
+
+export function makeTronPlugin (opts) {
```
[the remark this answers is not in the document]
MarcelBlockchain: all tests pass, just one fails and I haven't figured out why yet. Will look into that tomorrow again.
`derivePublicKey for Wallet type wallet:tron
       Valid private key: 
AssertionError: expected 'TLACvKsccxHiHBgxGfZjaXdQaako7Rsp76' to equal 'TGMgNz9gEhwXEXZ3to5VdZxEKwkDD3R2BE'`

It's just the test. I triple-checked and importing and deriving keys works in the app just fine

paullinator:
Please run `yarn precommit`. you have several errors to address

src/tron/tronPlugin.js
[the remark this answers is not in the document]
peachbits: Is this test still failing?

src/tron/tronEngine.js
```diff
@@ -177,12 +189,12 @@ export class TronEngine extends CurrencyEngine {
       // ourReceiveAddresses.push(this.walletLocalData.publicKey.toLowerCase())
     }
 
-    const otherParams = {}
+    const otherParams: Object = {}
```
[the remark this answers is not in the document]
peachbits: This it still unresolved. Please create a type for otherParams.

peachbits:
Completed in https://github.com/EdgeApp/edge-currency-accountbased/pull/464

PR #106, opened 2019-12-11
Add Amberdata API and RPC calls to ethNetwork

sfncook:


src/ethereum/ethNetwork.js
```diff
@@ -889,14 +1049,66 @@ export class EthereumNetwork {
...
+  async checkTxsAmberdata(
+    startBlock: number,
+    currencyCode: string
+  ): Promise<EthereumNetworkUpdate> {
+    const address = this.ethEngine.walletLocalData.publicKey
+
+    let page = 0
+    const allTransactions: Array<EdgeTransaction> = []
+    while (1) {
+      const jsonObj = await this.fetchGetAmberdataApi(
```
paullinator: This is missing a way to query only the most recent transactions. Is this not available in the API? If it is not available then we will not implement this API. If it is available then please implement it

sfncook: Oh jeez, of course.  I completely forgot that.  I'll take a look later today.

sfncook: The Amberdata API  has a  "startDate" query param but it's the timestamp, not the blockNumber.  Is there a way to convert the starting blockheight to a timestamp?

PR #107, opened 2019-12-12
Updates RSK engine to the changed Blockscout API

alepc253:
Module `proxy` is no longer available.

### DIFF
--- a/src/rsk/rskEngine.js
+++ b/src/rsk/rskEngine.js
@@ -166,10 +166,14 @@ export class RskEngine extends CurrencyEngine {
         this.updateBalance('RBTC', balance)
       }
     } catch (e) {
-      this.log(`Error checking token balance: RBTC`, e)
-      //  Blockscout returns 404 & 'Balance not found' for new accounts
-      const balance = '0'
-      this.updateBalance('RBTC', balance)
+      let errorMsg = e.toString()
+      //  Blockscout returns 404 for new accounts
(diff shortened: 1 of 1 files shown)


src/rsk/rskEngine.js
```diff
@@ -160,7 +160,10 @@ export class RskEngine extends CurrencyEngine {
         this.updateBalance('RBTC', balance)
       }
     } catch (e) {
-      this.log(`Error checking token balance: RBTC`)
+      this.log(`Error checking token balance: RBTC`, e)
```
paullinator: This needs to actually detect for the 404 "balance not found" before it calls update balance. Otherwise if there is a network error then we will set the balance to zero

src/rsk/rskEngine.js
```diff
@@ -592,6 +582,13 @@ export class RskEngine extends CurrencyEngine {
               this.log(msg)
               throw new Error(msg)
             }
+            // Convert hex
+            result.result = result.result.replace('0x', '')
```
paullinator: Remove this line. Just use isHex as that should take care of the check

src/rsk/rskEngine.js
alepc253: Yes, I totally understand (I tried to get the message but I couldn't. Let's try harder).

src/rsk/rskEngine.js
alepc253: It fails without that. Maybe I can remove the hex validation directly...

PR #111, opened 2019-12-31
William/upgrade core

swansontec:
The latest core is exposing Flow errors because we are accessing `otherParams` without checking for `null`.


swansontec:
Made the requested adjustments from the over-the-shoulder review with @paullinator.

PR #121, opened 2020-02-27
Refactor of ETH (and RSK) Plugin

kylanhurt:


src/ethereum/ethNetwork.js
```diff
@@ -751,14 +795,21 @@ export class EthereumNetwork {
...
-        funcs2 = async () => {
-          const result = await this.fetchPostInfura('eth_getTransactionCount', [
-            params[0],
-            'latest'
-          ])
-          return { server: 'infura', result }
+        if (infuraServers.length > 0) {
+          funcs2 = async () => {
+            const adjustedParams =
+              this.currencyInfo.currencyCode === 'RBTC'
```
kylanhurt: Wondering what you guys think of the reference to 'RBTC'. For some reason their Infura nodes expect the `param` array to be nested within an array (at least that's what works with the node that we are using).

paullinator: Yuck. Put a flag in the currenyInfo

src/ethereum/ethNetwork.js
```diff
@@ -890,11 +947,22 @@ export class EthereumNetwork {
...
-      this.checkBlockHeightAmberdata,
-      this.checkBlockHeightBlockchair
-    ]).catch(err => {
+    const {
+      etherscanApiServers,
+      amberdataApiServers,
+      blockchairApiServers
+    } = this.currencyInfo.defaultSettings.otherSettings
+    const waterfallFuncs = []
+    if (etherscanApiServers.length > 0) {
```
paullinator: The IF statements should not be necessary here

src/ethereum/ethNetwork.js
```diff
@@ -931,10 +999,19 @@ export class EthereumNetwork {
...
-    return asyncWaterfall([
-      this.checkNonceEthscan,
-      this.checkNonceAmberdata
-    ]).catch(err => {
+    const {
+      etherscanApiServers,
+      amberdataApiServers
+    } = this.currencyInfo.defaultSettings.otherSettings
+    const waterfallFuncs = []
+    if (etherscanApiServers.length > 0) {
```
paullinator: The IF statements should not be necessary here

src/ethereum/ethNetwork.js
```diff
@@ -1315,21 +1395,26 @@ export class EthereumNetwork {
...
   }
 
   async checkTokenBalBlockchair(): Promise<EthereumNetworkUpdate> {
     const address = this.ethEngine.walletLocalData.publicKey
+    const {
+      blockchairUrlTokenString
+    } = this.currencyInfo.defaultSettings.otherSettings
     const jsonObj = await this.fetchGetBlockchair(
-      `/ethereum/dashboards/address/${address}?erc_20=true`,
+      `/${this.currencyInfo.pluginName}/dashboards/address/${address}?${blockchairUrlTokenString}=true`,
```
paullinator: blockchairUrlTokenString can be assumed to be `erc_20` all the time

src/ethereum/ethNetwork.js
```diff
@@ -1350,10 +1435,15 @@ export class EthereumNetwork {
...
-    return asyncWaterfall([
-      async () => this.checkTokenBalEthscan(tk),
-      this.checkTokenBalBlockchair
-    ]).catch(err => {
+    const {
+      blockchairApiServers
+    } = this.currencyInfo.defaultSettings.otherSettings
+    const waterfallFuncs = []
+    waterfallFuncs.push(async () => this.checkTokenBalEthscan(tk))
+    if (blockchairApiServers.length > 0) {
```
paullinator: The IF statements should not be necessary here

src/ethereum/ethPlugin.js
```diff
@@ -63,195 +47,583 @@ const defaultNetworkFees = {
...
-    opts: EdgeCurrencyEngineOptions
-  ): Promise<EdgeCurrencyEngine> {
-    const tools = await makeCurrencyTools()
-    const currencyEngine = new EthereumEngine(
-      tools,
-      walletInfo,
-      initOptions,
-      opts
-    )
+export const currencyInfo: EdgeCurrencyInfo = {
```
paullinator: Keep this in an ethInfo.js file

src/ethereum/rskPlugin.js
```diff
@@ -0,0 +1,99 @@
...
+  EdgeCorePluginOptions,
+  EdgeCurrencyInfo
+} from 'edge-core-js/types'
+
+import { makeEthereumBasedPluginInner } from './ethBasedPlugin.js'
+import type { EthereumSettings } from './ethTypes.js'
+
+export const imageServerUrl = 'https://developer.airbitz.co/content'
+
+const defaultNetworkFees = {
```
paullinator: Keep this in an rskInfo.js file

alepc253:
I tested it locally and seems to be working OK

src/ethereum/ethEngine.js
```diff
@@ -278,10 +289,15 @@ export class EthereumEngine extends CurrencyEngine {
   async startEngine() {
     this.engineOn = true
     this.addToLoop('checkUpdateNetworkFees', NETWORKFEES_POLL_MILLISECONDS)
-    this.addToLoop(
-      'checkUnconfirmedTransactionsInnerLoop',
-      UNCONFIRMED_TRANSACTION_POLL_MILLISECONDS
-    )
+    if (
+      this.currencyInfo.defaultSettings.otherSettings
```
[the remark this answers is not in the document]
kylanhurt: @paullinator I don't see anywhere in the code to checks that there actually is a URL for this call before deciding whether to do the `fetch` call or not. Here is the code in `ethEngine`:
```
    const url = `${this.currencyInfo.defaultSettings.otherSettings.superethServers[0]}/v1/${lowerCaseCurrencyCode}/main/txs/${address}`
    let jsonObj = null
    try {
      jsonObj = await this.ethNetwork.fetchGet(url)
```

Can I use the length of the `superethServers` array as the flag for whether or not to fire off the fetch routine?

src/ethereum/ethNetwork.js
```diff
@@ -671,142 +708,169 @@ export class EthereumNetwork {
...
-        promises.push(
-          broadcastWrapper(this.broadcastInfura(params[0]), 'infura')
-        )
-        promises.push(
-          broadcastWrapper(this.broadcastEtherscan(params[0]), 'etherscan')
-        )
-        promises.push(
-          broadcastWrapper(this.broadcastBlockCypher(params[0]), 'blockcypher')
-        )
+        if (infuraServers.length > 0) {
```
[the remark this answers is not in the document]
paullinator: Look at how eth_getBalance works

src/ethereum/ethNetwork.js
```diff
@@ -163,6 +161,7 @@ export class EthereumNetwork {
       if (tx.gasPrice) {
         nativeNetworkFee = bns.mul(tx.gasPrice, tx.gasUsed)
       } else {
+        // implemented for ETH, not RSK
```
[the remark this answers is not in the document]
kylanhurt: It just means that the `else` clause existed in the Ethereum version of the function, but not the RSK version, before the refactor. So if there was no `tx.contractAddress` and no `tx.gasPrice` then `nativeNetworkFee` was not being assigned to `'0'` for RSK before.

PR #124, opened 2020-03-13
Neo Integration

MorganDream:
Mostly I referred codes from other assets integration and it may have mistakes.
I have written the UT and it passed


swansontec:
In order to resolve the merge conflicts & test this, I rebased the branch as [william/rebased-neo-plugin](https://github.com/EdgeApp/edge-currency-accountbased/tree/william/rebased-neo-plugin). I also had to create a matching GUI branch at [william/add-neo](https://github.com/EdgeApp/edge-react-gui/tree/william/add-neo) to actually be able to use this plugin.

When I run the plugin on those branches, my wallet does not display a balance even though it [has funds](https://neotracker.io/address/AVeSyjDNqF2Xvsus4qq85fawrfuLzHJYZR). The error I see in the console is `Error checkTransactionsFetch NEO TypeError: e.sent.json is not a function`.

wanglongfei88:
@swansontec can you please try #154 ? We have fixed the issue that you mentioned above.

src/neo/neoSchema.js
```diff
@@ -0,0 +1,47 @@
+export const NeoTransactonOnline = {
```
peachbits, line 1: Typo in the variable name is causing tests to fail. Should be 'NeoTransactionOnline'

peachbits (changes requested):

src/neo/neoSchema.js
wanglongfei88: Let's move to #154

PR #141, opened 2020-04-23
Remove reference to EOS owner key (for import)

kylanhurt:


kylanhurt:
> I would think you also need to change any code that also tries to use `eosOwnerKey`

The places where `eosOwnerKey` is used are all protected by conditionals checking for its existence. These areas and their conditionals have all existed for 2 years

PR #143, opened 2020-05-07
Add ETC support

passabilities:


src/ethereum/etcInfo.js
```diff
@@ -0,0 +1,121 @@
...
+  etherscanApiServers: ['https://blockscout.com/etc/mainnet'],
+  blockcypherApiServers: [],
+  superethServers: [],
+  infuraServers: [],
+  isNestedInfuraParams: false,
+  infuraNeedProjectId: true,
+  uriNetworks: ['ethereumclassic', 'etherclass'],
+  ercTokenStandard: 'ERC20',
+  chainId: 1,
+  hdPathCoinType: 60,
```
peachbits, line 61: Needs to be 61

src/ethereum/etcInfo.js
```diff
@@ -0,0 +1,121 @@
...
+  checkUnconfirmedTransactions: true,
```
peachbits, line 62: Set to false. We don't have a server for unconfirmed transactions.

src/ethereum/etcInfo.js
```diff
@@ -0,0 +1,121 @@
...
+  iosAllowedTokens: {
```
peachbits, line 63: Remove iosAllowedTokens as it is deprecated.

passabilities: This object appears to be required by the parent object type. For now I just made it an empty object. Unless I should just remove the type and delete from both `ethInfo` and `etcInfo`

PR #146, opened 2020-05-20
FIO. Add domains to fio engine

andreyvEze:


src/fio/fioEngine.js
```diff
@@ -574,6 +591,15 @@ export class FioEngine extends CurrencyEngine {
           expiration: fioAddress.expiration
         })
       }
+
```
andreyvEze, line 594: check if changed before saving

src/fio/fioEngine.js
```diff
@@ -208,6 +214,15 @@ export class FioEngine extends CurrencyEngine {
...
```
andreyvEze, line 217: add new commit to remove this from loadEngine

PR #148, opened 2020-05-22
Add makeMutex to wrap makeSpend() to avoid entering it more than once…

peachbits:
… at a time.


peachbits:
Paul's verbal approval

PR #174, opened 2020-07-09
Fix FIO plugin multicast servers.

andreyvEze:


src/fio/fioUtil.js
```diff
@@ -0,0 +1,22 @@
+// @flow
```
andreyvEze, line 1: remove this, use existing one

src/fio/fioPlugin.js
```diff
@@ -106,22 +108,65 @@ export function makeFioPlugin(opts: EdgeCorePluginOptions): EdgeCurrencyPlugin {
...
+                json: e.json,
+                list: e.list
+              }
+            }
+          } else {
+            throw e
+          }
+        }
+
+        return res
```
andreyvEze, line 154: use `out` name

src/fio/fioEngine.js
```diff
@@ -15,13 +15,13 @@ import {
 } from 'edge-core-js/types'
 
 import { CurrencyEngine } from '../common/engine.js'
-import { asyncWaterfall, getDenomInfo } from '../common/utils'
+import { asyncWaterfall, getDenomInfo, shuffleArray } from '../common/utils'
```
andreyvEze, line 18: @paullinator I removed my shuffle method, but have a question, in my method I shuffled copy of the array, not actual array
```
export const shuffle = (array: any[]): any[] => {
  const shuffledArr = [...array]
```
should we add this to common shuffle method?

PR #176, opened 2020-07-15
Tezos - flow, fetch, and api fixes

peachbits:


src/tezos/tezosTypes.js
```diff
@@ -4,16 +4,30 @@ export type TezosSettings = {
 }
 
 export type XtzGetTransaction = {
-  block_hash: string,
+  type: string,
```
paullinator, line 7: This should just be a cleaner

PR #184, opened 2020-08-07
Add numTransactions to walletLocalData

peachbits:


src/common/engine.js
```diff
@@ -171,6 +171,10 @@ export class CurrencyEngine {
     try {
       const result = await disklet.getText(TRANSACTION_STORE_FILE)
       transactionList = JSON.parse(result)
+      for (const currencyCode in transactionList) {
```
paullinator, line 174: This needs to be done after line 204

src/common/engine.js
```diff
@@ -322,6 +326,7 @@ export class CurrencyEngine {
       }
       // add transaction to list of tx's, and array of changed transactions
       this.transactionList[currencyCode].push(edgeTransaction)
+      this.walletLocalData.numTransactions[currencyCode]++
```
paullinator, line 329: Safer to just set the number of transactions equal to the length of the transaction list array. Incrementing should work but just in case the count is out of sync, using the length will put it back in sync.

src/common/engine.js
```diff
@@ -322,6 +327,9 @@ export class CurrencyEngine {
       }
       // add transaction to list of tx's, and array of changed transactions
       this.transactionList[currencyCode].push(edgeTransaction)
+      this.walletLocalData.numTransactions[currencyCode] = this.transactionList[
+        currencyCode
+      ].length
 
       this.transactionListDirty = true
```
paullinator, line 334: Eyk. I just noticed. You also need to set `this.Walletlocaldatadirty = true` Otherwise your change to numtransactions will never get saved

PR #203, opened 2020-09-24
Update Telos Block Explorer and Import EOSIO Active Key

kylanhurt:


src/eos/eosPlugin.js
```diff
@@ -104,6 +104,8 @@ export class EosPlugin extends CurrencyPlugin {
       // but rather just different versions of the master public key
       // const publicKey = derivePubkey(walletInfo.keys.eosKey)
       // const publicKey = deriveAddress(walletInfo.keys.eosKey)
+      // set default publicKey to ACTIVE public key = eosKey
+      // because that is what keys we will encourage to import will get put, do NOT use ownerKey
```
paullinator, line 108: There is no actual code change here

yarn.lock
```diff
@@ -828,10 +828,10 @@
     randombytes "^2.1.0"
     text-encoding "0.7.0"
 
-"@fioprotocol/fiosdk@^1.0.2":
-  version "1.0.2"
-  resolved "https://registry.yarnpkg.com/@fioprotocol/fiosdk/-/fiosdk-1.0.2.tgz#fece083173ef7e309b49e9312538fffa1896a346"
-  integrity sha512-ABukfhRfUqmu+tHqfUUBqw2FSIfkaMUp4ALtlXZwYBd1ZgH0pYBbzrSGR8KHJyEXcs3sEvWb9nh1alu68ztDLg==
+"@fioprotocol/fiosdk@^1.1.0":
```
paullinator, line 831: This should not have changed. Don't commit lock files unless you actually change the package

paullinator (changes requested):

src/eos/eosPlugin.js
kylanhurt: Good catch

PR #213, opened 2020-11-06
RBF support for ETH, RSK, and ETH tokens

samholmes:
* Modifies `ethEngine.js`
* Adds support for new `rbfTxid` field in `EdgeSpendInfo` in `makeSpend`
* Adds new `nonceUsed` for `EthereumTxOtherParams`
* `makeSpend` uses replaced transaction's fee and nonce
* Saves `nonceUsed` in the `otherParams` field for an ETH `EdgeTransaction`

### DIFF
--- a/src/ethereum/ethEngine.js
+++ b/src/ethereum/ethEngine.js
@@ -402,7 +402,8 @@ export class EthereumEngine extends CurrencyEngine {
         errorVal: 0,
         tokenRecipientAddress: null,
         nonceArg: rbfNonce,
-        data: data
+        rbfTxid,
+        data
       }
       otherParams = ethParams
       value = bns.add(nativeAmount, '0', 16)
@@ -431,6 +432,7 @@ export class EthereumEngine extends CurrencyEngine {
         errorVal: 0,
         tokenRecipientAddress: publicAddress,
         nonceArg: rbfNonce,
+        rbfTxid,
         data
       }
       otherParams = ethParams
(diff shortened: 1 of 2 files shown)


src/ethereum/ethEngine.js
```diff
@@ -709,6 +711,28 @@ export class EthereumEngine extends CurrencyEngine {
...
+      const txid = edgeTransaction.otherParams.rbfTxid
+      const idx = this.findTransaction(currencyCode, txid)
+      const replacedEdgeTransaction = this.transactionList[currencyCode][idx]
+
+      // Update the transaction's blockHeight to -1 (drops the transaction)
+      const updatedEdgeTransaction: EdgeTransaction = {
+        ...replacedEdgeTransaction,
+        blockHeight: -1
+      }
+      this.updateTransaction(currencyCode, updatedEdgeTransaction, idx)
```
paullinator, line 731: This should actually use addTransaction. Yes, it's a poor name but add transaction will also update transactions if the transaction already exists. It also does additional validation necessary

samholmes: Amended commit and tested it using addTransaction.

src/ethereum/ethEngine.js
```diff
@@ -322,6 +322,31 @@ export class EthereumEngine extends CurrencyEngine {
...
+      if (rbfTxIndex > -1) {
+        const rbfTrx = this.transactionList[currencyCode][rbfTxIndex]
+
+        if (rbfTrx.otherParams) {
+          const { gasPrice, gas, nonceUsed } = rbfTrx.otherParams
+          rbfGasPrice = bns.mul(gasPrice, '2')
+          rbfGasLimit = gas
+          rbfNonce = nonceUsed
+        }
+      }
```
paullinator, line 347: Need to fail the make spend if the RBFTX ID cannot be found.

paullinator: As well if the other parameters cannot be found

paullinator: Otherwise we will end up sending a duplicate transaction that the user did not intend

samholmes: Rebased the commit to include check that throws an exception when data is missing. This catches both the case where the transaction is not found or the data on a found transaction is missing.

https://github.com/EdgeApp/edge-currency-accountbased/pull/213/commits/bb5e3b7f48bda4de9ee9a9f885e6a372bcfe3d41#diff-4839c75ac65fd8060df92dcdc5cb0ab5786e8fa09c3ba01287bc97a2c53cf971R349-R351

paullinator (approved):

PR #231, opened 2020-12-22
Changed parseUri, check fio options, add prefix to uri.

andreyvEze:


src/binance/bnbPlugin.js
```diff
@@ -73,11 +74,33 @@ export class BinancePlugin extends CurrencyPlugin {
...
     uri: string,
-    currencyCode?: string,
+    currencyOptions?: string | EdgeParseUriOptions,
     customTokens?: Array<EdgeMetaToken>
   ): Promise<EdgeParsedUri> {
     const networks = { binance: true }
 
+    const currencyCode =
+      typeof currencyOptions === 'string'
+        ? currencyOptions
```
paullinator, line 84: This will crash if currency options is undefined

src/binance/bnbPlugin.js
```diff
@@ -73,11 +74,33 @@ export class BinancePlugin extends CurrencyPlugin {
...
+        : currencyOptions.currencyCode
+    if (
+      typeof currencyOptions === 'object' &&
+      currencyOptions !== null &&
```
paullinator, line 88: Line 88 is unneeded and redundant. Line 87 implies that the variable is defined

src/binance/bnbPlugin.js
```diff
@@ -73,11 +74,33 @@ export class BinancePlugin extends CurrencyPlugin {
...
+    ) {
+      let noNetwork = true
+      for (const networkName in networks) {
+        if (uri.indexOf(networkName) === 0) {
+          noNetwork = false
+          break
+        }
+      }
+
+      if (noNetwork) {
```
paullinator, line 99: Also need to check that the URI does not begin with any scheme at all. Otherwise it will put the binance scheme on top of the scheme it already has

paullinator (changes requested):

src/binance/bnbPlugin.js
```diff
@@ -16,6 +16,7 @@ import {
   type EdgeIo,
   type EdgeMetaToken,
   type EdgeParsedUri,
+  type EdgeParseUriOptions,
```
paullinator, line 19: In the core, make this type a cleaner that you can use throughout the plug-in to validate the options

PR #238, opened 2021-01-07
Upgrade lint

peachbits:


package.json
```diff
@@ -37,7 +37,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run precommit"
```
swansontec, line 40: I think this line should stay, since the `precommit` script does more than just run lint.

swansontec (approved):
One teeny-tiny change needed.

PR #243, opened 2021-01-25
Throw error when entered custom fee is less than currency minimum

peachbits:


src/ethereum/ethMiningFees.js
```diff
@@ -32,6 +32,22 @@ export function calcMiningFee(
...
+      } = currencyInfo.defaultSettings.otherSettings.defaultNetworkFees.default
+      if (
+        gasLimit < minGasLimit ||
+        gasPrice < minGasPrice ||
+        isNaN(gasLimit) ||
+        isNaN(gasPrice)
+      ) {
+        // TODO: Throw proper InvalidCustomFee error
+        throw new Error(
+          `Gas Limit: ${minGasLimit} Gas Price (Gwei): ${minGasPrice}`
```
paullinator, line 47: This does not look like a clean error. Instead create an error object and set `e.name` to `ErrorBelowMinimumFee`

paullinator (changes requested):

samholmes (approved):
Approved but needs an autosquash rebase

PR #252, opened 2021-02-12
Fix contractAddress variable name

peachbits:
Also log result object when checkTokenBalEthscan throws an error


swansontec (approved):
Approved, but in the future we should use `?? ''` instead of `|| ''`, because `??` is the "if null" operator, but `||` is the "if false or empty or zero or null" operator, which is a bit more non-specific and therefore error-prone.

PR #254, opened 2021-02-23
Add dfuse API to getKeyAccounts method

peachbits:


src/eos/eosEngine.js
```diff
@@ -532,86 +532,107 @@ export class EosEngine extends CurrencyEngine {
...
+              throw new Error('dfuse only supports EOS')
+            const response = await this.eosJsConfig.fetch(
+              `${server}/v0/state/key_accounts?public_key=${params[0]}`
+            )
+            if (!response.ok) {
+              throw new Error(
+                `${server} get_account failed with ${response.code}`
+              )
+            }
+            const responseJson = await response.json()
```
paullinator, line 628: Looks like a good place for a cleaner

src/eos/eosEngine.js
```diff
@@ -532,86 +532,107 @@ export class EosEngine extends CurrencyEngine {
...
+            return {
+              server,
+              result: {
+                account_name: responseJson.account_names[0]
```
paullinator, line 632: Might be worth checking that the account name is not undefined since there may be no elements in the array

paullinator (changes requested):

paullinator (approved):

PR #258, opened 2021-03-02
Add dfuse get transactions graphql query

peachbits:


samholmes (commented):
A better approach would be to transform or morph the response from dfuse into the expected results from hyperion. This would mean that we would not need to write extra/duplicate code to handle a dfuse response vs a hyperion one. This is also more scalable when we add other APIs into the mix.

src/eos/eosEngine.js
```diff
@@ -548,7 +553,70 @@ export class EosEngine extends CurrencyEngine {
...
+            // Convert txs to Hyperion
+            const actions = responseJson.data.searchTransactionsBackward.results.map(
+              tx =>
+                asHyperionTransaction({
+                  trx_id: tx.trace.id,
+                  // Remove the Z (2020-02-07T02:47:22.5Z => 2020-02-07T02:47:22.5)
+                  '@timestamp': tx.trace.block.timestamp.slice(
+                    0,
+                    tx.trace.block.timestamp.length - 1
+                  ),
```
samholmes, line 597: Don't remove the "Z". Do a bit of discovery with date parsing on hyperion responses to see if there is issues with localization of the time string.

samholmes (changes requested):

samholmes:
After change try to test as thoroughly as possible given the up-time of hyperion.

samholmes (approved):

PR #259, opened 2021-03-04
Add promiseNy

peachbits:


src/common/utils.js
```diff
@@ -140,6 +140,80 @@ function promiseAny(promises: Promise<any>[]): Promise<any> {
...
+        },
+        error => {
+          if (++failed + results.length >= promises.length) {
+            reject(error)
+          }
+        }
+      )
+    }
+  })
+}
```
samholmes, line 191: In summary, change this promiseNy utility function to accept a compareFn argument that will be used to check the equality of two results. Use a resultCountMap object (`new Map`), to keep a running count on the number of "equivalent" result objects. Once a count for a particular result has exceeded the `n` parameter, then exit the algorithm and resolve the promise.

Rought, example code:

```
promise.then(result => {
  for (const [existingResult, count] of resultCountMap.entries()) {
    if (compare(existingResult, result)) {
      resultCountMap.set(existingResult, count + 1)
    }

    if (resultCountMap.get(existingResult) >= n) {
      return resolve(result)
    }
  }
})
```

src/fio/fioEngine.js
```diff
@@ -824,6 +827,14 @@ export class FioEngine extends CurrencyEngine {
       if (!res) {
         throw new Error('Service is unavailable')
       }
+    } else if (actionName === 'getFioNames') {
+      res = await promiseNy(
+        this.currencyInfo.defaultSettings.apiUrls.map(apiUrl =>
+          timeout(this.fioApiRequest(apiUrl, actionName, params), 5000)
+        ),
+        asGetFioName,
```
samholmes, line 835: Suggested comparison function

```ts
const mapOfStringifiedResults = new Map()
// ...
(a, b) => {
    const aString = mapOfStringifiedResults.get(a) ?? JSON.stringify(asGetFioName(a))
    const bString = mapOfStringifiedResults.get(b) ?? JSON.stringify(asGetFioName(b))

    mapOfStringifiedResults.set(a, aString)
    mapOfStringifiedResults.set(b, bString)
    
    return a === b
  },
```

samholmes (commented):
@swansontec What is your opinion on my proposal?

src/common/utils.js
```diff
@@ -140,6 +140,74 @@ function promiseAny(promises: Promise<any>[]): Promise<any> {
...
+/**
+ * Waits for the promises to resolve and uses a provided checkResult function
+ * to return a key to identify the result. The returned promise resolves when
+ * n number of promises resolve to identical keys.
+ */
+async function promiseNy<T, I>(
+  promises: Promise<T>[],
+  checkResult: T => I | void,
+  n?: number = promises.length
+): Promise<T> {
```
samholmes, line 152: Remove `I` type variable in favor of `checkResult: T => string | void`; object keys can only ever be strings, so the variable here is not needed.

src/fio/fioEngine.js
```diff
@@ -824,6 +827,14 @@ export class FioEngine extends CurrencyEngine {
       if (!res) {
         throw new Error('Service is unavailable')
       }
+    } else if (actionName === 'getFioNames') {
+      res = await promiseNy(
+        this.currencyInfo.defaultSettings.apiUrls.map(apiUrl =>
+          timeout(this.fioApiRequest(apiUrl, actionName, params), 10000)
+        ),
+        fioGetNameChecker,
```
samholmes, line 835: Could `fioGetNameChecker` be inline/anonymous function instead? Might be worth while to keep things tidy by not having the extra import. Just suggestion, I'm indifferent.

src/fio/fioSchema.js
```diff
@@ -2,6 +2,32 @@
...
+})
+
+type GetFioName = $Call<typeof asGetFioName>
+
+export function fioGetNameChecker(result: GetFioName): string | void {
+  let value
+  try {
+    value = JSON.stringify(asGetFioName(result))
+  } catch (e) {}
+  return value
```
samholmes, line 28: No need for the `value` var:

```ts
try {
  return JSON.stringify(asgetFioName(result))
} catch (e) {}
```

Also, I'm wary of the exception handling here. Are we okay with silencing all errors here; both stringify exceptions and the cleaner exception?

samholmes (changes requested):
Although the implementation of promiseN is not exactly the proposal I had, I think this approach will work and I can see some benefits (simpler/more straightforward). Aside, couple of minor changes and then I'll approve.

samholmes (approved):
Approved. Run `git rebase -i --autosquash <base-commit>`, force-push, merge, and then it's done. ✅

PR #263, opened 2021-03-25
Changed logging level in fioEngine.

andreyvEze:


src/fio/fioEngine.js
```diff
@@ -841,7 +842,7 @@ export class FioEngine extends CurrencyEngine {
       const { balance } = await this.multicastServers('getFioBalance')
       nativeAmount = balance + ''
     } catch (e) {
-      this.log.error('checkAccountInnerLoop error: ' + e)
+      this.log.warn('checkAccountInnerLoop error: ' + e)
```
thehobbit85, line 845: This should probably be a "log.info" since you know how to handle the error

thehobbit85 (changes requested):
Small change

thehobbit85 (approved):

PR #267, opened 2021-03-30
Use https://content.edge.app/ for image files

thehobbit85:


thehobbit85 (commented):
reuse the constant in utilities instead of having it in multiple files.

PR #268, opened 2021-03-30
wip: If the user specifies a nonce, use that

swansontec:
The nonce for RBF must be the nonce of the replaced transaction.

There are also situations when working with smart contracts where it is important to craft a sequence of transactions with increasing nonces, sign them, and then send them at once. This requires explicit nonce control.

So, if the user is providing a nonce, just use it, and don't mess around with other stuff.

### DIFF
--- a/src/ethereum/ethEngine.js
+++ b/src/ethereum/ethEngine.js
@@ -555,7 +555,6 @@ export class EthereumEngine extends CurrencyEngine {
     }
 
     // Nonce:
-
     const nonceArg: string = otherParams.nonceArg
     let nonce: string = nonceArg
     if (!nonce) {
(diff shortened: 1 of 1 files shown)


src/ethereum/ethEngine.js
```diff
@@ -588,11 +587,7 @@ export class EthereumEngine extends CurrencyEngine {
           throw e
         }
       } else {
-        nonce = this.walletLocalData.otherData.nextNonce
-        this.walletLocalData.otherData.unconfirmedNextNonce = bns.add(
-          this.walletLocalData.otherData.nextNonce,
-          '1'
-        )
+        nonce = otherParams.nonceArg
```
samholmes, line 590: Isn't this a noop because of lines 558 and 559? This is a clue to me that perhaps the issue with the code changed in this WIP PR.

samholmes (changes requested):
I'll need a bit more context of the issue to understand what the proposed fix is. As it stands, I cannot see how this change fixes anything, but that could be due to my lack of context.

PR #274, opened 2021-04-15
Added error when XRP destination tag exceed 9 characters

amanalili:
NOTE: should merge with https://github.com/EdgeApp/edge-react-gui/pull/2541

Sidenote: Will be better if there is a way to export the `UniqueIdentifierExceedsLengthError` like in the edge-core-js

### DIFF
--- a/src/pluginError.js
+++ b/src/pluginError.js
@@ -1,0 +1,32 @@
+// @flow
+
+export const pluginErrorCodes = [400, 403, 404]
+export const pluginErrorName = {
+  XRP_ERROR: 'XrpError'
+}
+
+export class PluginError extends Error {
+  list: Array<{ field: string, message: string }>
+  labelCode: string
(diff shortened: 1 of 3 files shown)


thehobbit85 (changes requested):
The core can export the error objects directly because we have access to the core's code inside the app, but you don't have the same access to the plugin's code so this is the wrong way to 'export' error codes from the plugin.

A better implementation would be like it's done on Fio:
https://github.com/EdgeApp/edge-currency-accountbased/blob/master/src/fio/fioError.js
https://github.com/EdgeApp/edge-currency-accountbased/blob/master/src/fio/fioInfo.js#L41

amanalili:
^Agreed, updated the PR to have a generic Error class like the fioError

thehobbit85 (approved):

PR #276, opened 2021-04-17
Reduce Fio logs verbosity

thehobbit85:


src/fio/fioEngine.js
```diff
@@ -719,7 +719,7 @@ export class FioEngine extends CurrencyEngine {
     } catch (e) {
       // handle FIO API error
       if (e.errorCode && fioApiErrorCodes.indexOf(e.errorCode) > -1) {
-        this.log.error(
+        this.log(
```
paullinator, line 722: This looks like an error in transaction creation which should not be silenced

src/fio/fioEngine.js
```diff
@@ -736,7 +736,7 @@ export class FioEngine extends CurrencyEngine {
         }
         throw e
       } else {
-        this.log.error(
+        this.log(
```
paullinator, line 739: This looks like an error in transaction creation which should not be silenced

paullinator (changes requested):

thehobbit85:
I agree that it's an error in transaction and it should not be silenced, but, there are some inconsistencies between the result type we expect from the server to what we actually get, which cause this error to be triggered far more than it should. I do plan to talk to Andrew about a more permanent fix, but I thought that this is good enough until the next release.

thehobbit85:
And the comment I wrote applies for both logs

thehobbit85:
Got a verbal approval from Paul

PR #279, opened 2021-04-30
Remove allowance transaction filtering from addTransaction

samholmes:


swansontec (approved):
You have my enthusiastic approval.

PR #302, opened 2021-06-22
Fix fio wallet transactions time

axenov-it:


axenov-it:
Fixed different time error in transactions of fio wallet  list

FIO-1
FIO-2

thehobbit85 (approved):

PR #310, opened 2021-07-22
Fix blockbook urls

peachbits:


samholmes (approved):
Add `https://eth2.trezor.io` also

PR #315, opened 2021-08-11
EIP-681 Support for Ethereum Plugin URI Parsing

samholmes:


src/common/plugin.js
```diff
@@ -76,23 +75,24 @@ export class CurrencyPlugin {
       throw new Error('InvalidUriError') // possibly scanning wrong crypto type
     }
 
-    if (parsedUri.host) {
-      address = parsedUri.host
-    } else if (parsedUri.pathname) {
-      address = parsedUri.pathname
-    } else {
+    // If no host and no path, then it's not a valid URI
+    if (parsedUri.host === '' && parsedUri.pathname === '') {
```
thehobbit85, line 79: You also need to check for "nullish" values here.

src/common/utils.js
```diff
@@ -352,6 +352,16 @@ export function cleanTxLogs(tx: EdgeTransaction) {
   return JSON.stringify(asCleanTxLogs(tx), null, 2)
 }
 
+// Convert number strings in scientific notation to decimal notation using biggystring
+export function biggyScience(num: string): string {
```
thehobbit85, line 356: You don't need this function, you can just do:
`parseFloat('1e10').toFixed(18)`

src/ethereum/ethPlugin.js
```diff
@@ -219,7 +220,79 @@ export class EthereumPlugin extends CurrencyPlugin {
...
+      // Handle contract function invocations
+      // This is a very important measure to prevent accidental payment to contract addresses
+      switch (functionName) {
+        // ERC-20 token transfer
+        case 'transfer': {
+          const publicAddress = parameters.address ?? ''
+          const contractAddress = targetAddress ?? ''
+          const nativeAmount =
+            parameters.uint256 != null
+              ? biggyScience(parameters.uint256)
```
thehobbit85, line 239: `parseFloat(parameters.uint256).toFixed(DECIMAL_NUMBERS)`

thehobbit85 (changes requested):

src/common/utils.js
samholmes: We need a big num library because `Number.MAX_SAFE_INTEGER.toString().length < 18`.

src/ethereum/ethPlugin.js
samholmes: There may not be enough precision in IEEE 754

PR #319, opened 2021-08-18
Fix unable to send on ETC, FTM, and RSK networks

samholmes:


src/ethereum/ethTypes.js
```diff
@@ -26,12 +26,38 @@ export type EthereumInitOptions = {
...
+  blockcypherApiServers: string[],
+  blockbookServers: string[],
+  uriNetworks: string[],
+  ercTokenStandard: string,
+  chainId: number,
+  hdPathCoinType: number,
+  checkUnconfirmedTransactions: boolean,
+  iosAllowedTokens: {
+    [currencyCode: string]: true
+  },
```
swansontec, line 48: This one is duplicated.

src/ethereum/ethTypes.js
```diff
@@ -26,12 +26,38 @@ export type EthereumInitOptions = {
...
+    token: string
+  } | null,
+  amberdataRpcServers: string[],
+  amberdataApiServers: string[],
+  amberDataBlockchainId: string,
+  pluginMnemonicKeyName: string,
+  pluginRegularKeyName: string,
+  ethGasStationUrl: string | null,
+  // eslint-disable-next-line no-use-before-define
+  defaultNetworkFees: EthereumFees
```
swansontec: Can we sort these to make dupes easier to see / find?

src/ethereum/ethEngine.js
```diff
@@ -724,10 +725,9 @@ export class EthereumEngine extends CurrencyEngine {
     }
 
     // Select the chain
-    const common = new Common({
-      // EIP 155 chainId - ETH mainnet: 1, ETH ropsten: 3
-      chain: this.currencyInfo.defaultSettings.otherSettings.chainId
-    })
+    const { chainParams } = (this.currencyInfo.defaultSettings
+      .otherSettings: EthereumSettings)
```
swansontec, line 729: The colon syntax breaks VSCode pretty bad, so we typically do something like this when we need to do a cast:

```js
const otherSettings: EthereumSettings = this.currencyInfo.defaultSettings.otherSettings
const { chainParams } = otherSettings
```

swansontec (changes requested):

swansontec (approved):

PR #329, opened 2021-10-04
Always return results with the newest request first.

peachbits:
This also resolves an off-by-one error when using slice


peachbits:
Eli approved

PR #330, opened 2021-10-07
Misc cleanups

peachbits:
---
  - https://app.asana.com/0/0/1201149582774707

### DIFF
--- a/src/common/engine.js
+++ b/src/common/engine.js
@@ -341,7 +341,7 @@ export class CurrencyEngine {
     // if transaction doesn't exist in database
     if (idx === -1) {
       if (
-        // if unconfirmed spend then increment # uncofirmed spend TX's
+        // if unconfirmed spend then increment # unconfirmed spend TX's
         this.isSpendTx(edgeTransaction) &&
         edgeTransaction.blockHeight === 0
       ) {
--- a/src/common/engine.js
+++ b/src/common/engine.js
@@ -36,7 +36,12 @@ import {
   TXID_MAP_FILE,
   WalletLocalData
 } from './types.js'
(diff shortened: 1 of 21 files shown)


src/common/utils.js
```diff
@@ -390,6 +390,15 @@ function getFetchCors(opts: EdgeCorePluginOptions): Function {
   }
 }
 
+export function sāf(e?: Error): string {
```
thehobbit85, line 393: wtf? why? who? when? what? sāf?

src/common/engine.js
```diff
@@ -268,7 +268,7 @@ export class CurrencyEngine {
         )
       } catch (e) {
         this.log.error(
-          'Error writing to localDataStore. Engine not started:' + err
+          'Error writing to localDataStore. Engine not started:' + sāf(e)
```
thehobbit85, line 271: Instead of wrapping the error function with the "sāf" call every single time, in the constructor, instead of line 82, you should do something like this:

```js
this.log = {
   error: (message, e) => opts.log.error(message + sāf(e))
   warn: ...
   ....
 }
 ```

File: src/ethereum/ethNetwork.js [hunk 3]
```diff
@@ -1179,6 +1171,29 @@ export class EthereumNetwork {
     return await asyncWaterfall(funcs)
   }
 
+  checkAmberdata(func: string, ...args: any[]): void => Promise<any> {
```
thehobbit85, line 1174: I want to discuss this

thehobbit85 (changes requested):

peachbits:
closed in favor of https://github.com/EdgeApp/edge-currency-accountbased/pull/361

PR #347, opened 2021-12-30
Append addBundledTransactions case to fioAction

andreyvEze:
---
  - https://app.asana.com/0/0/1201602422919538

### DIFF
--- a/src/fio/fioEngine.js
+++ b/src/fio/fioEngine.js
@@ -269,6 +269,31 @@ export class FioEngine extends CurrencyEngine {
             }
             return res
           }
+          case 'addBundledTransactions': {
+            const res = await this.multicastServers(actionName, params)
+            const fioAddress = this.otherData.fioAddresses.find(
+              ({ name }) => name === params.fioAddress
+            )
+
+            if (!fioAddress) return res
+
+            try {
+              const { fio_addresses: fetchedFioAddresses } =
+                await this.multicastServers('getFioAddresses', {
+                  fioPublicKey: this.walletInfo.keys.publicKey
+                })
+              const updatedBundlesValue = fetchedFioAddresses.find(
+                ({ fio_address: name }) => name === params.fioAddress
+              ).remaining_bundled_tx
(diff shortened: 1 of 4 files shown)


src/fio/fioEngine.js
```diff
@@ -269,6 +269,31 @@ export class FioEngine extends CurrencyEngine {
...
+                ({ fio_address: name }) => name === params.fioAddress
+              ).remaining_bundled_tx
+              fioAddress.bundles = updatedBundlesValue
+              this.localDataDirty()
+              return { bundles: updatedBundlesValue, ...res }
+            } catch (e) {
+              //
+            }
+
+            return { bundles: fioAddress.bundles, ...res }
```
thehobbit85, line 295: Why do you need to refetch the 'getFioAddresses' data. Don't you know how many bundles you added from the previous call already?

thehobbit85 (changes requested):

src/fio/fioEngine.js
andreyvEze: no, I don't know this information. let me check if this a default value

src/fio/fioEngine.js
```diff
@@ -269,6 +269,31 @@ export class FioEngine extends CurrencyEngine {
             }
             return res
           }
+          case 'addBundledTransactions': {
+            const res = await this.multicastServers(actionName, params)
+            const fioAddress = this.otherData.fioAddresses.find(
+              ({ name }) => name === params.fioAddress
+            )
+
+            if (!fioAddress) return res
```
andreyvEze, line 278: put this before multicast and of false - throw

thehobbit85: Also, don't test for "null" value implicitly.
If you want to check if something is null or undefined, use the correct:
`            if (fioAddress == nulll) return res`

thehobbit85:
Replaced by the clean PR - https://github.com/EdgeApp/edge-currency-accountbased/pull/352

PR #348, opened 2021-12-30
Jon/fix/wcInvalidUri

Jon-edge:
wc was not included as a valid uri network for Polygon.

---
  - https://app.asana.com/0/0/1201354677026256


src/ethereum/ethPlugin.js
```diff
@@ -153,6 +153,9 @@ export class EthereumPlugin extends CurrencyPlugin {
       }
     )
 
+    // By default, all EVM clones should be WalletConnect compatible.
```
thehobbit85, line 156: Move up to where you initialize the param

thehobbit85 (changes requested):

thehobbit85 (approved):

PR #313, opened 2021-08-05
Add balance data in fio engine

andreyvEze:
FIOSDK version need to be updated
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1200692419869086
  - https://app.asana.com/0/0/1200759268116595

PR #363, opened 2022-01-25
FIO Staking

samholmes:
Replaces https://github.com/EdgeApp/edge-currency-accountbased/pull/313

FIOSDK version need to be updated
---
  - https://app.asana.com/0/0/1200692419869086
  - https://app.asana.com/0/0/1200759268116595

### DIFF
--- a/src/fio/fioEngine.js
+++ b/src/fio/fioEngine.js
@@ -1286,6 +1286,31 @@ export class FioEngine extends CurrencyEngine {
       params.newOwnerKey = publicAddress // todo: move this to the gui
     }
 
+    if (name === ACTIONS.stakeFioTokens) {
+      params.amount = quantity
+    }
+
+    if (name === ACTIONS.unStakeFioTokens) {
+      const stakedBalance =
+        this.walletLocalData.totalBalances[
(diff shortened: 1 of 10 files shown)


samholmes:
I cleaned up the PR's commits. The only thing I want to discuss is the refactor commit for makeSpend. Seems significant enough to spend a little bit of time to lookout for regressions.

samholmes:
We resolved the concerns about the refactor commit with Andrey.

src/fio/fioEngine.js
```diff
@@ -902,6 +903,13 @@ export class FioEngine extends CurrencyEngine {
       switch (actionName) {
         case 'getChainInfo':
           res = await fioSdk.transactions.getChainInfo()
+          break
+        case 'getFioBalance':
+          res = await fioSdk.genericAction(actionName, params)
+          asGetFioBalanceResponse(res)
+          if (res.balance != null && res.balance < 0)
```
samholmes, line 910: Remove the extra null-check.

samholmes: Write comment:

```js
// Assert that balance should never be negative just to be extra safe
```

PR #364, opened 2022-01-27
FIO staking status amount fix

andreyvEze:


samholmes (changes requested):
Let's discuss together the fixes

src/fio/fioEngine.js
```diff
@@ -1440,7 +1445,10 @@ export class FioEngine extends CurrencyEngine {
       this.recentFioFee = { publicAddress, fee }
     }
 
-    if (bns.gt(bns.add(quantity, `${fee}`), nativeBalance)) {
+    if (
```
samholmes, line 1457: Add a comment explaining why we don't need to check the available balance for unstaked action (because that's handled separately below).

src/fio/fioInfo.js
```diff
@@ -29,8 +28,7 @@ const defaultSettings: any = {
   historyNodeUrls: [
     'https://fio.greymass.com/v1/',
     'https://fio.greymass.com/v1/',
-    'https://fio.eosphere.io/v1/',
-    'https://fio.eossweden.org/v1/'
+    'https://fio.eosphere.io/v1/'
```
samholmes, line 31: Make this a fixup commit if that makes more sense

src/fio/fioEngine.js
```diff
@@ -477,6 +477,15 @@ export class FioEngine extends CurrencyEngine {
     blockTime: string,
     txId: string
   ): void {
+    if (
```
samholmes, line 480: Add comment:

```js
// Might not be necessary, but better to be safe than sorry
```

src/fio/fioInfo.js
samholmes: This is a change on its own. No fixup

PR #365, opened 2022-01-27
FIO staking balances refactor

andreyvEze:


src/fio/fioEngine.js
```diff
@@ -1101,14 +1099,12 @@ export class FioEngine extends CurrencyEngine {
...
+      const { balance, available, staked, srps, roe } = await this.multicastServers(
+        'getFioBalance'
+      )
       nativeAmount = balance != null ? String(balance) : '0'
-      balances.available = available != null ? String(available) : '0'
       balances.staked = staked != null ? String(staked) : '0'
-      balances.locked = bns.sub(nativeAmount, balances.available)
-      balances.accrued =
-        srps != null && roe != null ? bns.mul(String(srps), roe) : '0'
+      balances.locked = bns.sub(nativeAmount, available + '')
```
samholmes, line 1107: From the developer guide about `available`:

> Amount of FIO Tokens (SUFs) which can be spent. available = balance - staked - locked.

So shouldn't locked be:

```js
balances.locked = bns.sub(bns.sub(nativeAmount, String(available)), String(staked))
```

Also we should clean the data coming from multicastServers so that we can always expect number values for these fields (as they're specified in the FIO developer guide.

samholmes: Yesterday this was my solution to clean the network data:

```js
        case 'getFioBalance':
          res = await fioSdk.genericAction(actionName, params)
          res = asObject({
            available: asNumber,
            balance: asNumber,
            roe: asString,
            srps: asNumber,
            staked: asNumber
          })(res)
          break
```

Add that to the `fioApiRequest` implementation above the `default` case for `actionName`

samholmes (changes requested):

src/fio/fioEngine.js
samholmes: 1. We include staked amount in our version of what we call "locked". So we must include `staked` in the locked amount. Though, we should cleanup using `String` constructor or `toString()` method.
2. This is resolved with another commit in the main accountbased PR

src/fio/fioInfo.js
```diff
@@ -52,9 +52,7 @@ const defaultSettings: any = {
   fioRequestsTypes: FIO_REQUESTS_TYPES,
   balanceCurrencyCodes: {
     staked: 'FIO:STAKED',
-    available: 'FIO:AVAILABLE',
-    locked: 'FIO:LOCKED',
-    accrued: 'FIO:ACCRUED'
+    locked: 'FIO:LOCKED'
```
samholmes, line 55: Note: We kept staked because of the cost to refactor would be too high for today.

samholmes (approved):

PR #372, opened 2022-01-31
FIO Staking process tx fix

andreyvEze:


samholmes:
This PR fixes two things:

1. After transfer token transactions were sent, the fee processed twice doubling the calculated fee amount for the transaction.
2. Remove processing of transaction for 'FIO:LOCKED' currencyCode. This fixes an bug which is: a transaction notification is shown with an empty currency code in the GUI.

samholmes (approved):

PR #376, opened 2022-02-04
Add Solana

peachbits:
---
  - https://app.asana.com/0/0/1201425242712354

### DIFF
--- a/src/solana/solanaEngine.js
+++ b/src/solana/solanaEngine.js
@@ -14,7 +14,7 @@ import {
 
 import { CurrencyEngine } from '../common/engine.js'
 import { asyncWaterfall, cleanTxLogs, getOtherParams } from '../common/utils.js'
-import { SolanaPlugin } from './solanaPlugin.js'
+import { createKeyPair, SolanaPlugin } from './solanaPlugin.js'
 import {
   type RpcGetTransaction,
   type RpcSignatureForAddress,
@@ -39,7 +39,6 @@ const TRANSACTION_POLL_MILLISECONDS = 3000
 
 export class SolanaEngine extends CurrencyEngine {
   keypair: Keypair
-  createKeyPair: (seed: string) => Promise<Keypair>
   base58PublicKey: string
   feePerSignature: string
   recentBlockhash: string
@@ -55,7 +54,6 @@ export class SolanaEngine extends CurrencyEngine {
     fetchCors: EdgeFetchFunction
   ) {
     super(currencyPlugin, walletInfo, opts)
-    this.createKeyPair = currencyPlugin.createKeyPair
     this.chainCode = currencyPlugin.currencyInfo.currencyCode
     this.fetchCors = fetchCors
     this.feePerSignature = '5000'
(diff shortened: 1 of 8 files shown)


src/solana/solanaEngine.js
```diff
@@ -0,0 +1,390 @@
...
+  settings: SolanaSettings
+
+  constructor(
+    currencyPlugin: SolanaPlugin,
+    walletInfo: EdgeWalletInfo,
+    opts: any, // EdgeCurrencyEngineOptions
+    fetchCors: EdgeFetchFunction
+  ) {
+    super(currencyPlugin, walletInfo, opts)
+    this.keypair = currencyPlugin.createKeyPair(
```
paullinator, line 50: This should not happen on every engine creation. createKeyPair uses slow cryptography. This should only be done when needed to sign a transaction. The public key should be created at the time you import or create the private key and be returned in the keys object.

src/solana/solanaEngine.js
```diff
@@ -0,0 +1,390 @@
...
+      this.updateBalance(this.chainCode, balance.value.toString())
+    } catch (e) {
+      // fetching of account balances for uninitiated accounts throws error
+      if (
+        this.tokenCheckTransactionsStatus[this.chainCode] === 1 &&
+        this.transactionList[this.chainCode].length === 0
+      ) {
+        this.updateBalance(this.chainCode, '0')
+      }
+      this.error(`Error checking ${this.chainCode} address balance`, e)
```
paullinator, line 119: This isn't really an error if the account is uninitiated. Can you differentiate a true error and only log that?

src/solana/solanaEngine.js
```diff
@@ -0,0 +1,390 @@
...
+            commitment: this.settings.commitment
+          }
+        ]
+        const response: RpcSignatureForAddress[] = await this.fetchRpc(
+          'getSignaturesForAddress',
+          params
+        )
+
+        // remove failed transactions from list
+        txids = txids.concat(
```
paullinator, line 201: Don't remove failed transactions if they incurred a fee to the user's wallet. Just like ETH, those are valid transactions, and the fee should still be shown. Set nativeamount to 0 and add it.

src/solana/solanaEngine.js
```diff
@@ -0,0 +1,390 @@
...
+      return
+    }
+
+    if (txids.length === 0) {
+      this.tokenCheckTransactionsStatus[this.chainCode] = 1
+      this.updateOnAddressesChecked()
+      return
+    }
+
+    let newestTxid = 0
```
paullinator, line 219: Should be called `newestTxIndex`. It's not a txid

src/solana/solanaEngine.js
```diff
@@ -0,0 +1,390 @@
...
+      } catch (e) {
+        // Note the oldest failed tx query so we try again next loop
+        newestTxid = i
+      }
+    }
+
+    // Don't update the newestTxid if the txids array length is length 1 or the oldest failed query is the end of the array.
+    // The previously saved value is the best to use in those cases
+    if (txids.length > 1 && txids.length > newestTxid + 1)
+      this.otherData.newestTxid = txids[newestTxid + 1].signature
```
paullinator, line 244: `newestTxid` has an inconsistent type. sometimes a number, sometimes a string. make it consistent.

paullinator (changes requested):

src/solana/solanaPlugin.js
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  async parseUri(
+    uri: string,
+    currencyCode?: string,
+    customTokens?: EdgeMetaToken[]
+  ): Promise<EdgeParsedUri> {
+    const networks = { [this.pluginId]: true }
+
+    const { parsedUri, edgeParsedUri } = this.parseUriCommon(
```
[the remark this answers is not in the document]
peachbits: Yes it's standard bip21 https://github.com/solana-labs/solana-pay/blob/master/SPEC.md we'll need special handling for spl-token once we add token support

paullinator (approved):

PR #390, opened 2022-02-25
Set the 'canReplaceByFee' to true for all the EVM chains

thehobbit85:


peachbits (commented):
Add the upgrade to edge-core-js v0.19.10

PR #404, opened 2022-05-04
Fix initial local networkFees assignment

peachbits:
---
  - https://app.asana.com/0/0/1202219160560388

### DIFF
--- a/src/ethereum/fees/ethMiningFees.js
+++ b/src/ethereum/fees/ethMiningFees.js
@@ -34,10 +34,14 @@ export function calcMiningFee(
     const { customNetworkFee } = spendInfo || {}
     if (spendInfo.networkFeeOption === ES_FEE_CUSTOM && customNetworkFee) {
       const { gasLimit, gasPrice } = customNetworkFee
-      const {
-        gasLimit: { minGasLimit },
-        gasPrice: { minGasPrice }
-      } = currencyInfo.defaultSettings.otherSettings.defaultNetworkFees.default
+      const minGasLimit =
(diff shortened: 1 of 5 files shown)


thehobbit85 (changes requested):
1. Use the ‘minFee’ from the  ‘currencyEngine.otherData.networkFees’ if exists, (if not default to the currencyInfo)
1. ‘DeepMerge’ the result from the info server into the local fees -   https://github.com/EdgeApp/edge-core-js/blob/1252ff06348925342a55c62b38d8bbd2e12117ca/src/util/util.js#L16

src/ethereum/fees/ethMiningFees.js
```diff
@@ -34,10 +34,14 @@ export function calcMiningFee(
     const { customNetworkFee } = spendInfo || {}
     if (spendInfo.networkFeeOption === ES_FEE_CUSTOM && customNetworkFee) {
       const { gasLimit, gasPrice } = customNetworkFee
-      const {
-        gasLimit: { minGasLimit },
-        gasPrice: { minGasPrice }
-      } = currencyInfo.defaultSettings.otherSettings.defaultNetworkFees.default
+      const minGasLimit =
+        networkFees.default.gasLimit.minGasLimit ??
```
thehobbit85, line 38: Shouldn't you use more question marks incase something is undefined like this:
`networkFees?.default?.gasLimit?.minGasLimit` ?

thehobbit85 (changes requested):

thehobbit85 (approved):

PR #411, opened 2022-05-20
Sam/eth testnets

samholmes:
---
  - https://app.asana.com/0/0/1202303706365551


thehobbit85 (changes requested):
Can you rename the plugins to 'ethereumRinkeby', 'ethereumRopsten', etc..?
Other then that it looks ok.

samholmes:
> Can you rename the plugins to 'ethereumRinkeby', 'ethereumRopsten', etc..? Other then that it looks ok.

Based on what's taken from this site as a resource of reference: https://chainlist.org/ it would seem appropriate to omit "ethereum" from the name

thehobbit85 (approved):

PR #415, opened 2022-06-05
Remove queries for internal transactions

paullinator:
Internal transactions are by definition between contracts and don't affect the end balance of an EVM account. These queries were overwriting regular transactions with the same txid causing incorrect transaction amounts.

### DIFF
--- a/src/ethereum/ethNetwork.js
+++ b/src/ethereum/ethNetwork.js
@@ -35,13 +35,11 @@ import {
   type CheckTokenBalBlockchair,
   type EthereumSettings,
   type EthereumTxOtherParams,
-  type EvmScanInternalTransaction,
   type EvmScanTransaction,
   type FetchGetAlethio,
   type FetchGetAmberdataApiResponse,
   type RpcResultString,
   asAlethioAccountsTokenTransfer,
-  asAmberdataAccountsFuncs,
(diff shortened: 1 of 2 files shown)


src/ethereum/ethNetwork.js
```diff
@@ -243,6 +242,12 @@ export class EthereumNetwork {
     } else {
       throw new Error('Invalid transaction result format')
     }
+    if (
```
peachbits, line 245: Remove this debugging log

peachbits (changes requested):

src/ethereum/ethNetwork.js
paullinator: good catch

src/ethereum/ethNetwork.js
```diff
@@ -41,7 +41,6 @@ import {
   type FetchGetAmberdataApiResponse,
   type RpcResultString,
   asAlethioAccountsTokenTransfer,
-  asAmberdataAccountsFuncs,
```
peachbits, line 44: Duh I didn't catch these wouldn't be used anymore. Can you remove them the ethTypes too?

peachbits (changes requested):

peachbits (approved):

PR #421, opened 2022-07-05
Build feeRateUsed for Ethereum receiving txns

GoodluckH:
Fetches gas related variables to build the `feeRateUsed` object for GUI
display purpose.

Note that the `gasUsed` has no corresponding key string in the GUI 
repo. We might need to open a task for creating localized strings for
`gasUsed`.

Simulator Screen Shot - iPhone 13 - 2022-07-05 at 13 33 34

---
  - https://app.asana.com/0/0/1200984756686295

### DIFF
--- a/src/ethereum/ethNetwork.js
+++ b/src/ethereum/ethNetwork.js
@@ -22,6 +22,7 @@ import {
   snooze,
   validateObject
 } from '../common/utils'
+import { WEI_MULTIPLIER } from './ethConsts.js'
 import { EthereumEngine } from './ethEngine'
 import { EtherscanGetAccountNonce, EtherscanGetBlockHeight } from './ethSchema'
 import {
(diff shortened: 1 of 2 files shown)


src/ethereum/ethNetwork.js
```diff
@@ -248,11 +249,29 @@ export class EthereumNetwork {
...
+          WEI_MULTIPLIER.toString(),
+          WEI_MULTIPLIER.toString().length - 1
+        ),
+        gasUsed: tx.gasUsed,
+        gasLimit: tx.gas
+      }
+    } catch (e) {
+      console.log(`Failed to construct feeRateUssed: ${e}`)
+    }
+
```
thehobbit85, line 267: Don't repeat yourself - https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

Turn this code into a function: `(gasPrice: string, gasUsed: string, gasLimit: string) => feeRateUsed`

thehobbit85 (changes requested):

thehobbit85 (changes requested):
1. The last commit should have been a fixup commit and not a new commit. Remember that you should try to avoid having multiple commits that change the same code over and over. That's why we use fixups.
2. I'm pretty sure this function can be used in 1 or more places in the code, can you replace those as well?

thehobbit85 (approved):

PR #422, opened 2022-07-07
Add Polkadot

peachbits:
This PR includes little bug fixes and an upgrade to Webpack v5

---
  - https://app.asana.com/0/0/1202564031135083


src/common/engine.js
```diff
@@ -47,8 +47,10 @@ const SAVE_DATASTORE_MILLISECONDS = 10000
 const MAX_TRANSACTIONS = 1000
 const DROPPED_TX_TIME_GAP = 3600 * 24 // 1 Day
 
-export class CurrencyEngine {
-  currencyPlugin: CurrencyPlugin
+type CurrencyPluginType<T> = T & CurrencyPlugin
+
+export class CurrencyEngine<T> {
```
thehobbit85, line 52: It can be cleaner if you do this:
1. No need to pass T to the type
2. Change line 53 to: ```currencyPlugin: CurrencyPluginType<any>```
3. Change the constructor signature to this:
```js
constructor<T>(
    currencyPlugin: CurrencyPluginType<T>,
    walletInfo: EdgeWalletInfo,
    opts: EdgeCurrencyEngineOptions
  ) 
  ```

thehobbit85: Then you don't need to pass the type in all the other files and this:
```export class EosEngine extends CurrencyEngine<EosPlugin> {```
Goes back to this:
```export class EosEngine extends CurrencyEngine {```

AND you still get the typing :)

thehobbit85 (changes requested):
Very partial review but there are already some comments to work on

paullinator (changes requested):

src/common/engine.js
peachbits: I tried but I'm not able to make this work with the test above. The types aren't accessible in other files and sucrase can't build with a type on the constructor "Error transforming src/common/engine.js: Expected context ID on open-paren starting constructor params."

PR #428, opened 2022-07-29
NaN and sent amount denom fixes

peachbits:
---
  - https://app.asana.com/0/0/1202681401455011


peachbits:
Approved OTS by @samholmes

PR #434, opened 2022-08-22
Allow specifying only gasPrice or gasLimit for custom fees

paullinator:
Necessary for services like Thorchain which require a specific gasPrice but don't specify a gasLimit

---
  - https://app.asana.com/0/0/1202834837562700


Jon-edge (approved):

paullinator:
@Jon-edge approved fixup commit

PR #436, opened 2022-08-27
Fix return types for engine.js:makeSpend

paullinator:
Rename engine.js:makeSpend to makeSpendCheck since it has a different return signature than the asset specific makeSpend.

This exposed Flow errors in every asset engine that then needed to be fixed.

### DIFF
--- a/src/binance/bnbEngine.js
+++ b/src/binance/bnbEngine.js
@@ -9,7 +9,8 @@ import {
   type EdgeSpendInfo,
   type EdgeTransaction,
   type EdgeWalletInfo,
-  InsufficientFundsError
+  InsufficientFundsError,
+  NoAmountSpecifiedError
 } from 'edge-core-js/types'
 
 import { CurrencyEngine } from '../common/engine.js'
@@ -416,10 +417,16 @@ export class BinanceEngine extends CurrencyEngine<BinancePlugin> {
   }
 
   async makeSpend(edgeSpendInfoIn: EdgeSpendInfo) {
-    const { edgeSpendInfo, currencyCode } = super.makeSpend(edgeSpendInfoIn)
+    const { edgeSpendInfo, currencyCode } = this.makeSpendCheck(edgeSpendInfoIn)
 
     const spendTarget = edgeSpendInfo.spendTargets[0]
-    const publicAddress = spendTarget.publicAddress
+    const { publicAddress } = spendTarget
+    let { nativeAmount } = spendTarget
+
+    if (publicAddress == null)
+      throw new Error('makeSpend Missing publicAddress')
+    if (nativeAmount == null) throw new NoAmountSpecifiedError()
+
     const data =
       spendTarget.otherParams != null ? spendTarget.otherParams.data : undefined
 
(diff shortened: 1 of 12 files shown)


src/common/engine.js
```diff
@@ -928,7 +928,13 @@ export class CurrencyEngine<T> {
     return dataDump
   }
 
-  makeSpend(edgeSpendInfo: EdgeSpendInfo): Object {
+  makeSpendCheck(edgeSpendInfo: EdgeSpendInfo): {
+    edgeSpendInfo: EdgeSpendInfo,
+    nativeBalance: string,
+    currencyCode: string,
+    denom: EdgeDenomination,
+    skipChecks: boolean
```
samholmes, line 936: I think we shouldn't return `skipChecks` here because it is not being mutated by `makeSpendCheck`.

src/binance/bnbEngine.js
```diff
@@ -416,10 +417,16 @@ export class BinanceEngine extends CurrencyEngine<BinancePlugin> {
...
-    const { edgeSpendInfo, currencyCode } = super.makeSpend(edgeSpendInfoIn)
+    const { edgeSpendInfo, currencyCode } = this.makeSpendCheck(edgeSpendInfoIn)
 
     const spendTarget = edgeSpendInfo.spendTargets[0]
-    const publicAddress = spendTarget.publicAddress
+    const { publicAddress } = spendTarget
+    let { nativeAmount } = spendTarget
+
+    if (publicAddress == null)
+      throw new Error('makeSpend Missing publicAddress')
```
samholmes, line 427: Peculiar casing; optional change: `'Missing publicAddress in makeSpend'`

src/zcash/zecEngine.js
```diff
@@ -318,6 +317,14 @@ export class ZcashEngine extends CurrencyEngine<ZcashPlugin> {
...
     // Create the unsigned EdgeTransaction
 
+    const spendTargets = edgeSpendInfo.spendTargets.map(si => ({
+      uniqueIdentifier: si.uniqueIdentifier,
+      memo: si.memo,
+      nativeAmount: si.nativeAmount ?? '0',
+      currencyCode,
+      publicAddress
+    }))
+
```
samholmes, line 327: What is happening here? This seems like it needs either to be in another commit so we can document in the message the change, or add a comment to explain the need for this change.

samholmes (changes requested):
One change request with two optional changes

src/zcash/zecEngine.js
paullinator: Damn. This was an unrelated change that also snuck in. I'll put this in a new PR. Too many simultaneous changes at the same time.

paullinator: This needs to be the same commit or you'll get Flow errors. the SpendInfo.spendTargets is a different shape than EdgeTransaction.spendTargets. This line does the conversion. When super.makeSpend returned the edgeSpendInfo as type `Object` it obscured the bug.

src/common/engine.js
paullinator: That wasn't my change, but I do think there's value since `makeSpendCheck` sort of mutates it. It defaults to `false` if undefined. We'd have to mirror that change in each makeSpend otherwise.

samholmes (approved):

PR #438, opened 2022-09-04
Restore internal txs

paullinator:
This is required to see transactions where contracts send ETH (or any mainnet coin) back to a user's wallet.
Also remove use of other APIs for tx query other than EVM scan APIs since others don't support internal txs

---
  - https://app.asana.com/0/0/1202903392787357


src/ethereum/ethNetwork.js
```diff
@@ -1343,401 +1265,6 @@ export class EthereumNetwork {
...
-      }
-      if (amberdataTxs.length === 0) {
-        break
-      }
-      page++
-    }
-    return allTransactions
-  }
-
-  async checkTxsAmberdata(
```
peachbits, line 1518: Amberdata can query internal txs so why is it removed?

src/ethereum/ethNetwork.js
```diff
@@ -237,13 +242,16 @@ export class EthereumNetwork {
     )
   }
 
-  processEvmScanTransaction(tx: EvmScanTransaction, currencyCode: string) {
+  processEvmScanTransaction(
```
peachbits, line 227: I can't tell from this PR if the original issue of incorrect transaction amounts is resolved or if we're just reintroducing the problem.

src/ethereum/ethNetwork.js
```diff
@@ -1447,451 +1265,6 @@ export class EthereumNetwork {
...
-        [`${this.currencyInfo.currencyCode}`]: {
-          blockHeight: startBlock,
-          edgeTransactions: [...allTxsRegular, ...allTxsInternal]
-        }
-      },
-      server: 'amberdata'
-    }
-  }
-
-  async checkTxsBlockbook(
```
peachbits, line 1539: This function is called checkTxs but it does more than that because the single api call can fill an entire ethereum network update. You can keep it still use it to get the nonce and balances but remove the tokenTxs if they aren't accurate for internal txs.

src/ethereum/ethNetwork.js
```diff
@@ -1447,451 +1265,6 @@ export class EthereumNetwork {
     }
   }
 
-  async checkTxsAlethio(
```
peachbits, line 1346: I'm less picky about this being removed (again) because I don't think the service even exists anymore.

peachbits (changes requested):

src/ethereum/ethNetwork.js
paullinator: I could not reproduce the incorrect amounts with only Etherscan. I can’t verify Amberdata because our api key is maxed out. I can only assume amberdata may have been returning the bad info.

src/ethereum/ethNetwork.js
```diff
@@ -1587,7 +1643,8 @@ export class EthereumNetwork {
     }
     if (blockbookServers.length > 0) {
       blockheight.push(this.checkBlockHeightBlockbook)
-      // txs.push(this.checkTxsBlockbook)
+      tokenBal.push(this.checkAddressBlockbook)
```
peachbits, line 1646: Delete the tokenBal line. The token balance query is a loop over each enabled currency code which would be lot a redundant fetches for the exact same data. Leaving it in the nonce loop returns the same data without a loop. This is where the original function that grabbed txs should have been the whole time

peachbits (changes requested):

peachbits (approved):

PR #441, opened 2022-09-15
Sam/unconfirmed nonce fix

samholmes:
---
  - https://app.asana.com/0/0/1202958974812437


src/ethereum/ethEngine.js
```diff
@@ -949,6 +949,8 @@ export class EthereumEngine extends CurrencyEngine<EthereumPlugin> {
       otherParams // otherParams
     }
 
+    this.warn(`makeSpend\n${cleanTxLogs(edgeTransaction)}`)
```
paullinator, line 952: This could get noisy and lag the system as each keystroke in the FlipInput would call makeSpend. I'd request removing this.

paullinator (changes requested):

paullinator (approved):

PR #443, opened 2022-09-17
Convert to typescript

paullinator:
---
  - https://app.asana.com/0/0/1201622044701693


src/ethereum/ethEngine.ts
```diff
@@ -986,16 +986,16 @@ export class EthereumEngine extends CurrencyEngine<EthereumPlugin> {
       // Otherwise, use the next nonce
       if (
         this.walletLocalData.numUnconfirmedSpendTxs &&
```
peachbits, line 1065: Missed sketchy null

src/fio/fioEngine.ts
```diff
@@ -1440,7 +1443,7 @@ export class FioEngine extends CurrencyEngine<FioPlugin> {
       }
     }
 
-    const { name, params }: { name: string, params: any } = otherParams.action
+    const { name, params }: { name: string; params: any } = otherParams.action
```
peachbits, line 1552: Replace ; with ,

src/stellar/stellarEngine.ts
```diff
@@ -154,8 +153,8 @@ export class StellarEngine extends CurrencyEngine<StellarPlugin> {
 
     const date: number = Date.parse(tx.created_at) / 1000
     const denom = getDenomInfo(this.currencyInfo, currencyCode)
-    if (denom && denom.multiplier) {
-      nativeAmount = bns.mul(exchangeAmount, denom.multiplier)
+    if (denom != null && denom.multiplier) {
```
peachbits, line 166: missed sketchy null

src/stellar/stellarEngine.ts
```diff
@@ -296,8 +295,8 @@ export class StellarEngine extends CurrencyEngine<StellarPlugin> {
           currencyCode = bal.asset_type
         }
         const denom = getDenomInfo(this.currencyInfo, currencyCode)
-        if (denom && denom.multiplier) {
-          const nativeAmount = bns.mul(bal.balance, denom.multiplier)
+        if (denom != null && denom.multiplier) {
```
peachbits, line 318: missed sketchy null

src/stellar/stellarEngine.ts
```diff
@@ -373,7 +372,7 @@ export class StellarEngine extends CurrencyEngine<StellarPlugin> {
     // Check if destination address is activated
     let mustCreateAccount = false
     const activated = this.activatedAccountsCache[publicAddress]
-    if (activated === false) {
+    if (!activated) {
```
peachbits, line 403: This shouldn't be changed. If activated is undefined we don't test the second conditional

src/stellar/stellarEngine.ts
```diff
@@ -385,19 +384,19 @@ export class StellarEngine extends CurrencyEngine<StellarPlugin> {
...
+    const exchangeAmount = div(nativeAmount, denom.multiplier, 7)
 
     const account = new this.stellarApi.Account(
       this.walletLocalData.publicKey,
       this.otherData.accountSequence
     )
     let memoId: ?string
     if (
-      edgeSpendInfo.spendTargets[0].otherParams &&
+      edgeSpendInfo.spendTargets[0].otherParams != null &&
```
peachbits, line 429: Replace entire condition with optional chaining

peachbits (changes requested):

src/ethereum/ethEngine.ts
paullinator: The goal of this PR is not to fix all the possible TS errors but to just not be "worse than Flow". A future commit auto adds the appropriate eslint exception comment which we can remove in future work and slowly clean up.

src/fio/fioEngine.ts
paullinator: Commas are actually incorrect TS syntax. TS uses newlines to separate type definitions for objects. But if each param is on the same line, they are separated by ;

src/stellar/stellarEngine.ts
paullinator: The goal of this PR is not to fix all the possible TS errors but to just not be "worse than Flow". A future commit auto adds the appropriate eslint exception comment which we can remove in future work and slowly clean up. Otherwise you'd be making 100s of these type of comments

src/stellar/stellarEngine.ts
paullinator: The goal of this PR is not to fix all the possible TS errors but to just not be "worse than Flow". A future commit auto adds the appropriate eslint exception comment which we can remove in future work and slowly clean up. Otherwise you'd be making 100s of these type of comments

src/stellar/stellarEngine.ts
paullinator: The goal of this PR is not to fix all the possible TS errors but to just not be "worse than Flow". A future commit auto adds the appropriate eslint exception comment which we can remove in future work and slowly clean up. Otherwise you'd be making 100s of these type of comments

src/stellar/stellarEngine.ts
paullinator: This was caused by an eslint autoformat. For some reason eslint won't let you use `activated === false` since it thinks that `activated` can never be undefined. I re-ordered the IF statement to first check undefined then for false.

peachbits:
PR is approved by me and the final commit approved OTS by @paullinator

PR #445, opened 2022-09-24
Fix transaction query on BNB

paullinator:
Implement new txs query API

---
  - https://app.asana.com/0/0/1202925838052724


src/binance/bnbInfo.ts
```diff
@@ -6,11 +6,11 @@ import { BinanceSettings } from './bnbTypes'
 
 const otherSettings: BinanceSettings = {
   binanceApiServers: [
-    'https://dex.binance.org',
```
peachbits, line 9: I think this endpoint is still serving requests and shouldn't be removed.

src/binance/bnbInfo.ts
```diff
@@ -6,11 +6,11 @@ import { BinanceSettings } from './bnbTypes'
...
     'https://dex-atlantic.binance.org',
     'https://dex-asiapacific.binance.org',
     'https://dex-european.binance.org'
-  ]
+  ],
+  binanceNewApiServers: ['https://api.binance.org']
```
peachbits, line 13: These are only used to hit the tx endpoints so the name should reflect that instead of 'new'

peachbits (changes requested):

src/binance/bnbInfo.ts
paullinator: For now we are only using the `txs` endpoints, but the servers themselves are of a new API class. Binance hasn't given them a precise name so I can only call them "new" API servers.

src/binance/bnbInfo.ts
paullinator: I'll add back. I think it was a mistaken removal.

src/binance/bnbEngine.ts
```diff
@@ -400,17 +394,16 @@ export class BinanceEngine extends CurrencyEngine<BinancePlugin> {
     const data =
       spendTarget.otherParams != null ? spendTarget.otherParams.data : undefined
 
-    let otherParams: Object = {}
+    let otherParams: BinanceTxOtherParams | undefined
```
peachbits, line 397: I believe this just needs to be BinanceTxOtherParams since it gets defined no matter what. That alone doesn't really matter BUT maybe removing the conditional below it is because we only support BNB transfers and the else condition is arguably dead code.

peachbits (changes requested):

peachbits (approved):

PR #464, opened 2022-11-22
Add Tron (TRX)

peachbits:
---
  - https://app.asana.com/0/0/1191598790263107


package.json
```diff
@@ -42,7 +42,7 @@
     "prepare": "husky install && npm-run-all clean rollup postinstall lib types webpack checkpoints",
     "rollup": "rollup -c",
     "test": "nyc mocha 'test/**/*.test.ts'",
-    "postinstall": "sh ./postinstall.sh",
```
swansontec, line 45: Pretty sure this fix is already on master.

src/common/engine.ts
```diff
@@ -75,7 +75,7 @@ export class CurrencyEngine<T> {
   log: EdgeLog
   warn: (message: string, e?: Error) => void
   error: (message: string, e?: Error) => void
-  otherData: Object
+  otherData: { [key: string]: any }
```
swansontec, line 78: We might want to cherry-pick this first commit as its own PR. Seems like a pretty juicy cleanup that shouldn't depend on Tron getting in.

src/tron/tronUtils.ts
```diff
@@ -0,0 +1,5 @@
+// @ts-expect-error
+import { pkToAddress } from '@tronscan/client/src/utils/crypto'
```
swansontec: Add `declare module '@tronscan/client/src/utils/crypto'` to src/declare-modules.d.ts to make this come out clean.

If you want to get really fancy, you can even put full type definitions in that file to avoid this whole re-export hack.

src/tron/tronPlugin.ts
```diff
@@ -44,7 +73,13 @@ export class TronPlugin extends CurrencyPlugin {
   }
 
   async derivePublicKey(walletInfo: EdgeWalletInfo): Promise<Object> {
-    throw new Error('Must implement checkBlockchainInnerLoop')
+    const type = walletInfo.type.replace('wallet:', '')
+    if (type === this.currencyInfo.pluginId) {
```
swansontec, line 78: This is wrong. You should be doing `if (walletType !== this.currencyInfo.walletType) { throw ... }`, instead of slicing up the strings.

src/tron/tronPlugin.ts
```diff
@@ -27,11 +30,37 @@ export class TronPlugin extends CurrencyPlugin {
...
+      return {
+        tronMnemonic: userInput,
+        tronKey
+      }
+    }
   }
 
   async createPrivateKey(walletType: string): Promise<Object> {
-    throw new Error('Must implement checkBlockchainInnerLoop')
+    const type = walletType.replace('wallet:', '')
```
swansontec, line 56: This is wrong. You should be doing  `if (walletType !== this.currencyInfo.walletType) { throw ... }`, instead of slicing up the strings.

src/tron/tronEngine.ts
```diff
@@ -0,0 +1,1060 @@
...
+    currencyPlugin: TronPlugin,
+    walletInfo: EdgeWalletInfo,
+    opts: EdgeCurrencyEngineOptions,
+    fetchCors: EdgeFetchFunction
+  ) {
+    super(currencyPlugin, walletInfo, opts)
+    this.fetchCors = fetchCors
+    this.log = opts.log
+    this.settings = this.currencyInfo.defaultSettings
+      .otherSettings as TronSettings
```
swansontec, line 96: We need to stop jamming our network info into `defaultSettings`. Take a look at the `EosNetworkInfo` type, and how I pass it to the engine / tools constructors. We need to rename `TronSettings` to `TronNetworkInfo` and pass it into the engine alongside the currency info, instead of inside it.

src/tron/tronEngine.ts
```diff
@@ -0,0 +1,1060 @@
...
+    this.energyEstimateCache = {} // Minimize calls to check energy estimate
+    this.processTRXTransaction = this.processTRXTransaction.bind(this)
+    this.processTRC20Transaction = this.processTRC20Transaction.bind(this)
+  }
+
+  async fetch(
+    server: string,
+    path: string,
+    opts: Object = {}
+  ): Promise<Object> {
```
swansontec, line 122: The return type should be `{ server: string, result: EdgeFetchResponse }`, not `Object.`

src/tron/tronEngine.ts
```diff
@@ -0,0 +1,1060 @@
...
+export class TronEngine extends CurrencyEngine<TronPlugin> {
+  fetchCors: EdgeFetchFunction
+  log: EdgeLog
+  readonly recentBlock: ReferenceBlock
+  accountResources: TronAccountResources
+  networkFees: TronNetworkFees
+  settings: TronSettings
+  accountExistsCache: { [address: string]: boolean }
+  energyEstimateCache: { [addressAndContract: string]: number }
+  tronscan: any
```
swansontec, line 84: `tronscan: TronScan`, since you have in fact made a class for this one.

swansontec (changes requested):
This is just a first-round review. We need to get this rebased onto master, which will change a few things, and then I can do another round.

Also, I don't see you implement `getTokenId`, which is necessary because this chain has token support.

src/tron/tronEngine.ts
peachbits: Instead of threading this through and rebasing I just added as an additional commit

package.json
peachbits: Ok will drop it

peachbits: I did some testing and install fails when using a github URL but if you install from npm tarball it works fine. I'm ok with omitting.

swansontec (approved):
There are a few small things that you should fix here, but I am going to approve this. All of the fixes are tiny.

src/tron/tronPlugin.ts
```diff
@@ -0,0 +1,187 @@
...
+      }
+    }
+  }
+
+  async createPrivateKey(walletType: string): Promise<Object> {
+    if (walletType !== this.currencyInfo.walletType) {
+      throw new Error('InvalidWalletType')
+    }
+
+    const entropy = Buffer.from(this.io.random(32)).toString('hex')
```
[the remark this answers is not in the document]
peachbits: Leaving for now since that pattern is used across accountbased. If we update it lets do it across the board in a separate PR

src/tron/tronPlugin.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  async importPrivateKey(userInput: string): Promise<Object> {
+    if (/^(0x)?[0-9a-fA-F]{64}$/.test(userInput)) {
+      // It looks like a private key, so validate the hex:
+      const tronKeyBuffer = Buffer.from(userInput.replace(/^0x/, ''), 'hex')
+      if (EthereumUtil.isValidPrivate(tronKeyBuffer) === true) {
+        throw new Error('Invalid private key')
+      }
+      const tronKey = tronKeyBuffer.toString('hex')
```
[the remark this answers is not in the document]
peachbits: No it creates a string without the prefix

PR #465, opened 2022-11-22
Currency tools cleanup

swansontec:
All the `EdgeCurrencyTools` instances were called "plugin", which is wrong. They also contain logical errors, which need to be fixed, and use a base class that isn't useful (it's getting in the way of the bundle splitting).

## Changelog

- removed: Do not implement `getTokenId` on plugins that do not support tokens.

---
  - https://app.asana.com/0/0/1203417989127074


src/xrp/xrpPlugin.ts
```diff
@@ -29,7 +29,7 @@ import { asyncWaterfall, getDenomInfo, safeErrorMessage } from '../common/utils'
 import { XrpEngine } from './xrpEngine'
 import { currencyInfo } from './xrpInfo'
 
-export class XrpPlugin implements EdgeCurrencyTools {
+export class RippleTools implements EdgeCurrencyTools {
```
peachbits, line 32: Ah yes, because '"ripple" is network name. XRP is just an asset'

peachbits (approved):
I finished most of this earlier but when I returned to it the commits had been force pushed. re-reviewed it and couldn't figure out what was different and it all still looks good

PR #466, opened 2022-11-22
Filter blockscout from broadcasting tx method

adil651:
Unlike etherscan, the blackscout does not provide proxy for sending/broadcasting a signed transaction. 

Currently, one of the four ways to broadcast a transaction is through `evmScanServer` using the url `https://blockscout.com/etc/mainnet/api?module=proxy&action=eth_sendRawTransaction...`.

Blockscout is being used for Ethereum Classic and RSK blockchains. However, the proxy endpoints are not supported by Blockscout.

- API docs for [RSK blockscout](https://blockscout.com/rsk/mainnet/api-docs) 
- API docs for [Ethereum Classic (ETC)](https://blockscout.com/etc/mainnet/api-docs)

This pull request will filter the `evmScanServers` for blockscout and will disable the send requests for broadcasting transaction using blockscout.

---
  - https://app.asana.com/0/0/1203476994536515

### DIFF
--- a/src/ethereum/ethNetwork.ts
+++ b/src/ethereum/ethNetwork.ts
@@ -761,12 +761,15 @@ export class EthereumNetwork {
         })
 
         evmScanApiServers.forEach(baseUrl => {
-          promises.push(
-            broadcastWrapper(
-              this.broadcastEtherscan(params[0], baseUrl),
-              'etherscan'
+          // blockscout doesn't support Geth/Parity Proxy (eth_sendRawTransaction/broadcasting tx) for both ethereum classic and rsk
+          if (!baseUrl.includes('blockscout')) {
+            promises.push(
+              broadcastWrapper(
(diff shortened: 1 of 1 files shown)


swansontec:
I have assigned this PR for review. We are closing our our development sprint this weekend, so assuming we get this approved, the soonest it would go to app stores is mid-December.

peachbits:
For broadcasting, specifically, it's OK for an endpoint to throw an error because we broadcast to every possible server type and consider it good when the first server returns a success message. That means we can keep this code (which is shared across many currencies) clean and not clutter it with exceptions for individual currencies. 

The plugin will bubble up the last error message seen in the case that every broadcast attempt fails. If you're seeing an error, it's likely caused by an issue with the transaction itself and wouldn't be caused by a single broadcast hitting a non-existent endpoint.

PR #470, opened 2022-12-02
Replace all JSON Schema uses with Cleaners

swansontec:
I have managed to test everything except the EOS receive logic. The EOS node we use in `checkIncomingTransactions` simply isn't returning any results, even with incoming transactions, so the cleaner never runs.

---
  - https://app.asana.com/0/0/1203483455408818

### DIFF
--- a/src/eos/eosSchema.ts
+++ b/src/eos/eosSchema.ts
@@ -121,31 +121,17 @@ query ($query: String!, $limit: Int64, $low: Int64, $high: Int64) {
 }
 `
 
-// note that transfers are regular EOS transactions
-export const EosTransactionSuperNodeSchema = {
-  type: 'object',
-  properties: {
-    act: {
-      type: 'object',
-      properties: {
-        data: {
-          type: 'object',
-          properties: {
-            from: { type: 'string' },
-            to: { type: 'string' },
-            amount: { type: 'number' },
-            symbol: { type: 'string' },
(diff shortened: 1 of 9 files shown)


src/ethereum/ethSchema.ts
```diff
@@ -25,7 +25,7 @@ export const asEtherscanGetBlockHeight = asObject({
 })
 
 export const asEtherscanGetAccountNonce = asObject({
-  result: asString
+  result: asHexString
```
swansontec, line 28: I got this wrong originally, but testing quickly revealed the issue.

src/eos/eosSchema.ts
```diff
@@ -121,31 +121,17 @@ query ($query: String!, $limit: Int64, $low: Int64, $high: Int64) {
...
-    required: ['act', 'trx_id', '@timestamp', 'block_num']
-  }
-}
+export const asEosTransactionSuperNodeSchema = asObject({
+  act: asObject({
+    data: asObject({
+      from: asString,
+      to: asString,
+      amount: asNumber,
+      symbol: asString,
```
peachbits, line 130: This appears to be optional

peachbits: Oof, nevermind it's required. The schema was wrong.

peachbits (approved):

PR #475, opened 2022-12-06
Add new KNC token and rename old one

peachbits:
Old token is no longer used anywhere so we need to change the currency code so its untradeable. 
---
  - https://app.asana.com/0/0/1203144923997226


peachbits:
> Approved, but I suppose we could have left the denomination as "KNC", since that's just a display thing. However, I think the way you have it is clearer.

I think that would break where we lookup denoms by name === currencyCode. Maybe one day we restructure denoms into and object with exchangeDenom and then other options

PR #478, opened 2022-12-14
Upgrade Zcash plugin and add Piratechain

peachbits:
---
  - https://app.asana.com/0/0/1203550381026874


swansontec (changes requested):
I see a few small things to fix here, but overall this is really clean & nice.

swansontec (approved):

PR #483, opened 2022-12-23
Implement Accelerate Core API

samholmes:
### CHANGELOG

- Changed: Implement accelerate transaction feature using new core API

### Dependencies

- https://github.com/EdgeApp/edge-core-js/pull/493

---
  - https://app.asana.com/0/0/1203272763143177

### DIFF
--- a/src/ethereum/ethEngine.ts
+++ b/src/ethereum/ethEngine.ts
@@ -2,6 +2,7 @@ import Common from '@ethereumjs/common'
 import { Transaction } from '@ethereumjs/tx'
 import WalletConnect from '@walletconnect/client'
 import { add, div, gt, lt, lte, mul, sub } from 'biggystring'
+import { asMaybe } from 'cleaners'
 import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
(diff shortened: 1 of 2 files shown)


src/ethereum/ethEngine.ts
```diff
@@ -1169,6 +1174,116 @@ export class EthereumEngine extends CurrencyEngine<EthereumTools> {
...
+      return null
+    }
+    // Other params checks
+    if (
+      replacedTxOtherParams == null ||
+      replacedTxOtherParams.nonceUsed == null ||
+      replacedTxOtherParams.data != null
+    ) {
+      return null
+    }
```
swansontec, line 1177: Can't we extract the nonce and the data from the raw on-chain transaction?

Edit: Yes, we can use @ethereumjs/tx to do this. However, it doesn't look like this plugin stores the raw on-chain transaction anywhere to begin with. So, maybe we could update `processEvmScanTransaction` and friends to stick the data and nonce into `otherParams`, but for now we rely on our `makeSpend` to do this, and pray the user doesn't do a resync.

swansontec (approved):
Approved, with the caveat that a resync will disable the feature for any pending sends.

PR #486, opened 2023-01-06
Audit getMaxSpendable implementations

peachbits:
There are five implementations:

ethereum: Able to remove expensive recursion
ripple: Left comment about logic regarding future token support
zcash: Works as long as network fees remain hardcoded
tron: Unable to remove recursion since the network fee is dependent on the amount sent
polkadot: Same as tron

---
  - https://app.asana.com/0/0/1203660431077330


swansontec (approved):
This makes sense, but we really need to QA the heck out of it.

PR #489, opened 2023-01-11
DOT: Add hard cap of 1 to transaction query status

peachbits:


peachbits:
I wasn't able to reproduce the inaccurate progress but this would prevent the transaction count math from ever exceeding 1.

swansontec (approved):

PR #491, opened 2023-01-16
Tron: Add memo support

peachbits:
---
  - https://app.asana.com/0/0/1203557533664624


swansontec (changes requested):
I do not see the changes to the `EdgeCurrencyInfo`, giving the supported memo type as `memoType: "text"`, as well as the`memoMaxLength`, if there is one.

swansontec (approved):

PR #495, opened 2023-01-19
Cleanup Polkadot and fix runtime queryInfo error

peachbits:
Cleanup error suppression from typescript conversion
Use types from SDK
Fix disconnect bug
Upgrade to latest SDK

---
  - https://app.asana.com/0/0/1203772384220476


src/polkadot/polkadotEngine.ts
```diff
@@ -190,8 +185,7 @@ export class PolkadotEngine extends CurrencyEngine<PolkadotTools> {
       } catch (e: any) {
         if (
           typeof e?.message === 'string' &&
-          // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
-          e.message.includes('Subscan /scan/transfers failed with 429')
+          e.message.includes('Subscan /scan/transfers failed with 429') === true
```
swansontec, line 181: We are getting an error because `e` is `any`, and so ESLint doesn't know what `e.message.includes` returns. Another fix would be to do `e instanceof error`, instead of `e?.message === 'string'` on the line above. That would tell ESLint that this is an error, and so it would realize the type of `message.includes`

src/polkadot/polkadotEngine.ts
```diff
@@ -245,25 +238,23 @@ export class PolkadotEngine extends CurrencyEngine<PolkadotTools> {
...
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    super.startEngine()
+    this.addToLoop('queryBlockheight', BLOCKCHAIN_POLL_MILLISECONDS).catch(
+      () => {}
+    )
+    this.addToLoop('queryBalance', ACCOUNT_POLL_MILLISECONDS).catch(() => {})
+    this.addToLoop('queryTransactions', TRANSACTION_POLL_MILLISECONDS).catch(
+      () => {}
+    )
+    super.startEngine().catch(() => {})
```
swansontec, line 246: Please `await` the super call.

yarn.lock
```diff
@@ -169,13 +169,20 @@
...
 
-"@babel/runtime@^7.10.4", "@babel/runtime@^7.11.2", "@babel/runtime@^7.12.5", "@babel/runtime@^7.18.9", "@babel/runtime@^7.9.2":
+"@babel/runtime@^7.10.4", "@babel/runtime@^7.11.2", "@babel/runtime@^7.12.5", "@babel/runtime@^7.9.2":
   version "7.20.6"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.20.6.tgz#facf4879bfed9b5326326273a64220f099b0fce3"
   integrity sha512-Q+8MqP7TiHMWzSfwiJwXCjyf4GYA4Dgw3emg/7xmwsdLJOZUp+nMqcOwOzzYheuM1rhDu8FSj2l0aoMygEuXuA==
   dependencies:
     regenerator-runtime "^0.13.11"
 
+"@babel/runtime@^7.20.6", "@babel/runtime@^7.20.7":
```
swansontec, line 179: Please run yarn-deduplicate to clean some of these up.

swansontec (changes requested):
Found some small stuff.

src/polkadot/polkadotEngine.ts
peachbits: Nice, done.

src/polkadot/polkadotEngine.ts
peachbits: done

yarn.lock
peachbits: done

src/polkadot/polkadotEngine.ts
```diff
@@ -243,7 +243,7 @@ export class PolkadotEngine extends CurrencyEngine<PolkadotTools> {
     this.addToLoop('queryTransactions', TRANSACTION_POLL_MILLISECONDS).catch(
       () => {}
     )
-    super.startEngine().catch(() => {})
+    await super.startEngine().catch(() => {})
```
swansontec, line 246: Please remove this catch. We want to fail if the super fails, since we are pretty much dead in the water.

yarn.lock
```diff
@@ -169,14 +169,7 @@
...
 
-"@babel/runtime@^7.10.4", "@babel/runtime@^7.11.2", "@babel/runtime@^7.12.5", "@babel/runtime@^7.9.2":
-  version "7.20.6"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.20.6.tgz#facf4879bfed9b5326326273a64220f099b0fce3"
-  integrity sha512-Q+8MqP7TiHMWzSfwiJwXCjyf4GYA4Dgw3emg/7xmwsdLJOZUp+nMqcOwOzzYheuM1rhDu8FSj2l0aoMygEuXuA==
-  dependencies:
-    regenerator-runtime "^0.13.11"
-
-"@babel/runtime@^7.20.6", "@babel/runtime@^7.20.7":
+"@babel/runtime@^7.10.4", "@babel/runtime@^7.11.2", "@babel/runtime@^7.12.5", "@babel/runtime@^7.20.6", "@babel/runtime@^7.20.7", "@babel/runtime@^7.9.2":
```
swansontec, line 172: This should be `fixup! Upgrade @polkadot/api to latest`. Running yarn-deduplicate should be SOP whenever adding or removing dependencies.

swansontec (changes requested):

swansontec (approved):

PR #496, opened 2023-01-19
Tron: Fix derivation path

peachbits:
---
  - https://app.asana.com/0/0/1203769262371208

### DIFF
--- a/src/tron/tronPlugin.ts
+++ b/src/tron/tronPlugin.ts
@@ -24,12 +24,14 @@ export class TronTools implements EdgeCurrencyTools {
   io: EdgeIo
   currencyInfo: EdgeCurrencyInfo
   log: EdgeLog
+  networkInfo: TronNetworkInfo
 
   constructor(env: PluginEnvironment<TronNetworkInfo>) {
-    const { currencyInfo, io, log } = env
+    const { currencyInfo, io, log, networkInfo } = env
     this.io = io
     this.currencyInfo = currencyInfo
     this.log = log
+    this.networkInfo = networkInfo
   }
 
   async importPrivateKey(userInput: string): Promise<Object> {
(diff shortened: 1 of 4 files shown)


src/tron/tronPlugin.ts
```diff
@@ -20,6 +20,12 @@ import { encodeUriCommon, parseUriCommon } from '../common/uriHelpers'
 import { getDenomInfo } from '../common/utils'
 import { TronNetworkInfo } from './tronTypes'
 
+interface TronKeys {
+  tronMnemonic?: string
+  tronKey: string
+  derivationPath?: string
+}
```
swansontec, line 27: This type could live in tronTypes.ts. If you make a matching cleaner for it, you would use that cleaner in both `derivePublicKey` and in the currency engine, where you currently treat the keys as `any`.

swansontec (approved):
Approving with an optional fix.

I think I see why I was confused during our meeting - for Bitcoin, we don't store anything like the `tronKey` in the keys structure. We derive the Bitcoin keys every time, which is why we care about the format so much. For Tron, you actually save the `tronKey`, so the derivation path is a one-time issue affecting creation and importing. After that, it's baked in, so the engine doesn't care. This makes the fix dramatically simpler, which is great.

src/tron/tronTypes.ts
```diff
@@ -12,6 +12,18 @@ import {
   Cleaner
 } from 'cleaners'
 
+export interface TronKeys {
+  tronMnemonic?: string
+  tronKey: string
+  derivationPath?: string
+}
+
+export const asTronKeys = asObject({
```
swansontec, line 21: You can mark this as `asObject<TronKeys>`, to link the two types.

swansontec (approved):
One more tiny optional change. No need for re-review if you want to do this.

PR #500, opened 2023-01-28
Build fixes

swansontec:
---
  - https://app.asana.com/0/0/1203840174083015


webpack.config.js
```diff
@@ -29,21 +29,19 @@ module.exports = {
...
-          }
+          options: { loader: 'ts' }
         }
       }
     ]
   },
   optimization: {
     minimizer: [
       new ESBuildMinifyPlugin({
-        target: 'chrome67' // Syntax to compile to (see options below for possible values)
```
peachbits, line 46: I'm not sure if we should specify a target in the loader or minifier, I'll trust that this is the correct place, but it cannot be es2015. The exponentiation operator will be transformed and break bigints https://esbuild.github.io/content-types/#javascript We should keep it chrome67

peachbits (changes requested):

webpack.config.js
swansontec: Ok, very well.

peachbits (approved):

PR #502, opened 2023-01-31
Paul/xrp tokens

paullinator:
---
  - https://app.asana.com/0/0/1203583157723847

### DIFF
--- a/src/common/engine.ts
+++ b/src/common/engine.ts
@@ -143,6 +143,7 @@ export class CurrencyEngine<
       lastCheckedTxsDropped: 0,
       numUnconfirmedSpendTxs: 0,
       numTransactions: {},
+      unactivatedTokenIds: [],
       otherData: undefined
     }
     this.log(
@@ -353,6 +354,7 @@ export class CurrencyEngine<
     await Promise.all(addTokenPromises)
 
     this.doInitialBalanceCallback()
+    this.doInitialUnactivatedTokenIdsCallback()
   }
 
   findTransaction(currencyCode: string, txid: string): number {
@@ -614,6 +616,21 @@ export class CurrencyEngine<
           e
         )
       }
+    }
+  }
+
+  doInitialUnactivatedTokenIdsCallback(): void {
+    try {
+      if (
+        this.walletLocalData.unactivatedTokenIds != null &&
+        this.walletLocalData.unactivatedTokenIds.length > 0
+      ) {
+        this.currencyEngineCallbacks.onUnactivatedTokenIdsChanged(
+          this.walletLocalData.unactivatedTokenIds
+        )
+      }
+    } catch (e: any) {
+      this.error(`doInitialUnactivatedTokenIdsCallback Error`, e)
     }
   }
 
--- a/src/common/engine.ts
+++ b/src/common/engine.ts
@@ -38,6 +38,7 @@ import {
 import {
   cleanTxLogs,
   getDenomInfo,
+  matchJson,
   normalizeAddress,
   safeErrorMessage
 } from './utils'
@@ -77,7 +78,7 @@ export class CurrencyEngine<
   log: EdgeLog
   warn: (message: string, e?: Error) => void
   error: (message: string, e?: Error) => void
-  otherData: { [key: string]: any }
+  otherData: unknown
 
   constructor(
     env: PluginEnvironment<{}>,
@@ -113,6 +114,7 @@ export class CurrencyEngine<
     this.customTokens = []
     this.allTokensMap = { ...customTokens, ...env.builtinTokens }
     this.timers = {}
+    this.otherData = undefined
 
     this.transactionList[currencyCode] = []
     this.txIdMap[currencyCode] = {}
@@ -140,9 +142,8 @@ export class CurrencyEngine<
       lastCheckedTxsDropped: 0,
       numUnconfirmedSpendTxs: 0,
       numTransactions: {},
-      otherData: {}
+      otherData: undefined
     }
-    this.otherData = {}
     this.log(
       `Created Wallet Type ${this.walletInfo.type} for Currency Plugin ${this.currencyInfo.pluginId}`
     )
@@ -166,6 +167,10 @@ export class CurrencyEngine<
       }
     }
     return out
+  }
+
+  setOtherData(raw: any): void {
+    throw new Error(`Unimplemented setOtherData for ${this.walletInfo.type}`)
   }
 
   async loadTransactions(): Promise<void> {
@@ -276,6 +281,11 @@ export class CurrencyEngine<
         throw e
       }
     }
+    this.setOtherData(this.walletLocalData.otherData ?? {})
+    this.walletLocalDataDirty = !matchJson(
+      this.otherData,
+      this.walletLocalData.otherData
+    )
 
     // Add the native token currency
     this.tokenCheckBalanceStatus[currencyCode] = 0
@@ -577,6 +587,7 @@ export class CurrencyEngine<
     }
     if (this.walletLocalDataDirty) {
       this.log('walletLocalDataDirty. Saving...')
+      this.walletLocalData.otherData = this.otherData
       const jsonString = JSON.stringify(this.walletLocalData)
       await disklet
         .setText(DATA_STORE_FILE, jsonString)
@@ -704,7 +715,7 @@ export class CurrencyEngine<
     this.txIdList = {}
     this.txIdMap = {}
     this.transactionListDirty = true
-    this.otherData = this.walletLocalData.otherData
+    this.setOtherData({})
     await this.saveWalletLoop()
   }
 
--- a/src/common/engine.ts
+++ b/src/common/engine.ts
@@ -108,7 +108,7 @@ export class CurrencyEngine<
     this.txIdMap = {}
     this.txIdList = {}
     this.walletInfo = walletInfo
-    this.walletId = walletInfo.id != null ? `${walletInfo.id} - ` : ''
+    this.walletId = walletInfo.id
     this.currencyInfo = currencyInfo
     this.allTokens = currencyInfo.metaTokens.slice(0)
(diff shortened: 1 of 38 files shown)


src/common/types.ts
```diff
@@ -1,18 +1,40 @@
...
+    if (raw != null) {
+      try {
+        return cleaner(raw)
+      } catch (e) {
+        // Do nothing. return fallback
+      }
+    }
+    return fallback
+  }
+}
```
swansontec, line 32: This is called `asMaybe`, and already includes the fallback behavior. You should use that instead.

src/common/utils.ts
```diff
@@ -412,6 +412,9 @@ export function biggyRoundToNearestInt(float: string): string {
 export const prettyPrintObject = (obj: any): void =>
   console.log(JSON.stringify(obj, null, 2))
 
+export const jsonDiff = (obj1: any, obj2: any): boolean =>
+  JSON.stringify(obj1) !== JSON.stringify(obj2)
+
```
swansontec, line 417: While this works, it is sensitive to the order of the keys. There is a `matchJson` function in the GUI you could steal if you want it to be more stable.

src/eos/eosTypes.ts
```diff
@@ -81,10 +85,10 @@ export interface EosAction {
...
-  accountName: string
-  // @ts-expect-error
-  lastQueryActionSeq: { [string]: number }
-  // @ts-expect-error
-  highestTxHeight: { [string]: number }
-}
+export const asEosWalletOtherData = asObject({
+  accountName: asSafeOptional(asString, ''),
+  lastQueryActionSeq: asSafeOptional(asMap(asNumber), {}),
+  highestTxHeight: asSafeOptional(asMap(asNumber), {})
```
swansontec, line 91: Just replace `asMap` with `asObject`. When `asObject` receives a cleaner as its parameter (as opposed to an object), it works like `asMap` used to work. The `asMap` cleaner is deprecated, and will go away in the next breaking version.

src/common/types.ts
```diff
@@ -48,57 +55,17 @@ export interface TransactionList {
...
+  blockHeight: asSafeOptional(asNumber, 0),
+  lastAddressQueryHeight: asSafeOptional(asNumber, 0),
+  lastTransactionQueryHeight: asSafeOptional(asMap(asNumber), {}),
+  lastTransactionDate: asSafeOptional(asMap(asNumber), {}),
+  publicKey: asSafeOptional(asString, ''),
+  totalBalances: asSafeOptional(asMap(asString), {}),
+  lastCheckedTxsDropped: asSafeOptional(asNumber, 0),
+  numUnconfirmedSpendTxs: asSafeOptional(asNumber, 0),
+  numTransactions: asSafeOptional(asMap(asNumber), {}),
+  otherData: asSafeOptional(asUnknown, {})
```
swansontec, line 70: All of these should become `asMaybe`, except the last one, which should be `asOptional(asUnknown, {})` to get the right fallback logic. This edge case is the only difference between `asMaybe` and `asSafeOptional`.

src/xrp/xrpEngine.ts
```diff
@@ -293,14 +288,77 @@ export class XrpEngine extends CurrencyEngine<RippleTools> {
...
+            const { account: issuer, currency } = line
+            const lineTokenId = makeTokenId({ currency, issuer })
+            const edgeToken = this.allTokensMap[lineTokenId]
+            if (
+              edgeToken != null &&
+              tokenCurrencyCode === edgeToken.currencyCode
+            ) {
+              return true
+            }
+            return false
```
swansontec, line 333: You could just `return edgeToken != null && tokenCurrencyCode === edgeToken.currencyCode`.

src/xrp/xrpEngine.ts
```diff
@@ -293,14 +288,77 @@ export class XrpEngine extends CurrencyEngine<RippleTools> {
...
+      } else {
+        this.error(`Error fetching address info: `, e)
         return
       }
-      this.error(`Error fetching address info: `, e)
+    }
+
+    if (
+      JSON.stringify(newUnactivatedTokenIds) !==
+      JSON.stringify(this.walletLocalData.unactivatedTokenIds)
```
swansontec, line 371: Wouldn't this be your `jsonDiff` routine from earlier?

src/common/engine.ts
```diff
@@ -108,7 +108,7 @@ export class CurrencyEngine<
     this.txIdMap = {}
     this.txIdList = {}
     this.walletInfo = walletInfo
-    this.walletId = walletInfo.id != null ? `${walletInfo.id} - ` : ''
+    this.walletId = walletInfo.id ?? ''
```
swansontec, line 111: The `?? ''` part isn't even needed, since `walletInfo.id` is non-optional, and the core can't live without it.

src/xrp/xrpEngine.ts
```diff
@@ -614,6 +627,121 @@ export class XrpEngine extends CurrencyEngine<RippleTools> {
...
+            `SUCCESS activateWallet.approve()\n${cleanTxLogs(edgeTransaction)}`
+          )
+          transactions.push(edgeTransaction)
+          await this.saveTx(edgeTransaction)
+        }
+        return { transactions }
+      },
+      close: async (): Promise<void> => {}
+    }
+    console.log('hello')
```
swansontec, line 745: Leftover?

swansontec (changes requested):
This has a lot of good cleanups in it! We could have gotten those submitted in their own PR's, if the deadline was less tight, but now we get to deal with everything all at once.

src/common/utils.ts
paullinator: I didn't care about the stability since I'm always comparing against an object that would have been created by the same routine in the same way, but I'll go ahead and pull in the matchJson from the GUI. thx

src/eos/eosTypes.ts
paullinator: 👍

src/common/types.ts
paullinator: Crud. Didn't realize that was asMaybe. thx

src/xrp/xrpEngine.ts
paullinator: 👍 I had those lines in for breakpoint testing. Unneeded now.

src/xrp/xrpEngine.ts
paullinator: yes, will remove

test/builtinTokens.test.ts
```diff
@@ -20,9 +21,11 @@ const fakePluginOptions: EdgeCorePluginOptions = {
 }
 
 const pluginIds = Object.keys(plugins) as Array<keyof typeof plugins>
+const pluginsWithoutMetatokens: BooleanMap = { ripple: true }
 
 describe('builtinTokens', function () {
   for (const pluginId of pluginIds) {
+    if (pluginsWithoutMetatokens[pluginId]) continue
```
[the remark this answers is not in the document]
paullinator: Per convo, agreed to allow readonly static BooleanMap

test/builtinTokens.test.ts
```diff
@@ -20,9 +21,11 @@ const fakePluginOptions: EdgeCorePluginOptions = {
 }
 
 const pluginIds = Object.keys(plugins) as Array<keyof typeof plugins>
+const pluginsWithoutMetatokens: BooleanMap = { ripple: true }
```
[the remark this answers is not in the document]
paullinator: Per convo, agreed to allow readonly static BooleanMap

src/xrp/xrpEngine.ts
```diff
@@ -103,28 +110,91 @@ export class XrpEngine extends CurrencyEngine<RippleTools> {
...
+    // For some reason, XRPL types exclude Transaction hash, date, and ledger_index.
+    // Check if the tx has these undocumented fields and throw if not
+    asXrpTransaction(accountTx.tx)
+
+    if (accountTx.tx.TransactionType !== 'Payment') {
+      return
+    }
+
+    const temp: any = accountTx.tx
+    const tx: Payment & XrpTransaction = temp
```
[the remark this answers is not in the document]
paullinator: fixed

src/xrp/xrpPlugin.ts
```diff
@@ -156,6 +157,18 @@ export class RippleTools implements EdgeCurrencyTools {
...
+  // issuer addresses can issue more than one token so we need
+  // the currency code to make the token id unique
+  async getTokenId(token: EdgeToken): Promise<string> {
+    const location = token?.networkLocation
+    if (location == null) {
+      throw new Error('ErrorInvalidNetworkLocation')
+    }
+    const { currency, issuer } = location
+    return `${currency}-${issuer}`
+  }
```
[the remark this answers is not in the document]
paullinator: fixed

src/common/types.ts
```diff
@@ -10,6 +10,10 @@ export const asErrorMessage = asObject({
   message: asString
 })
 
+export interface BooleanMap {
+  [key: string]: boolean
+}
```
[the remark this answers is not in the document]
paullinator: Per convo, agreed to allow readonly static BooleanMap

src/common/types.ts
```diff
@@ -1,18 +1,40 @@
...
 export const asErrorMessage = asObject({
   message: asString
 })
 
 export interface BooleanMap {
   [key: string]: boolean
 }
+export interface StringMap {
+  [key: string]: string
+}
```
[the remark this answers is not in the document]
paullinator: Made BooleanMap readonly and replaced the one use of StringMap with in-line { [key: string]: string }

swansontec (approved):
The fix for getTokenId does not include `.toUpperCase` or `.toLowerCase`, so I presume you have verified that the currency codes *are* case-sensitive at the network level. If so, approved.

PR #504, opened 2023-01-31
No meta tokens

swansontec:
This removes `metaTokens` from all our currency info's.

A big goal here is to have a common source of truth for each currency's tokenId's, so there is no risk of a human or a plugin being inconsistent. We also want to be really clear about what code lives in "info land", and is therefore cheap to load, and which code lives in "currency land", and is therefore expensive to load.

The engines may still refer to these legacy tokens internally, so that still needs to be cleaned up in another PR. We do pass the builtin tokens in our `PluginEnvironment`, so the engine has access to them.

---
  - https://app.asana.com/0/0/1203871927438613


src/common/innerPlugin.ts
```diff
@@ -132,21 +141,25 @@ export function makeOtherMethods<T>(
...
+ * Downgrades EdgeToken objects to the legacy EdgeMetaToken format.
+ */
+function makeMetaTokens(tokens: EdgeTokenMap): EdgeMetaToken[] {
+  const out: EdgeMetaToken[] = []
+  for (const tokenId of Object.keys(tokens)) {
+    const { currencyCode, displayName, denominations, networkLocation } =
+      tokens[tokenId]
+
+    const cleanLocation = asMaybeContractLocation(networkLocation)
+    if (cleanLocation == null) continue
```
paullinator, line 155: if an plugin doesn't use contractAddress in EdgeMetaToken (xrp doesn't) then you won't get any metatokens. Is this okay? If not, you should probably do the inverse and check if a plugin has metaTokens defined, if so then use them, if not derive them.

src/common/tokenHelpers.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * The `networkLocation` field is untyped,
+ * but many currency plugins will put a contract address in there.
+ */
+export const asContractLocation = asObject({
+  contractAddress: asString
+})
+
+export function getEosTokenId(token: EdgeToken): string {
```
paullinator, line 12: This doesn't feel clean to have asset specific logic in a `common` file. I would prefer to move these to asset specific files. Besides, they are only used by those specific asset files.

src/ethereum/info/avaxInfo.ts
```diff
@@ -1,8 +1,220 @@
-import { EdgeCurrencyInfo } from 'edge-core-js/types'
+import { EdgeCurrencyInfo, EdgeTokenMap } from 'edge-core-js/types'
 
 import { makeOuterPlugin } from '../../common/innerPlugin'
+import { getEvmTokenId, makeTokenMap } from '../../common/tokenHelpers'
 import type { EthereumTools } from '../ethPlugin'
-import { EthereumFees, EthereumSettings } from '../ethTypes'
+import type { EthereumFees, EthereumSettings } from '../ethTypes'
+
+const builtinTokens: EdgeTokenMap = makeTokenMap(getEvmTokenId, [
```
paullinator, line 8: Hmm. I'm not liking that the "source of truth" is in a weird array format that's neither the old or new EdgeTokenMap. Why not just use an EdgeTokenMap? That's how XRP is coded.

paullinator: Admittedly you'd want some code to automate the conversion the first time but then you're done and can commit the results. Given that the format of the EdgeMetaToken array and the new array are different, I presume you did use automation to do the conversion. Might as well go straight to a token map

src/ethereum/info/bscInfo.ts
```diff
@@ -82,12 +84,11 @@ export const currencyInfo: EdgeCurrencyInfo = {
       symbol: 'BNB'
     }
   ],
-  metaTokens: [
-    // Array of objects describing the supported metatokens
-  ]
+  metaTokens: [] // Deprecated
```
paullinator, line 87: Wow. how do we not have any tokens on BSC?

paullinator (changes requested):
Definitely the direction we want to head in! Main request is to go-all-the-way and use EdgeTokenMap for the built-in tokens

src/common/innerPlugin.ts
swansontec: I have made this up to the plugins to decide. Tron and all the EVM's are calling `metaTokens: makeMetaTokens(builtinTokens)` in their currencyInfo's, but the Ripple plugin can do whatever it needs to make this work.

src/common/tokenHelpers.ts
swansontec: I deleted all of them, and just left the common cleaner.

src/ethereum/info/avaxInfo.ts
swansontec: I did the conversion using regular expression search & replace. For the `EdgeTokenMap` conversion, I had to do it by hand. Now that it's done, we have unit tests to ensure that it always stays correct going forward.

src/common/innerPlugin.ts
paullinator: Ok. got it.

src/tron/tronInfo.ts
```diff
@@ -1,8 +1,88 @@
-import { EdgeCurrencyInfo } from 'edge-core-js/types'
+import { EdgeCurrencyInfo, EdgeTokenMap } from 'edge-core-js/types'
 
 import { makeOuterPlugin } from '../common/innerPlugin'
+import { makeMetaTokens } from '../common/tokenHelpers'
 import type { TronTools } from './tronPlugin'
-import { TronNetworkInfo } from './tronTypes'
+import type { TronNetworkInfo } from './tronTypes'
+
+const builtinTokens: EdgeTokenMap = {
```
paullinator, line 8: The previous code flow is a bit tough to follow, but wouldn't Tron tokens also go through the `loadInnerPlugin()` routine which calls `upgradeMetaTokens` which would have lowercased the tokenIds? This is what I ran into when implementing XRP even though my getTokenId didn't do the lowercasing. While the true upper/lower case tokenIds would have been preferred for Tron, we might have tokenIds already saved in user's accounts that are all lowercase.

paullinator (approved):
This looks great. Much cleaner and really motivates deprecating EdgeMetatoken.

Just one concern about Tron that's more of a question.

src/tron/tronInfo.ts
swansontec: After a sidebar with Matthew to discuss pros & cons, we have decided to go ahead and change the tokenId's. We have a plan for dealing with the fallout, which isn't as terrible as we thought at first.

PR #505, opened 2023-02-01
Token cleanup

paullinator:
Cleanup account based to preparation for XRP tokens.


src/xrp/xrpEngine.ts
```diff
@@ -208,23 +208,23 @@ export class XrpEngine extends CurrencyEngine<RippleTools> {
...
-      () => {}
+      e => console.log(e.message)
     )
     this.addToLoop(
       'checkTransactionsInnerLoop',
       TRANSACTION_POLL_MILLISECONDS
-    ).catch(() => {})
-    await super.startEngine()
+    ).catch(e => console.log(e.message))
+    await super.startEngine().catch(e => console.log(e.message))
```
swansontec, line 228: We should delete these `.catch` statements and just `await` all these calls, since the function is `async` anyhow. We really don't want to return success if `addToLoop` has crashed or something.

testConfig.json
```diff
@@ -0,0 +1,47 @@
+{
```
swansontec, line 1: Did this sneak into the PR on accident too?

src/binance/bnbEngine.ts
```diff
@@ -52,9 +52,10 @@ export class BinanceEngine extends CurrencyEngine<BinanceTools> {
     tools: BinanceTools,
     walletInfo: EdgeWalletInfo,
     initOptions: any, // BinanceInitOptions,
-    opts: any // EdgeCurrencyEngineOptions
+    opts: EdgeCurrencyEngineOptions,
+    env: PluginEnvironment<{}>
```
swansontec, line 56: Optional: In the `makeCurrencyEngine` function that the the innerPlugin system calls, the parameter order is:

```
(
  env: PluginEnvironment<...>,
  tools: EosTools,
  walletInfo: EdgeWalletInfo,
  opts: EdgeCurrencyEngineOptions
```

It would be nice if the engine constructors used the same order, followed by whatever coin-specific stuff they want. So `env` would go first in all these edits, instead of last.

swansontec (changes requested):
The only thing that needs to get fixed is the extra JSON files that snuck in. The other stuff would be "nice to have".

results.json
```diff
@@ -0,0 +1,949 @@
+{
```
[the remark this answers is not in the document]
paullinator: Ack. I pulled out a commit that had a .gitignore change and the file got included. I'll pull it out.

src/common/engine.ts
```diff
@@ -309,21 +287,20 @@ export class CurrencyEngine<
...
         networkLocation = {}
       } = customTokens[token]
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
       this.addCustomToken({
         currencyCode,
         currencyName: displayName,
         multiplier: denominations[0].multiplier,
         contractAddress: networkLocation?.contractAddress
-      })
+      }).catch(e => console.log(e.message))
```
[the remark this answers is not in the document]
paullinator: sounds good!

src/xrp/xrpEngine.ts
paullinator: I think this needs a bit of a discussion, I'm not sure if all currencies will throw due to "safe" errors like network errors. We don't want to kill the loop because of a failed network op.

Valid point though but a little out of scope of this PR.

testConfig.json
paullinator: yep. will remove

src/binance/bnbEngine.ts
paullinator: Ack. Pretty please can we make this a separate PR. There are just so many of these assets to modify and I have just one more day to get this in.

swansontec (approved):
I am leaving my review in the form or fixup commits, to help lighten the load a bit.

PR #508, opened 2023-02-08
EOS: Add PowerUp

peachbits:
---
  - https://app.asana.com/0/0/1200797042560630


src/eos/eosEngine.ts
```diff
@@ -1018,6 +1167,9 @@ export class EosEngine extends CurrencyEngine<EosTools> {
   async broadcastTx(
     edgeTransaction: EdgeTransaction
   ): Promise<EdgeTransaction> {
+    if (!this.getResourcesMutex) {
+      await this.getResources()
```
swansontec, line 1080: The way you have written `getResources`, it should be safe to call without any checks. So, it would be safer to remove the `if` statement here, so we don't leak that implementation detail out of the function.

src/eos/eosEngine.ts
```diff
@@ -229,7 +229,9 @@ export class EosEngine extends CurrencyEngine<EosTools> {
     }
   }
 
-  processIncomingTransaction(action: EosTransactionSuperNode): number {
+  processIncomingTransaction(
+    action: ReturnType<typeof asHyperionTransaction>
```
swansontec, line 230: You could add these as typedefs near the cleaners themselves. That's what we normally do.

src/eos/eosEngine.ts
```diff
@@ -465,13 +467,10 @@ export class EosEngine extends CurrencyEngine<EosTools> {
         // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
         low: newHighestTxHeight + 1
       }
-      const actionsObject = await this.multicastServers(
-        'getIncomingTransactions',
-        params
-      )
+      const actionsObject: ReturnType<typeof asHyperionGetTransactionResponse> =
```
swansontec, line 458: This could also be a typedef.

swansontec (changes requested):
For such a big PR, I couldn't find much to comment on, just some small stuff.

Your architecture requires the resource state to be up-to-date before the user can spend money successfully. Otherwise, the engine might think they have resources when they really don't, and skip the power up step. For instance, if the user buys resources, but then they expire 24 hours later while the phone is asleep.

I don't think this is a problem, since we expect spending to fail when the wallet is out-of-sync, but we should keep our ears to the ground. If support gets a bunch of reports, we may want to be more aggressive with the queries as part of spending.

swansontec (approved):

PR #511, opened 2023-02-14
Add Optimism

peachbits:
---
  - https://app.asana.com/0/0/1203970651584036

### DIFF
--- a/src/ethereum/fees/feeProviders.ts
+++ b/src/ethereum/fees/feeProviders.ts
@@ -31,7 +31,7 @@ export const printFees = (log: EdgeLog, fees: Object) => {
     // @ts-expect-error
     const value = fees[key]
     if (typeof value === 'string')
-      log.warn(`  ${key}: ${div(value, '1000000000', 18)} gwei`)
+      log(`  ${key}: ${div(value, '1000000000', 18)} gwei`)
   }
 }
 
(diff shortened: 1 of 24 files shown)


swansontec (changes requested):
Found some tiny things, but this is a great cleanup.

src/ethereum/fees/feeProviders.ts
```diff
@@ -65,10 +67,10 @@ export const fetchFeesFromEvmScan = async (
   fetch: EdgeFetchFunction,
   currencyInfo: EdgeCurrencyInfo,
   initOptions: EthereumInitOptions,
-  log: EdgeLog
+  log: EdgeLog,
+  networkInfo: EthereumNetworkInfo
```
swansontec, line 130: I noticed this the other day, and I didn't say anything, but maybe it's worth mentioning anyhow: `PluginEnvironment<NetworkInfo>` has all these things inside. You get `currencyInfo`, `networkInfo`, `log`, `initOptions`, and while it doesn't have `fetch` itself, you do get `io`.

However, the reason I didn't say anything is because this would be a bit bigger refactor, and I don't necessarily want to spend time on it, given the deadline. Just thoughts for later.

src/ethereum/fees/feeProviders.ts
```diff
@@ -62,6 +68,59 @@ export const FeeProviders = (
...
+  const { result } = rpcGasResponse
+  const gasPrice = hexToDecimal(result)
+
+  const out = {
+    lowFee: mul(gasPrice, '1'),
+    standardFeeLow: mul(gasPrice, '1.06'),
+    standardFeeHigh: mul(gasPrice, '1.12'),
+    highFee: mul(gasPrice, '1.25')
+  }
+  log.warn(`fetchFeesFromRpc: ${currencyInfo.currencyCode}`)
```
swansontec, line 119: Nothing is wrong here, so we should use `log()`, not `log.warn()`

swansontec (approved):
This all looks reasonable.

src/ethereum/fees/feeProviders.ts
peachbits: Went ahead and quieted the other fee fetching functions as well

src/ethereum/fees/feeProviders.ts
peachbits: Definitely worth mentioning and that would be a great commit in a future "Cleanup Ethereum" PR

PR #512, opened 2023-02-14
Fix missing XRP token transactions

paullinator:
Check both DeliveredAmount and delivered_amount is transaction meta object

---
  - https://app.asana.com/0/0/1203962644159309

### DIFF
--- a/src/xrp/xrpEngine.ts
+++ b/src/xrp/xrpEngine.ts
@@ -168,7 +168,7 @@ export class XrpEngine extends CurrencyEngine<RippleTools> {
         return
       }
 
-      const { DeliveredAmount } = meta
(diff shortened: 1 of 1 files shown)


src/xrp/xrpEngine.ts
```diff
@@ -168,7 +168,7 @@ export class XrpEngine extends CurrencyEngine<RippleTools> {
         return
       }
 
-      const { DeliveredAmount } = meta
+      const DeliveredAmount = meta.DeliveredAmount ?? meta.delivered_amount
```
swansontec, line 171: Weird that the servers do not agree, but I get it.

swansontec (approved):

PR #516, opened 2023-02-19
Apply hexToDecimal on eth_call token balance

paullinator:
---
  - https://app.asana.com/0/0/1204001370454367


swansontec (approved):
Well, this certainly looks correct, but it's hard to say for sure without more context. Let's merge this right away, but make a point to have @peachbits check this when he gets back.

peachbits (approved):

PR #519, opened 2023-03-01
Add optional checkEnvironment method to OuterPlugin

peachbits:
This allows the plugin to identify any minimal environment requirements before returning to the core

---
  - https://app.asana.com/0/0/1204070120353789

### DIFF
--- a/src/polkadot/polkadotInfo.ts
+++ b/src/polkadot/polkadotInfo.ts
@@ -45,6 +45,12 @@ export const polkadot = makeOuterPlugin<{}, PolkadotTools>({
   currencyInfo,
   networkInfo: {},
 
+  checkEnvironment: () => {
+    if (global.BigInt == null) {
+      throw new Error('Polkadot requires bigint support')
+    }
+  },
+
   async getInnerPlugin() {
     return await import(
       /* webpackChunkName: "polkadot" */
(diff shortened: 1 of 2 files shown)


peachbits:
blocked by https://github.com/EdgeApp/edge-core-js/pull/509

src/polkadot/polkadotInfo.ts
```diff
@@ -45,6 +46,12 @@ export const polkadot = makeOuterPlugin<{}, PolkadotTools>({
   currencyInfo,
   networkInfo: {},
 
+  checkEnvironment: () => {
+    if (typeof BigInt(1) !== 'bigint') {
+      throw new Error('Environment error: bigint unsupported')
+    }
+  },
```
swansontec, line 53: As mentioned in the other core PR, this is not the right design. We want the Polkadot plugin to exist, but simply fail when creating the tools or engine. This allows users to see that their wallets exist, but simply can't load on this particular phone. That way, they don't call support saying their funds disappeared if they switch phones or something.

One thing that will improve the UX is to make the wallet list loading row capable of showing an error message. Instead of a spinner, the dead Polkadot wallet would then show a message like "Not supported on this phone", so it's clear what happened. This would be powered by the new `EdgeAccount.currencyWalletErrors`

However, that's not the goal right away. The goal is just to have the red drop-down show something more meaningful.

swansontec (changes requested):

src/polkadot/polkadotInfo.ts
peachbits: After many iterations (and foot guns) I was able to get this to fail late with a custom error message

swansontec (approved):
Ah, this solution makes sense now. You need to do the check on the "good" side, since otherwise the script will simply fail to import, so it's too late to do anything about it. But you also need to do the check as part of engine / tools boot-up, and not prevent the currency info from loading.

PR #520, opened 2023-03-02
Incorporate L1fee into native amount

peachbits:
---
  - https://app.asana.com/0/0/1204067042367665


peachbits:
Previously, we weren't incoporating the l1fee into the native amount for mainnet sends

swansontec (approved):

PR #521, opened 2023-03-07
Incorporate L1 gas price max multiplier

peachbits:
Optimism guarantees the L1 gas price will not spike more than 25% between rollups.


src/ethereum/ethEngine.ts
```diff
@@ -994,7 +994,13 @@ export class EthereumEngine
...
       const txData: CalcL1RollupFeeParams = {
         nonce: otherParams.nonceUsed,
-        gasPriceL1Wei: this.l1RollupParams.gasPriceL1Wei,
+        gasPriceL1Wei: ceil(
+          mul(
+            this.l1RollupParams.gasPriceL1Wei,
+            this.l1RollupParams.maxGasPriceL1Multiplier
+          ),
+          0
+        ),
```
swansontec, line 1029: This math looks wrong. `ceil(x, 0)` rounds to the nearest 0? That doesn't make sense. Do you mean '1'?

src/ethereum/ethEngine.ts
```diff
@@ -751,7 +756,28 @@ export class EthereumEngine
...
+
+        if (this.l1RollupParams != null) {
+          const txData: CalcL1RollupFeeParams = {
+            nonce: this.otherData.unconfirmedNextNonce,
+            gasPriceL1Wei: ceil(
+              mul(
+                this.l1RollupParams.gasPriceL1Wei,
+                this.l1RollupParams.maxGasPriceL1Multiplier
+              ),
+              0
```
swansontec, line 769: I just saw this same code in the previous commit. Can it be common-ized? If not, no worries, but it needs the same fix as the other place.

swansontec (changes requested):

src/ethereum/ethEngine.ts
peachbits: According to the docs that's how biggystring "ceil"s https://github.com/EdgeApp/biggystring/blob/798ccf7f0b12732c739a53a55d89837511e01757/README.md#rounding-functions-base-10-only

swansontec: Holy cow, that's weird! Well, whatever. Looks good to me.

PR #522, opened 2023-03-07
Smartpay PIX support

paullinator:
### CHANGELOG

Add ability to parse and quote PIX codes for payment with Tron/USDT

### Dependencies

https://github.com/EdgeApp/edge-core-js/pull/510

### Description

none

---
  - https://app.asana.com/0/0/1203963906731673

### DIFF
--- a/src/common/smartPay.ts
+++ b/src/common/smartPay.ts
@@ -1,0 +1,150 @@
+import { mul } from 'biggystring'
+import { asEither, asNumber, asObject, asString, asValue } from 'cleaners'
+import { EdgeIo, EdgeParsedUri, EdgeTokenMap } from 'edge-core-js'
+
+import { cleanMultiFetch, makeQueryParams, QueryParams } from './network'
+import { computeCRC, formatPixKey } from './pixkey'
+
+const MAX_TIMEOUT_S = 60 * 60 * 24 * 7
+
+const asSmartPayQrDecode = asEither(
+  asObject({
+    status: asValue('ok'),
+    msg: asString,
+    data: asObject({
+      amount: asNumber,
+      name: asString,
+      key: asString,
+      timeout: asNumber
+    })
+  }),
+  asObject({
+    status: asValue('failed'),
+    msg: asString
+  })
+)
+
+const asSmartPaySwapQuote = asObject({
+  status: asValue('ok'),
+  msg: asString,
+  data: asObject({
+    amount_usd: asString, // '0.000020',
+    price_brl: asString, // '0.1923',
+    total_brl: asString, // '0.00',
+    fee_brl: asString, // '0.00',
+    send_brl: asString, // '0.0001',
+    timeout: asNumber, // 491,
+    amount_txusdt: asString, // '0.000020',
+    price_txusdt: asNumber, // 1,
+    value_usd: asString, // '0.000020',
+    total_txusdt: asString // '0.000020'
+  })
+})
+
+export const parsePixKey = async (
+  io: EdgeIo,
+  tokens: EdgeTokenMap,
+  code: string,
+  smartPayPublicAddress: string,
+  smartPayUserId: string
+): Promise<EdgeParsedUri | undefined> => {
+  const now = new Date()
+  if (code.length > 36) {
+    const crc = computeCRC(code.slice(0, -4))
+    if (!code.endsWith(crc)) {
+      return
+    }
+    try {
+      const qrcode = encodeURIComponent(code)
+      const decode = await cleanMultiFetch(
+        asSmartPayQrDecode,
+        ['https://connect.smartpay.com.vc'],
+        `api/pix/qrdecode?qrcode=${qrcode}`,
+        undefined,
+        undefined,
+        io.fetch
+      )
+      if (decode.status !== 'ok') {
+        throw new Error(decode.msg)
+      }
+      const { data: decodeData } = decode
+      const { amount, key, name, timeout: decodeTimeout } = decodeData
+      if (decodeTimeout > 0 && decodeTimeout < 120) {
+        throw new Error('ErrorPixExpired')
+      }
+
+      let nativeAmount: string | undefined
+      let expireDate: Date | undefined
+      if (amount !== 0) {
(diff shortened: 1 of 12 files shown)


src/common/smartPay.ts
```diff
@@ -0,0 +1,475 @@
...
+
+function numToHex(n: number, digits: number): string {
+  const hex = n.toString(16).toUpperCase()
+  if (digits > 0) {
+    return ('0'.repeat(digits) + hex).slice(-digits)
+  }
+  return hex.length % 2 === 0 ? hex : '0' + hex
+}
+
+function textEncode(string: string): Uint8Array {
```
swansontec, line 414: This is supposed to turn a `string` into UTF-8 text? Then grab the `utf8.parse` / `utf8.stringify` pair out of edge-core-js: https://github.com/EdgeApp/edge-core-js/blob/master/src/util/encoding.ts

Otherwise, if we are slicing this whole third-party thing into its own file, I guess we can leave the logic as it is.

src/common/smartPay.ts
```diff
@@ -0,0 +1,475 @@
...
+      publicAddress: 'TUmgPbM5J6om7Z2PJjzrbSEbXit84ZhVCj',
+      uniqueIdentifier: pixKey
+    }
+    return out
+  }
+}
+
+// ---------------------------------------------------------------------------
+// All code below provided by smartpay.com.vc
+// ---------------------------------------------------------------------------
```
swansontec, line 151: Can we put all this third-party code into its own file, like `pixKey.ts` or something? That makes it pretty clean in terms of ours vs. theirs.

src/common/smartPay.ts
```diff
@@ -0,0 +1,475 @@
...
+        )
+
+        const { data: quoteData } = quote
+        const { amount_txusdt: amountTxusdt, timeout: quoteTimeout } = quoteData
+        if (quoteTimeout > 0 && quoteTimeout < 120) {
+          throw new Error('ErrorPixExpired')
+        }
+
+        // Get USDT info
+        const token = currencyInfo.metaTokens.find(
```
swansontec, line 106: We are going to delete `metaTokens` soon. Can you give this guy access to `builtinTokens` instead?

test/plugin/smartPay.nettest.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+]
+
+for (const fixture of fixtures) {
+  let tools: EdgeCurrencyTools
+
+  const WALLET_TYPE = fixture.WALLET_TYPE
+
+  const fakeIo = makeFakeIo()
```
swansontec, line 94: Can you put this with the normal tests? There is no reason to create a whole new "nettest" command, unless we were maybe trying to separate tests that hit the network from tests that run locally. In that case, maybe we could go deeper on the separation, or at least create a task to do it later.

If you don't want to run the whole test suite during development, you can just do `mocha -g <seach term>` to just run matching tests.

Then, since this test seems to focus on SmartPay, maybe use this testing setup:

```ts
describe('SmartPay', async function () {
   // The same setup logic you already have:
   const fakeIo = ...
   const plugin = ...
   const tools = await plugin.getCurrencyTools()

  // The same loop you already have, but moved inside the `describe`:
  for (const fixture of fixtures) {
    it(fixture.uri, function() {
       // The same test as before...
     })
  })
})
```

This is more idiomatic. The `describe` block is meant to group related tests together, so it needs to be on the outside. Then, since `describe` makes it possible to use `async` / `await`, you can ditch the ugly `before` block.

Assuming the "nettest" separation really is about network-using tests, then this would ***not*** be a place to test generic URI parsing features. Therefore, this test should be named "SmartPay", specifically, so we aren't tempted to recycle it.

swansontec (changes requested):

src/common/smartPay.ts
swansontec: Yeah, it's possible. You can grab the `builtinTokens` in the tools constructor, assign it to `this`, and then it will be available in parseUri.

src/common/smartPay.ts
paullinator: will move into it's own file. I don't want to change their logic

test/plugin/smartPay.nettest.ts
```diff
@@ -86,34 +85,27 @@ const fixtures: SmartPayTestFixture = [
...
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    io: { ...fakeIo, fetch, random: size => fixture.key },
-    log: fakeLog,
-    nativeIo: {},
-    pluginDisklet: fakeIo.disklet
-  }
-  // @ts-expect-error
-  const factory = edgeCorePlugins[fixture.pluginId]
-  const plugin: EdgeCurrencyPlugin = factory(opts)
+    const factory = edgeCorePlugins[fixture.pluginId]
```
swansontec, line 102: We can fix the type error by annotating ` pluginId: keyof typeof edgeCorePlugins` in your fixture type definition.

test/plugin/smartPay.nettest.ts
```diff
@@ -86,34 +85,27 @@ const fixtures: SmartPayTestFixture = [
...
+const opts: EdgeCorePluginOptions = {
+  initOptions: { smartPayPublicAddress },
+  io: { ...fakeIo, fetch, random: size => new Uint8Array(32) },
+  log: fakeLog,
+  nativeIo: {},
+  pluginDisklet: fakeIo.disklet
+}
 
-  const WALLET_TYPE = fixture.WALLET_TYPE
+describe('SmartPay PIX parseUri', () => {
```
swansontec, line 97: This test structure is much more awesome. Thanks.

swansontec (approved):
Approved, with an optional fix for the ts-expect-error,

test/plugin/smartPay.nettest.ts
paullinator: good catch. done

paullinator: Nice catch. Done

PR #528, opened 2023-03-14
Prevent sending amounts below min threshold to new addresses

peachbits:
---
  - https://app.asana.com/0/0/1203925052101687

### DIFF
--- a/src/common/engine.ts
+++ b/src/common/engine.ts
@@ -1,4 +1,4 @@
-import { add, eq, gt, lt } from 'biggystring'
+import { add, div, eq, gt, gte, lt } from 'biggystring'
 import { Disklet } from 'disklet'
 import {
   EdgeCurrencyCodeOptions,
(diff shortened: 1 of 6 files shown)


src/common/engine.ts
```diff
@@ -1022,6 +1022,30 @@ export class CurrencyEngine<
     return { edgeSpendInfo, nativeBalance, currencyCode, denom, skipChecks }
   }
 
+  async checkRecipientMinimumBalance(
+    balancePromise: (address: string) => Promise<string>,
```
swansontec, line 1026: Optional: I would call this `getBalance`, since it's a function, not a promise.

src/common/engine.ts
```diff
@@ -1022,6 +1022,30 @@ export class CurrencyEngine<
...
+      )
+      if (denom == null) throw new Error('Unknown denom')
+
+      const exchangeDenomString = div(
+        this.minimumAddressBalance,
+        denom.multiplier
+      )
+      throw new Error(
+        `Recipient address not activated. A minimum ${exchangeDenomString} ${this.currencyInfo.currencyCode} transfer is required to send funds to this address`
+      )
```
swansontec, line 1045: We have a pending task in the core / GUI to add more `makeSpend` error types. I will mention this case in that task.

src/stellar/stellarEngine.ts
```diff
@@ -67,12 +67,39 @@ export class StellarEngine extends CurrencyEngine<StellarTools> {
...
     this.minimumAddressBalance = this.networkInfo.baseReserve
+
+    this.getRecipientBalance = this.getRecipientBalance.bind(this)
   }
 
   setOtherData(raw: any): void {
     this.otherData = asStellarWalletOtherData(raw)
   }
 
+  async getRecipientBalance(recipient: string): Promise<string> {
```
swansontec, line 78: If you use an arrow function `getRecipientBalance = async () => {}`, you can drop the `bind` call in the constructor.

src/polkadot/polkadotEngine.ts
```diff
@@ -63,6 +63,15 @@ export class PolkadotEngine extends CurrencyEngine<PolkadotTools> {
     this.otherData = asPolkadotWalletOtherData(raw)
   }
 
+  async getRecipientBalance(recipient: string): Promise<string> {
```
swansontec, line 66: You probably want an arrow function here, so `this` works.

src/xrp/xrpEngine.ts
```diff
@@ -106,6 +117,19 @@ export class XrpEngine extends CurrencyEngine<RippleTools> {
     this.otherData = asXrpWalletOtherData(raw)
   }
 
+  async getRecipientBalance(recipient: string): Promise<string> {
```
swansontec, line 120: You probably want an arrow function here, so `this` works.

swansontec (changes requested):

swansontec (approved):

PR #529, opened 2023-03-14
Refactor FIO use makeSpend, signTx, and broadcastTx instead of ambiguous otherMethods

peachbits:
---
  - https://app.asana.com/0/0/1204112329611718


swansontec (approved):
Only one trivial thing here. I am also not 100% happy with forking the upstream repo, so please be sure they have a PR created to include types in their bundle, so we can go back to vanilla.

PR #533, opened 2023-03-22
Allow engines to run with only pubkeys

samholmes:
### CHANGELOG

- changed: Refactor all engines to only deal with private keys directly from privelaged functions

### Dependencies

- https://github.com/EdgeApp/edge-core-js/pull/508

### Description

 none


src/ethereum/ethTypes.ts
```diff
@@ -496,3 +505,34 @@ export interface EthereumOtherMethods {
...
+        [`${pluginId}Key`]: asString
+      })(value)
+      const to = {
+        mnemonic: from[`${pluginId}Mnemonic`],
+        privateKey: from[`${pluginId}Key`]
+      }
+      return asObject({
+        mnemonic: asString,
+        privateKey: asString
+      })(to)
```
swansontec, line 530: Apparently, according to the block of code that got deleted in the Ethereum engine constructor, there are keys out there that are double-nested. Something like this:

```
{
  type: "wallet:ethereum",
  id: "...",
  keys: {
    keys: {
      ethereumKey: '0x...'
    }
  }
}
```

So I think we should preserve the ability to read these jacked-up keys by fixing them in the cleaner.

src/fio/fioEngine.ts
```diff
@@ -107,13 +114,15 @@ export class FioEngine extends CurrencyEngine<FioTools> {
     opts: EdgeCurrencyEngineOptions,
     tpid: string
   ) {
-    super(env, tools, walletInfo, opts)
+    const safeWalletInfo = asSafeFioWalletInfo(walletInfo)
+    const fioPrivateKeys = asFioPrivateKeys(walletInfo.keys)
+    super(env, tools, safeWalletInfo, opts)
```
swansontec, line 119: The constructor won't get the private keys anymore, so this is no good. Are we still waiting to merge Matthew's FIO refactor before we can fix this?

src/hedera/hederaTypes.ts
```diff
@@ -84,3 +88,34 @@ export const asCheckAccountCreationStatus = asObject({
...
   account_id: asOptional(asString)
 })
+
+export type SafeHbarWalletInfo = ReturnType<typeof asSafeHbarWalletInfo>
+export const asSafeHbarWalletInfo = asSafeCommonWalletInfo
+
+export interface HbarPrivateKeys {
+  mnemonic?: string
+  privateKey: string
+}
```
swansontec, line 98: You call these types "Hbar", but we use "Hedera" almost everywhere else.

While we haven't been 100% consistent on this, I think we need to name things after the network, not the currency. This is because people can split and clone networks, creating lots of currencies (ETH / ETC), but the engine stays the same. So naming types after the coin is short-sighted if we ever need to implement another hash-graph fork with a competing coin.

src/polkadot/polkadotEngine.ts
```diff
@@ -416,9 +428,17 @@ export class PolkadotEngine extends CurrencyEngine<PolkadotTools> {
...
-      )
+      if (dotPrivateKeys.mnemonic != null) {
+        this.keypair.addFromUri(dotPrivateKeys.mnemonic)
+      } else {
+        const byteArray = []
+        for (let i = 0; i < dotPrivateKeys.privateKey.length; i += 2) {
+          const hexPair = dotPrivateKeys.privateKey.substr(i, 2)
+          byteArray.push(parseInt(hexPair, 16))
+        }
+        const uint8Array = Uint8Array.from(byteArray)
```
swansontec, line 439: Is this a hex to Unit8Array conversion? Then use `import base16 from 'rfc4648'` / `base16.parse(hex)`

src/polkadot/polkadotTypes.ts
```diff
@@ -53,3 +57,34 @@ export const asTransactions = asObject({
...
 })
+
+export type SafeDotWalletInfo = ReturnType<typeof asSafeDotWalletInfo>
+export const asSafeDotWalletInfo = asSafeCommonWalletInfo
+
+export interface DotPrivateKeys {
+  mnemonic?: string
+  privateKey: string
+}
+export const asDotPrivateKeys = (pluginId: string): Cleaner<DotPrivateKeys> =>
```
swansontec, line 68: s/Dot/Polkadot/

src/solana/solanaEngine.ts
```diff
@@ -382,18 +388,19 @@ export class SolanaEngine extends CurrencyEngine<SolanaTools> {
...
       throw new Error('Missing unsignedSerializedSolTx')
 
     const keypair = Keypair.fromSecretKey(
-      Uint8Array.from(
-        Buffer.from(
-          this.walletInfo.keys[`${this.currencyInfo.pluginId}Key`],
-          'hex'
-        )
-      )
+      Uint8Array.from(Buffer.from(solPrivateKeys.privateKey, 'hex'))
```
swansontec, line 403: A `Buffer` is a `Uint8Array` already, so no need to convert. But also, we should just use `base16.parse` here too.

src/solana/solanaTypes.ts
```diff
@@ -62,3 +66,34 @@ export const asRecentBlockHash = asObject({
...
 })
+
+export type SafeSolWalletInfo = ReturnType<typeof asSafeSolWalletInfo>
+export const asSafeSolWalletInfo = asSafeCommonWalletInfo
+
+export interface SolPrivateKeys {
+  mnemonic: string
+  privateKey: string
+}
+export const asSolPrivateKeys = (pluginId: string): Cleaner<SolPrivateKeys> =>
```
swansontec, line 77: s/Sol/Solana/ maybe here.

src/stellar/stellarTypes.ts
```diff
@@ -83,3 +85,11 @@ export const asFeeStats = asObject({
...
   })
 })
+
+export type SafeXlmWalletInfo = ReturnType<typeof asSafeXlmWalletInfo>
+export const asSafeXlmWalletInfo = asSafeCommonWalletInfo
+
+export type XlmPrivateKeys = ReturnType<typeof asXlmPrivateKeys>
+export const asXlmPrivateKeys = asObject({
+  stellarKey: asString
+})
```
swansontec, line 95: s/Xlm/Stellar/

src/tezos/tezosTypes.ts
```diff
@@ -123,3 +125,17 @@ export interface OperationsContainer {
...
+    publicKey: asString,
+    publicKeyEd: asString
+  })
+)
+
+export type XtzPrivateKeys = ReturnType<typeof asXtzPrivateKeys>
+export const asXtzPrivateKeys = asObject({
+  mnemonic: asString,
+  privateKey: asString
+})
```
swansontec, line 141: s/Xtz/Tezos/

src/tron/tronEngine.ts
```diff
@@ -1329,12 +1334,13 @@ export class TronEngine extends CurrencyEngine<TronTools> {
     return edgeTransaction
   }
 
-  getDisplayPrivateSeed(): string {
-    return this.walletInfo.keys?.tronMnemonic ?? this.walletInfo.keys?.tronKey
+  getDisplayPrivateSeed(privateKeys: JsonObject): string | null {
+    const tronPrivateKeys = asTronPrivateKeys(privateKeys)
+    return tronPrivateKeys.tronMnemonic ?? tronPrivateKeys.tronKey
```
swansontec, line 1339: I think this structure would make sense for the other plugins where we currently return "" if the mnemonic is missing. We should always return something.

src/tron/tronTypes.ts
```diff
@@ -423,3 +413,17 @@ export const asBroadcastResponse = asObject({
   txid: asString,
   message: asString
 })
+
+export type SafeTrxWalletInfo = ReturnType<typeof asSafeTrxWalletInfo>
+export const asSafeTrxWalletInfo = asSafeCommonWalletInfo
```
swansontec, line 418: s/Trx/Tron/

src/xrp/xrpTypes.ts
```diff
@@ -32,3 +34,11 @@ export const asXrpNetworkLocation = asObject({
...
   issuer: asString
 })
+
+export type SafeXrpWalletInfo = ReturnType<typeof asSafeXrpWalletInfo>
+export const asSafeXrpWalletInfo = asSafeCommonWalletInfo
+
+export type XrpPrivateKeys = ReturnType<typeof asXrpPrivateKeys>
+export const asXrpPrivateKeys = asObject({
+  rippleKey: asString
+})
```
swansontec, line 44: This is the one exception to s/Xrp/Ripple/, since everything else is already messed up and using the wrong name anyhow.

src/zcash/zecTypes.ts
```diff
@@ -114,3 +123,48 @@ export interface ZcashSynchronizer {
...
+
+export type SafeZecWalletInfo = ReturnType<typeof asSafeZecWalletInfo>
+export const asSafeZecWalletInfo = asWalletInfo(asZecPublicKey)
+
+export interface ZecPrivateKeys {
+  mnemonic: string
+  spendKey: string
+  birthdayHeight: number
+}
+export const asZecPrivateKeys = (pluginId: string): Cleaner<ZecPrivateKeys> =>
```
swansontec, line 144: s/Zec/Zcash/

test/engine/engine.test.ts
```diff
@@ -102,31 +100,25 @@ for (const fixture of fixtures) {
...
-        id: '1',
-        type: WALLET_TYPE,
-        // @ts-expect-error
-        keys
-      }
       // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
       if (!plugin) throw new Error('ErrorNoPlugin')
       return await plugin
-        .makeCurrencyEngine(info, currencyEngineOptions)
+        .makeCurrencyEngine(privateWalletInfo, currencyEngineOptions)
```
swansontec, line 121: Should we pass in a public wallet info, using the `publicKey` above, to make this a better test?

swansontec (changes requested):

src/binance/bnbEngine.ts
```diff
@@ -7,8 +7,9 @@ import {
   EdgeTransaction,
   EdgeWalletInfo,
   InsufficientFundsError,
+  JsonObject,
   NoAmountSpecifiedError
-} from 'edge-core-js/types'
+} from 'edge-core-js'
```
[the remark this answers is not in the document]
samholmes: Did you mean this: https://github.com/EdgeApp/edge-currency-accountbased/pull/534 ?

src/fio/fioEngine.ts
samholmes: Yes, this is still blocked by Matt's work.

src/hedera/hederaTypes.ts
samholmes: But, that competing coin would not be called Hedera either, so then why would this be any better?

src/tron/tronEngine.ts
samholmes: What are you asking for here? This actuall will always return a string.

test/engine/engine.test.ts
samholmes: Yes, but FIO is blocking this from passing at least for now.

src/ethereum/ethTypes.ts
```diff
@@ -513,21 +513,37 @@ export interface EthPrivateKeys {
...
+  // Derived cleaners from the generic parameter:
+  const asFromKeys = asObject({
+    [`${pluginId}Mnemonic`]: asString,
+    [`${pluginId}Key`]: asString
+  })
+  const asFromJackedKeys = asObject({ keys: asFromKeys })
+  const asToKeys = asObject({
+    mnemonic: asString,
+    privateKey: asString
+  })
```
swansontec, line 528: This `asToKeys` isn't needed - TypeScript is fully aware that our return type is `EthPrivateKeys`, and will complain if we fail to match that expectation. You have already cleaned the raw input data, so that side is safe.

The rest of the changes here look good. I appreciate the humor in the variable naming.

swansontec (approved):
Approving, because the one remaining issue isn't that bad.

src/tron/tronEngine.ts
swansontec: The `getDisplayPrivateSeed` should never return `null` or `''` if it can help it. Tron is doing things the right way, since it always returns something useful, but there are other plugins in this repo where we do return `''` as a fallback. We should find those and fix them.

src/hedera/hederaTypes.ts
swansontec: For instance, if we call the Ethereum stuff EthEngine and so forth, then it wouldn't make sense for Ethereum Classic, since that's "ETC", not "ETH". However, it's still an Ethereum fork, and still uses things like "ERC20" for its tokens. So the network name is more sticky than the token code.

Also, all the filenames, tools, engines, and so forth already use the network name, so we should be consistent with that.

src/ethereum/ethTypes.ts
samholmes: Ah, you're right. Got caried away with cleaners.

Time to decision is a valuable resource. Naming is hard; you made it easier. :)

PR #535, opened 2023-03-30
FIO upgrade cleanups

peachbits:
---
  - https://app.asana.com/0/0/1204295865014355

### DIFF
--- a/src/fio/fioEngine.ts
+++ b/src/fio/fioEngine.ts
@@ -101,7 +101,6 @@ import {
 const ADDRESS_POLL_MILLISECONDS = 10000
 const BLOCKCHAIN_POLL_MILLISECONDS = 15000
 const TRANSACTION_POLL_MILLISECONDS = 10000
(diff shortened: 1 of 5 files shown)


src/fio/fioEngine.ts
```diff
@@ -894,6 +895,22 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
...
+        (result: any) => {
+          const errorResponse = asMaybe(asFioEmptyResponse)(result)
+          if (errorResponse != null) return JSON.stringify(errorResponse)
+          return comparisonFioBalanceString(result)
+        },
+        2
+      )
+      if (res?.data?.json?.message === 'Public key not found') {
+        res = { balance: '0', available: '0', staked: '0', srps: '0', roe: '' }
+      }
```
swansontec, line 913: You can use the same cleaner idea here too.

swansontec (approved):
I'm approving, but we could do more cleaning.

swansontec (approved):

PR #537, opened 2023-03-31
Add Algorand

peachbits:
---
  - https://app.asana.com/0/0/1204310909543305

### DIFF
--- a/src/algorand/AlgorandTools.ts
+++ b/src/algorand/AlgorandTools.ts
@@ -139,4 +139,4 @@ export async function makeCurrencyTools(
   return new AlgorandTools(env)
 }
 
-export { makeCurrencyEngine } from './algorandEngine'
+export { makeCurrencyEngine } from './AlgorandEngine'
--- a/src/algorand/algorandEngine.ts
+++ b/src/algorand/algorandEngine.ts
@@ -367,7 +367,9 @@ export class AlgorandEngine extends CurrencyEngine<
     let balance = this.getBalance({
       currencyCode: spendInfo.currencyCode
     })
-    balance = sub(balance, this.networkInfo.minimumAddressBalance)
+    if (spendInfo.currencyCode === this.currencyInfo.currencyCode) {
+      balance = sub(balance, this.networkInfo.minimumAddressBalance)
+    }
 
     const publicAddress = spendInfo.spendTargets[0].publicAddress
     if (publicAddress == null) {
@@ -410,7 +412,8 @@ export class AlgorandEngine extends CurrencyEngine<
   }
 
   async makeSpend(edgeSpendInfoIn: EdgeSpendInfo): Promise<EdgeTransaction> {
-    const { edgeSpendInfo, currencyCode } = this.makeSpendCheck(edgeSpendInfoIn)
+    const { edgeSpendInfo, currencyCode, nativeBalance } =
+      this.makeSpendCheck(edgeSpendInfoIn)
 
     const spendableAlgoBalance = sub(
       this.getBalance({
@@ -435,7 +438,9 @@ export class AlgorandEngine extends CurrencyEngine<
 
     const { type }: BaseTxOpts = asMaybe(asBaseTxOpts)(
       edgeSpendInfo.otherParams
-    ) ?? { type: 'pay' }
+    ) ?? {
+      type: currencyCode === this.currencyInfo.currencyCode ? 'pay' : 'axfer'
+    }
 
     let note: Uint8Array | undefined
     if (memo != null) {
@@ -448,6 +453,7 @@ export class AlgorandEngine extends CurrencyEngine<
     let rawTx: Transaction
     let nativeAmount = amount
     let networkFee = '0'
+    let parentNetworkFee: string | undefined
     let fee: string
     let recipient: string | undefined
     switch (type) {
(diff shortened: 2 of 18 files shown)


src/algorand/algorandEngine.ts
```diff
@@ -113,10 +113,31 @@ export class AlgorandEngine extends CurrencyEngine<
...
+          const tokenId = Object.keys(this.allTokensMap).find(
+            tokenId => this.allTokensMap[tokenId].currencyCode === cc
+          )
+          if (tokenId == null) {
+            this.updateBalance(cc, '0')
+            continue
+          }
+
+          const asset = assets.find(
+            asset => asset['asset-id'] === parseInt(tokenId)
```
swansontec, line 145: Would it make sense to do `asset['asset-id'].toFixed() === tokenId`

src/algorand/algorandTypes.ts
```diff
@@ -142,6 +142,12 @@ export const asMaybeCustomFee = asMaybe(
   { fee: undefined }
 )
 
+export const asMaybeAssetIndexLocation = asMaybe(
+  asObject({
+    assetIndex: asString
```
swansontec: Nothing prevents the index from being negative or floating point or whatever. Even the `getTokenId` checks don't handle these cases. Instead, steal this cleaner from edge-core-js and use it for `assetIndex`:

```typescript
export function asIntegerString(raw: unknown): string {
  const clean = asString(raw)
  if (!/^\d+$/.test(clean)) {
    throw new Error('Expected an integer string')
  }
  return clean
}
```

src/algorand/algorandEngine.ts
```diff
@@ -470,6 +476,41 @@ export class AlgorandEngine extends CurrencyEngine<
 
         break
       }
+      case 'axfer': {
+        const edgeTokenId = Object.keys(this.allTokensMap).find(
+          tokenId => this.allTokensMap[tokenId].currencyCode === currencyCode
+        )
+        if (edgeTokenId == null) throw new Error('Unrecognized asset')
+        const assetIndex: string | undefined =
```
swansontec: You should clean the network location before accessing it.

src/algorand/algorandEngine.ts
```diff
@@ -470,6 +476,41 @@ export class AlgorandEngine extends CurrencyEngine<
...
+          this.allTokensMap?.[edgeTokenId]?.networkLocation?.assetIndex
+        if (assetIndex == null) throw new Error('Unrecognized asset')
+
+        rawTx = algosdk.makeAssetTransferTxnWithSuggestedParamsFromObject({
+          assetIndex: parseInt(assetIndex),
+          to: publicAddress,
+          from: this.walletInfo.keys.publicKey,
+          amount: BigInt(amount),
```
swansontec, line 518: This library requires native browser BigInt support to function? Do we need checks for this?

src/algorand/algorandEngine.ts
```diff
@@ -604,6 +626,81 @@ export class AlgorandEngine extends CurrencyEngine<
...
+        throw new Error(`Invalid tokenId to activate ${activateTokenId}`)
+
+      const spendInfo: EdgeSpendInfo = {
+        currencyCode: this.allTokensMap[activateTokenId].currencyCode,
+        skipChecks: true,
+        spendTargets: [
+          { publicAddress: this.walletInfo.keys.publicKey, nativeAmount: '0' }
+        ] // Activation ("opt-in") transactions are just asset transfers with a zero amount
+      }
+      const approvalTx = await this.makeSpend(spendInfo)
```
swansontec, line 674: As mentioned in the GUI PR, I would like to move the `catch` logic here so we still return a quote, even if we have insufficient funds for the `makeSpend` call.

src/common/engine.ts
```diff
@@ -974,7 +974,7 @@ export class CurrencyEngine<
     checkEdgeSpendInfo(edgeSpendInfo)
 
     for (const st of edgeSpendInfo.spendTargets) {
-      if (st.publicAddress === this.walletLocalData.publicKey) {
+      if (!skipChecks && st.publicAddress === this.walletLocalData.publicKey) {
```
swansontec, line 977: This affects a lot of coins. However, do I see how it does go along with the variable name. I guess it's fine?

swansontec (changes requested):

src/algorand/algorandTypes.ts
```diff
@@ -0,0 +1,37 @@
...
+    (value: unknown) => {
+      const from = asObject({
+        [`${pluginId}Mnemonic`]: asString
+      })(value)
+      const to = {
+        mnemonic: from[`${pluginId}Mnemonic`]
+      }
+      return asObject({
+        mnemonic: asString
+      })(to)
```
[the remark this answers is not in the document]
peachbits: I copy/pasted from another plugin. Turns out you left the same comment on those cleaners but the fix never made it into the committed branch

src/algorand/algorandEngine.ts
```diff
@@ -0,0 +1,117 @@
...
+  asSafeAlgorandWalletInfo,
+  SafeAlgorandWalletInfo
+} from './algorandTypes'
+
+export class AlgorandEngine extends CurrencyEngine<
+  AlgorandTools,
+  SafeAlgorandWalletInfo
+> {
+  otherData!: AlgorandWalletOtherData
+  networkInfo: {}
```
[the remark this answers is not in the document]
peachbits: We use the rule eslint/no-empty-interface which prevents that, but I did notice after reviewing a PR from you (after this was started) that you've put dummy data inside and delete it when actually adding keys

src/algorand/algorandEngine.ts
peachbits: That's handled in the checkEnvironment function defined in the info file same as Polkadot

src/common/engine.ts
peachbits: I thought about adding a `allowSpendToSelf` but I kind of like `skipChecks` as it should mean override the common checking and allow the engine to handle it

src/algorand/algorandTypes.ts
```diff
@@ -155,7 +155,7 @@ export const asMaybeCustomFee = asMaybe(
 
 export const asMaybeContractAddressLocation = asMaybe(
   asObject({
-    contractAddress: asString
+    contractAddress: asIntegerString
```
swansontec, line 158: Now that you are doing this, you can delete the `isNaN` check in AlgorandTools.ts.

src/algorand/AlgorandTools.ts
swansontec, line 1: You forgot to update this line: `export { makeCurrencyEngine } from './algorandEngine'`

It kinda-sorta works on a Mac because the filesystem is case-insensitive, but will fail in Travis.

src/common/engine.ts
swansontec: Sounds good.

swansontec (changes requested):
A few small things are left

swansontec (approved):

PR #538, opened 2023-04-04
Sam/fix create eos

samholmes:
### CHANGELOG

- fixed: Crash in `createPrivateKey` for `EosTools`

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204323976548769

### DIFF
--- a/src/eos/eosEngine.ts
+++ b/src/eos/eosEngine.ts
@@ -198,15 +198,6 @@ export class EosEngine extends CurrencyEngine<EosTools, SafeEosWalletInfo> {
     opts: EdgeCurrencyEngineOptions
   ): Promise<void> {
     await super.loadEngine(plugin, walletInfo, opts)
-
-    if (typeof this.walletInfo.keys.ownerPublicKey !== 'string') {
-      if (walletInfo.keys.ownerPublicKey != null) {
-        this.walletInfo.keys.ownerPublicKey = walletInfo.keys.ownerPublicKey
-      } else {
-        const pubKeys = await plugin.derivePublicKey(this.walletInfo)
-        this.walletInfo.keys.ownerPublicKey = pubKeys.ownerPublicKey
-      }
-    }
   }
 
   // Poll on the blockheight
--- a/src/eos/eosPlugin.ts
+++ b/src/eos/eosPlugin.ts
@@ -96,24 +96,22 @@ export class EosTools implements EdgeCurrencyTools {
   }
 
   async derivePublicKey(walletInfo: EdgeWalletInfo): Promise<Object> {
-    const type = walletInfo.type.replace('wallet:', '')
-    const currencyInfoType = this.currencyInfo.walletType.replace('wallet:', '')
-    if (type === currencyInfoType) {
-      const publicKey = PrivateKey.from(walletInfo.keys.eosKey)
-        .toPublic()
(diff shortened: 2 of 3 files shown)


src/eos/eosPlugin.ts
```diff
@@ -79,10 +81,16 @@ export class EosTools implements EdgeCurrencyTools {
 
     const currencyInfoType = this.currencyInfo.walletType.replace('wallet:', '')
     if (type === currencyInfoType) {
-      let entropy = Buffer.from(this.io.random(32)).toString('hex')
-      const eosOwnerKey = PrivateKey.from(entropy).toWif()
-      entropy = Buffer.from(this.io.random(32)).toString('hex')
-      const eosKey = PrivateKey.from(entropy).toWif()
+      const ownerKeyEntropy = Buffer.from(this.io.random(32)).toString('hex')
```
swansontec, line 84: The bytes / buffer / hex conversion is a bit weird here. Can their `Bytes` class just take a Uint8Array, as returned by the `random` function? If we really need to go hex, use `base16` from rfc4648. Maybe also put the whole random -> Bytes thing on one line, so we aren't converting halfway on one line, and then converting more on the next.

src/eos/eosEngine.ts
```diff
@@ -198,15 +198,6 @@ export class EosEngine extends CurrencyEngine<EosTools, SafeEosWalletInfo> {
...
     await super.loadEngine(plugin, walletInfo, opts)
-
-    if (typeof this.walletInfo.keys.ownerPublicKey !== 'string') {
-      if (walletInfo.keys.ownerPublicKey != null) {
-        this.walletInfo.keys.ownerPublicKey = walletInfo.keys.ownerPublicKey
-      } else {
-        const pubKeys = await plugin.derivePublicKey(this.walletInfo)
-        this.walletInfo.keys.ownerPublicKey = pubKeys.ownerPublicKey
-      }
-    }
```
swansontec, line 209: Ahhh, this was before the days where the core did the public key derivation for us. Yes, remove it.

swansontec (approved):

src/eos/eosPlugin.ts
```diff
@@ -98,24 +98,22 @@ export class EosTools implements EdgeCurrencyTools {
...
-      if (walletInfo.keys.eosOwnerKey != null) {
-        ownerPublicKey = PrivateKey.from(walletInfo.keys.eosOwnerKey)
-          .toPublic()
-          .toLegacyString()
-      }
-      return { publicKey, ownerPublicKey }
-    } else {
-      throw new Error('InvalidWalletType')
     }
+    return { publicKey, ownerPublicKey }
```
swansontec, line 116: Hide whitespace to view this one. It's another cleanup we should do while we were in there - I wanted to be sure your PR made sense, but this guy was all weird.

src/eos/eosEngine.ts
samholmes: This code predates `derivePublicKey`?

swansontec: I believe so, yes. Early on, engines would just get the private keys, and were on their own in terms of deriving public keys. This was slow, and everybody did it differently, so we moved it to the core. A few old coins still do it themselves, redundantly.

PR #543, opened 2023-04-07
Don't save network fees to disk

peachbits:
---
  - https://app.asana.com/0/0/1204351102858700


swansontec (approved):
This looks correct, assuming the goal even makes sense.

PR #545, opened 2023-04-13
Paul/is send

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204343038506600


swansontec (approved):
This looks right to me.

src/algorand/AlgorandEngine.ts
```diff
@@ -612,7 +612,7 @@ export class AlgorandEngine extends CurrencyEngine<
       const { txId } = await asyncWaterfall(
         this.networkInfo.algodServers.map(server => async () => {
           const client = new Algodv2('', server, '')
-          return await client.sendRawTransaction(signedTxBytes).do()
+          return client.sendRawTransaction(signedTxBytes).do()
```
swansontec, line 618: @typescript-eslint/return-await

paullinator: Weird. keeping the line seems to cause a lint error on my machine. Possibly a vscode bug. I'll revert

PR #547, opened 2023-04-14
Sam/zksync

samholmes:
### CHANGELOG

- Add zkSync wallet type

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204276716902890

### DIFF
--- a/src/ethereum/info/zksyncInfo.ts
+++ b/src/ethereum/info/zksyncInfo.ts
@@ -1,0 +1,113 @@
+import { EdgeCurrencyInfo, EdgeTokenMap } from 'edge-core-js/types'
+
+import { makeOuterPlugin } from '../../common/innerPlugin'
+import { makeMetaTokens } from '../../common/tokenHelpers'
+import type { EthereumTools } from '../ethPlugin'
+import type { EthereumFees, EthereumNetworkInfo } from '../ethTypes'
+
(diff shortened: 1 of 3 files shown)


src/ethereum/info/zksyncInfo.ts
```diff
@@ -0,0 +1,110 @@
...
+  amberDataBlockchainId: '',
+  pluginMnemonicKeyName: 'zksyncMnemonic',
+  pluginRegularKeyName: 'zksyncKey',
+  ethGasStationUrl: null,
+  defaultNetworkFees
+}
+
+const defaultSettings: any = {
+  customFeeSettings: ['gasLimit', 'gasPrice'],
+  otherSettings: { ...networkInfo }
```
swansontec, line 74: I would like to delete `otherSettings` from all the plugins, but since this plugin is new, maybe we shouldn't add it in the first place.

Unfortunately, there are two places where the GUI still looks at this field. So, maybe we can do the minimal amount of damage possible while still keeping the GUI happy:

```ts
otherSettings: {
  chainParams: networkInfo.chainParams,
  ercTokenStandard: networkInfo.ercTokenStandard,
}
```

swansontec (approved):
Approving, with one optional cleanup that would help reduce the amount of API surface area we are leaking to the GUI.

PR #548, opened 2023-04-18
Update cleaners and audit balance and transaction syncing statuses

peachbits:
---
  - https://app.asana.com/0/0/1204386783187194

### DIFF
--- a/src/binance/bnbEngine.ts
+++ b/src/binance/bnbEngine.ts
@@ -146,14 +146,12 @@ export class BinanceEngine extends CurrencyEngine<
         }
       }
     } catch (e: any) {
-      // fetching of account balances for uninitiated accounts returns 404 (throws error)
-      if (
-        this.tokenCheckTransactionsStatus.BNB === 1 &&
-        this.transactionList.BNB.length === 0
-      ) {
+      if (/error code 404/.test(e?.message)) {
+        // fetching of account balances for uninitiated accounts returns 404 (throws error)
         this.updateBalance('BNB', '0')
+      } else {
+        this.error(`Error checking BNB address balance ${String(e)}`)
       }
-      this.error(`Error checking BNB address balance`)
     }
   }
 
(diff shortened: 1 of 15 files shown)


src/binance/bnbEngine.ts
```diff
@@ -146,14 +146,12 @@ export class BinanceEngine extends CurrencyEngine<
...
-      // fetching of account balances for uninitiated accounts returns 404 (throws error)
-      if (
-        this.tokenCheckTransactionsStatus.BNB === 1 &&
-        this.transactionList.BNB.length === 0
-      ) {
+      if (e.message?.includes('error code 404') === true) {
+        // fetching of account balances for uninitiated accounts returns 404 (throws error)
         this.updateBalance('BNB', '0')
+      } else {
+        this.error(`Error checking BNB address balance ${e.message}`)
```
swansontec, line 153: If `e` happens to be `null`, this error handling will crash. You could do `asObject({ message: asString })` to check the error before accessing the message, or you could do `/error code 404/.test(e?.message)`, which doesn't rely on the message existing or even being a string.

src/eos/eosEngine.ts
```diff
@@ -765,6 +765,22 @@ export class EosEngine extends CurrencyEngine<EosTools, SafeEosWalletInfo> {
           this.currencyEngineCallbacks.onAddressChanged()
         }
       }
+    } catch (e: any) {
+      if (e.message?.includes('get_account failed with 400') === true) {
+      // dfuse servers will return 400 for new unused accounts and can be ignored
+      } else {
+        this.error(`getKeyAccounts error: `, e)
+      }
+    }
```
swansontec: Same thing, with needing to check that the catch actually got an Error object.

swansontec (changes requested):
The cleanups look good, aside from the way we access error messages without checking.

src/binance/bnbEngine.ts
swansontec: For the line `this.error('Error checking BNB address balance ${e.message}')`, it to `this.error('Error checking BNB address balance ${String(e)}')`. That will turn the error object into a string more safely.

swansontec (approved):

PR #550, opened 2023-04-19
Add ETH tokens

peachbits:


swansontec (commented):
These additions seem fine, but it doesn't look like you got all the tokens in the list. For instance, I can't find ARDR in either the existing token list or this new list. Is there some reason for the holes?

swansontec (approved):

PR #552, opened 2023-04-20
Add checksum address for `getFreshAddress` to ethereum plugin

samholmes:
### CHANGELOG

- Added: ERC-55 checksum address returned by `getFreshAddress` for ethereum plugin

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204101187072518

### DIFF
--- a/src/ethereum/ethEngine.ts
+++ b/src/ethereum/ethEngine.ts
@@ -8,6 +8,7 @@ import {
   EdgeCurrencyEngineOptions,
   EdgeCurrencyInfo,
   EdgeFetchFunction,
+  EdgeFreshAddress,
   EdgeSignMessageOptions,
   EdgeSpendInfo,
   EdgeSpendTarget,
@@ -645,6 +646,20 @@ export class EthereumEngine extends CurrencyEngine<
     await this.killEngine()
     await this.clearBlockchainCache()
     await this.startEngine()
+  }
+
+  async getFreshAddress(): Promise<EdgeFreshAddress> {
+    const legacyAddress = this.walletLocalData.publicKey.toLowerCase()
+
+    const { publicKey } = this.walletLocalData
+    const publicAddress = /[A-F]/.test(publicKey)
+      ? publicKey
+      : EthereumUtil.toChecksumAddress(publicKey.replace('0x', ''))
+
+    return {
+      legacyAddress,
+      publicAddress
+    }
   }
 
   async getMaxSpendable(spendInfo: EdgeSpendInfo): Promise<string> {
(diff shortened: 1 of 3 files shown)


src/ethereum/ethPlugin.ts
```diff
@@ -97,6 +103,14 @@ export class EthereumTools implements EdgeCurrencyTools {
     }
   }
 
+  async checkPublicKey(publicKey: JsonObject): Promise<boolean> {
+    const ethereumPublicKey = asMaybe(asEthereumPublicKey)(publicKey)
+    if (ethereumPublicKey == null) return false
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    if (!EthereumUtil.isValidChecksumAddress(ethereumPublicKey)) return false
```
swansontec, line 110: Does this pass with all-lowercase addresses? If so, maybe we could comment on that.

swansontec (approved):

swansontec (changes requested):
After talking on the phone, let's not break the keys on disk. Instead:

- New wallets can be created with checksum addresses.
- Old wallets can retain the lowercase addresses.

The `getFreshAddress` and `getDisplayPublicKey` routines can lowercase / uppercase as needed, so the user never knows the difference.

src/ethereum/ethEngine.ts
```diff
@@ -647,6 +648,23 @@ export class EthereumEngine extends CurrencyEngine<
...
+  async getFreshAddress(_options: any): Promise<EdgeFreshAddress> {
+    const legacyAddress = this.walletLocalData.publicKey.toLowerCase()
+
+    const { publicKey } = this.walletLocalData
+    const publicAddress = /[A-F]/.test(publicKey)
+      ? publicKey
+      : ((): string => {
+          const addressHex = publicKey.replace('0x', '')
+          return EthereumUtil.toChecksumAddress(addressHex)
+        })()
```
swansontec, line 660: This IIFE isn't needed. If you do it the "obvious" way, `? publicKey : EthereumUtil.toChecksumAddress(publicKey.replace('0x', ''))`, the right-hand-side won't be evaluated if the ternary fails.

If this is about formatting, you can just move the `const addressHex = publicKey.replace('0x', '')` line outside the ternary, and not worry about it too much. It's going to be pretty cheap, even if not needed.

src/ethereum/ethEngine.ts
```diff
@@ -647,6 +648,23 @@ export class EthereumEngine extends CurrencyEngine<
     await this.startEngine()
   }
 
+  async getFreshAddress(_options: any): Promise<EdgeFreshAddress> {
```
swansontec, line 651: Do `EdgeGetReceiveAddressOptions` instead of `any` here.

swansontec (changes requested):
Oh, this is much cleaner now! I wasn't expecting it to be this much simpler. Do we need to adjust `getDisplayPublicKey` too?

swansontec (approved):

PR #554, opened 2023-04-21
Add wallet connect v1 to Algorand

peachbits:
Wallet connect v1 functionality is directly managed by the engines. This PR updates the existing EVM integration to return more generic params (native amount, network fee, and tokenId) first and then adds wallet connect to Algorand. It's largely a copy/paste of the EVM handling with tweaks to handle the unique ALGO call_request response. 

Futher cleanups and optimizations were not attempted because wallet connect v1 is due to sunset at the end of June 2023.

---
  - https://app.asana.com/0/0/1204452616785593

### DIFF
--- a/src/algorand/AlgorandEngine.ts
+++ b/src/algorand/AlgorandEngine.ts
@@ -211,8 +211,7 @@ export class AlgorandEngine extends CurrencyEngine<
         this.tools.walletConnectors[uri].walletId = walletId
       },
       wcDisconnect: (uri: string) => {
-        // eslint-disable-next-line @typescript-eslint/no-floating-promises
-        this.tools.walletConnectors[uri].connector.killSession()
+        this.tools.walletConnectors[uri].connector.killSession().catch(() => {})
         // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
         delete this.tools.walletConnectors[uri]
       },
--- a/src/algorand/AlgorandEngine.ts
+++ b/src/algorand/AlgorandEngine.ts
@@ -575,6 +575,19 @@ export class AlgorandEngine extends CurrencyEngine<
     return edgeTransaction
   }
 
+  async signMessage(message: string, privateKeys: JsonObject): Promise<string> {
+    const algorandPrivateKeys = asAlgorandPrivateKeys(
+      this.currencyInfo.pluginId
+    )(privateKeys)
+
+    const rawTx = decodeUnsignedTransaction(base64.parse(message))
+    const secretKey = algosdk.mnemonicToSecretKey(algorandPrivateKeys.mnemonic)
+    const signedTxBytes = rawTx.signTxn(secretKey.sk)
+    const signedTx = base64.stringify(signedTxBytes)
+
+    return signedTx
+  }
+
   async signTx(
     edgeTransaction: EdgeTransaction,
     privateKeys: JsonObject
--- a/src/algorand/AlgorandEngine.ts
+++ b/src/algorand/AlgorandEngine.ts
@@ -1,3 +1,4 @@
+import WalletConnect from '@walletconnect/client'
 import algosdk, {
   decodeUnsignedTransaction,
   encodeUnsignedTransaction,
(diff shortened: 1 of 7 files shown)


src/common/types.ts
```diff
@@ -82,3 +83,41 @@ export function asIntegerString(raw: unknown): string {
...
+    name: asString
+  }),
+  chainId: asOptional(asNumber, 1)
+})
+
+export type WcDappDetails = {
+  timeConnected: number
+} & ReturnType<typeof asWcDappDetails>
+
+export type Dapp = { timeConnected: number } & WcProps & WcDappDetails
```
swansontec, line 110: The `WcDappDetails` already include `timeConnected`, so it's redundant to have it here too?

src/algorand/AlgorandEngine.ts
```diff
@@ -95,6 +106,152 @@ export class AlgorandEngine extends CurrencyEngine<
...
+                resolve(dApp)
+              }
+            )
+
+            // Subscribe to call requests
+            connector.on(
+              'call_request',
+              (error: Error | null, payload: AlgoWcRpcPayload) => {
+                try {
+                  if (error != null) throw error
```
swansontec, line 156: This line seems weird to me. If we throw the error, it will go back into the wallet-connect SDK. Then what? Does the SDK swallow it, or does it come out somewhere else? Do we want to show it to the user?

Same for the other 4 callbacks in this block.

src/algorand/AlgorandEngine.ts
```diff
@@ -95,6 +106,152 @@ export class AlgorandEngine extends CurrencyEngine<
...
+                  const cleanPayload = asAlgorandWalletConnectPayload(payload)
+                  const params = cleanPayload.params[0][0]
+                  const algoTx = decodeUnsignedTransaction(
+                    base64.parse(params.txn)
+                  )
+
+                  const nativeAmount =
+                    algoTx.amount != null ? algoTx.amount.toString() : '0'
+                  const networkFee = algoTx.fee.toString()
```
swansontec, line 165: Would `toFixed` make more sense? If the number is really big, you might get "1e+30", which wouldn't be valid for biggystring.

src/algorand/AlgorandEngine.ts
```diff
@@ -95,6 +106,152 @@ export class AlgorandEngine extends CurrencyEngine<
...
+      wcConnect: (uri: string, publicKey: string, walletId: string) => {
+        this.tools.walletConnectors[uri].connector.approveSession({
+          accounts: [publicKey],
+          chainId: 416001 // Not actually required in runtime. Using the placeholder value found in another implementation.
+        })
+        this.tools.walletConnectors[uri].walletId = walletId
+      },
+      wcDisconnect: (uri: string) => {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        this.tools.walletConnectors[uri].connector.killSession()
```
swansontec, line 215: `.catch(()=>{})` would remove the need for the eslint-disable, and would make it clear that we don't care about the return value at all.

src/algorand/AlgorandEngine.ts
```diff
@@ -95,6 +106,152 @@ export class AlgorandEngine extends CurrencyEngine<
...
+        // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+        delete this.tools.walletConnectors[uri]
+      },
+      wcApproveRequest: async (
+        uri: string,
+        payload: AlgoWcRpcPayload,
+        result: string
+      ): Promise<void> => {
+        this.tools.walletConnectors[uri].connector.approveRequest({
+          id: parseInt(payload.id.toString()),
```
swansontec, line 225: Why are we doing number -> string -> number here? I feel like this maybe hiding a bug, unless it's just superfluous.

Same for the other wc methods that do this.

src/algorand/algorandTypes.ts
```diff
@@ -182,3 +191,55 @@ export const asAlgorandPrivateKeys = (
...
+    payload: AlgoWcRpcPayload,
+    result: string
+  ) => Promise<void>
+  wcRejectRequest: (uri: string, payload: AlgoWcRpcPayload) => Promise<void>
+  wcGetConnections: () => Dapp[]
+}
+
+export const asAlgorandWalletConnectPayload = asObject({
+  id: asNumber,
+  jsonrpc: asString,
```
swansontec, line 225: Would it make sense for this to be `asValue("2.0")`?

swansontec (changes requested):
I have a bunch of questions on this PR, since I'm not 100% sure if things are really wrong, or just look weird.

src/algorand/AlgorandEngine.ts
swansontec: Ah, I see you have `number | string` as the input, so yes, this could be a bug. If the SDK really needs `number` and nothing else, you could do `Number(payload.id)`. This will leave numbers as-is, but also parse strings into numbers if possible. This will make it clear that random strings really aren't going to work, if the dapp decides to use them.

src/algorand/AlgorandEngine.ts
peachbits: I don't have a good answer for this as it was copied from the ETH implementation and I haven't seen an error come through before. We'd have to create a special return type for the gui to parse and I don't think it's worthwhile if the v1 functionality is set to sunset in 2 months. The gui-first implementation of v2 should include this kind of error handling

swansontec: Fair enough.

swansontec (approved):

PR #557, opened 2023-04-26
Name files properly

swansontec:
### CHANGELOG

 none

### Dependencies

 none

### Description

- Use the network name, not the currency code. Most networks support multiple currencies.
- Use uppercase names for files that export classes.
- Use loweracse names for files that export types and utilities.

---
  - https://app.asana.com/0/0/1204464718377053


peachbits:
Approved, but I'm not hitting that button until some inflight stuff

peachbits (approved):

PR #564, opened 2023-05-05
ZEC/ARRR Handle import key birthday height

peachbits:
---
  - https://app.asana.com/0/0/1204535089316845

### DIFF
--- a/src/zcash/zecPlugin.ts
+++ b/src/zcash/zecPlugin.ts
@@ -18,6 +18,7 @@ import {
 } from 'react-native-zcash'
 
 import { PluginEnvironment } from '../common/innerPlugin'
+import { asIntegerString } from '../common/types'
 import { encodeUriCommon, parseUriCommon } from '../common/uriHelpers'
 import { getDenomInfo } from '../common/utils'
 import {
@@ -72,7 +73,10 @@ export class ZcashTools implements EdgeCurrencyTools {
   }
 
   // will actually use MNEMONIC version of private key
-  async importPrivateKey(userInput: string): Promise<Object> {
+  async importPrivateKey(
+    userInput: string,
+    opts: JsonObject = {}
+  ): Promise<Object> {
     const { pluginId } = this.currencyInfo
     const isValid = validateMnemonic(userInput)
     if (!isValid)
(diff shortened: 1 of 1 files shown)


src/zcash/zecPlugin.ts
```diff
@@ -72,7 +72,10 @@ export class ZcashTools implements EdgeCurrencyTools {
   }
 
   // will actually use MNEMONIC version of private key
-  async importPrivateKey(userInput: string): Promise<Object> {
+  async importPrivateKey(
+    userInput: string,
+    opts?: JsonObject
```
swansontec, line 77: I would do `opts: JsonObject = {}` to avoid the `?? {}` later. Things are easier to read when we get rid of `undefined` as early as possible. It also means that there is a single source of truth for what the default should be.

src/zcash/zecPlugin.ts
```diff
@@ -86,12 +89,31 @@ export class ZcashTools implements EdgeCurrencyTools {
...
-    const birthdayHeight = await this.getNewWalletBirthdayBlockheight()
+    const currentNetworkHeight = await this.getNewWalletBirthdayBlockheight()
+
+    let height = currentNetworkHeight
+
+    const { birthdayHeight } = opts ?? {}
+    if (birthdayHeight != null) {
+      if (
+        typeof birthdayHeight !== 'string' ||
+        !/^[1-9]\d*$/gm.test(birthdayHeight) // must be a valid blockheight
```
swansontec, line 100: Did ChatGPT come up with this regexp? I ask because it matches an array of number strings, one per line. For instance, this would be valid input for the regexp:

```
12
45
23
```

If you delete the `gm` on the end of the regexp, then it will match only a single number like "123". It's such a weirdly specific thing to add, but still buggy, that it makes me think AI was involved.

We already have an `asIntegerString` cleaner that does this validation, so you could use that + `asMaybe` if you wanted.

swansontec (changes requested):

src/zcash/zecPlugin.ts
peachbits: Haha no I didn't use chatgpt. I grabbed it from stackoverflow because I'm still living in 2022. Will make the switch to asIntegerString

swansontec (approved):

PR #565, opened 2023-05-06
Fix zkSync gas limit issue by always using eth_estimateGas

samholmes:
### CHANGELOG

 none

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204535275865679

### DIFF
--- a/src/ethereum/ethEngine.ts
+++ b/src/ethereum/ethEngine.ts
@@ -350,8 +350,7 @@ export class EthereumEngine extends CurrencyEngine<
             // Subscribe to call requests
             connector.on(
               'call_request',
-              // @ts-expect-error
-              (error: Error, payload: EvmWcRpcPayload) => {
+              (error: Error | null, payload: EvmWcRpcPayload) => {
                 try {
                   // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
                   if (error) throw error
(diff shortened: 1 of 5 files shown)


src/ethereum/ethEngine.ts
```diff
@@ -361,26 +361,33 @@ export class EthereumEngine extends CurrencyEngine<
                   switch (payload.method) {
                     case 'eth_sendTransaction':
                     case 'eth_signTransaction': {
-                      const params = asObject({
+                      const txParam = asObject({
+                        from: asString,
+                        to: asOptional(asString),
+                        data: asString,
```
swansontec, line 367: If these 3 aren't used in the function, do we need to clean them? I only ask because the Ethereum engine is used in a lot of places, and I don't want to break if some other coin fails to provide these.

src/ethereum/ethEngine.ts
```diff
@@ -352,7 +352,7 @@ export class EthereumEngine extends CurrencyEngine<
               'call_request',
               // @ts-expect-error
               (error: Error, payload: EvmWcRpcPayload) => {
-                try {
+                const iif = async (): Promise<void> => {
```
swansontec, line 355: The try / catch block should catch all errors, both sync and async. I don't see how the IIF wrapper helps anything. What am I missing?

swansontec (approved):
Approving with some questions. If there are good answers, we can merge.

src/ethereum/ethEngine.ts
samholmes: The function is there to use the async/await syntax. The callback type wont allow the return of a Promise. I can use `then/catch` instead?

src/ethereum/ethEngine.ts
samholmes: I added them because I'm passing txParam to multicastServers where they're expected for the gas estimation. I can choose not to clean them if that seems less risky?

swansontec: Ah, that's where they are being used! Very good.

PR #572, opened 2023-05-24
FIO: Handle empty otherParams objects as null

peachbits:
---
  - https://app.asana.com/0/0/1204663161710905

### DIFF
--- a/src/fio/FioEngine.ts
+++ b/src/fio/FioEngine.ts
@@ -1331,7 +1331,8 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
       throw new Error('makeSpend Missing publicAddress')
     if (quantity == null) throw new NoAmountSpecifiedError()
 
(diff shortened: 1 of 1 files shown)


src/fio/FioEngine.ts
```diff
@@ -1331,7 +1331,11 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
       throw new Error('makeSpend Missing publicAddress')
     if (quantity == null) throw new NoAmountSpecifiedError()
 
-    const {
+    let { otherParams } = edgeSpendInfo
+    if (
+      otherParams == null ||
+      (otherParams != null && Object.keys(otherParams).length === 0)
```
paullinator, line 1337: Optional fix.  Check for != null is not necessary as it’s assumed if the first condition of the || is false.

paullinator (approved):
Approved with optional fix

PR #574, opened 2023-05-29
Fix Token Accelerate Tx (quick hack fix)

samholmes:
### CHANGELOG

- Changed: Re-enable token transaction accelerate for EVM currencies

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204597702128318

### DIFF
--- a/src/ethereum/EthereumEngine.ts
+++ b/src/ethereum/EthereumEngine.ts
@@ -1299,11 +1299,13 @@ export class EthereumEngine extends CurrencyEngine<
       replacedTx.otherParams
     )
 
-    // Transaction checks
-    if (replacedTx == null || replacedTx.blockHeight > 0) {
+    // Transaction checks:
+    // The transaction must be found and not confirmed or dropped.
+    if (replacedTx == null || replacedTx.blockHeight !== 0) {
       return null
     }
-    // Other params checks
+    // Other params checks:
+    // The transaction must have a known nonce used and no data payload.
     if (
       replacedTxOtherParams == null ||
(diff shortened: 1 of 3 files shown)


src/ethereum/EthereumEngine.ts
```diff
@@ -857,7 +857,8 @@ export class EthereumEngine extends CurrencyEngine<
         gasPrice: gasPrice,
         gasUsed: '0',
         nonceUsed,
-        data
+        data,
+        isFromMakeSpend: true
```
paullinator, line 861: This should be `isFromMakeSpend: !hasUserMemo`

src/ethereum/EthereumEngine.ts
```diff
@@ -1302,11 +1304,13 @@ export class EthereumEngine extends CurrencyEngine<
...
     }
     // Other params checks:
-    // The transaction must have a known nonce used and no data payload.
     if (
       replacedTxOtherParams == null ||
+      // The transaction must have a known nonce used.
       replacedTxOtherParams.nonceUsed == null ||
-      replacedTxOtherParams.data != null
+      // Must not have data payload unless it's normalized.
+      (replacedTxOtherParams.data != null &&
```
paullinator, line 1312: Should just be `!replacedTxOtherParams.isFromMakeSpend`. It's dangerous to check the `.data` as it might be null if we got it off the network even if it's a contract tx

paullinator (changes requested):

paullinator (approved):

PR #576, opened 2023-05-31
Fix memo handling before serialization

peachbits:


src/tron/TronEngine.ts
```diff
@@ -1018,6 +1018,8 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
       .slice(4, 8)
     const refBlockHash = this.recentBlock.hash.slice(16, 32)
     const expiration = this.recentBlock.timestamp + 60 * 5 * 1000 // five minutes
+    const data =
+      note != null ? base16.stringify(Buffer.from(note, 'ascii')) : undefined
```
swansontec, line 1022: Optional: In the GUI, we always put the `null` case first, so like `note == null ? undefined : base16.stringify(...)`. This way is easier to read, since all the null-handling stuff is gathered together at the beginning.

swansontec (approved):

PR #580, opened 2023-06-16
Fix broken max-spend for zkSync

samholmes:
### CHANGELOG

- Fixed: Fix broken max-spend for zkSync

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204825785199825
  - https://app.asana.com/0/0/1204829809536930

### DIFF
--- a/src/ethereum/EthereumEngine.ts
+++ b/src/ethereum/EthereumEngine.ts
@@ -807,7 +807,6 @@ export class EthereumEngine extends CurrencyEngine<
       this.currencyInfo,
       this.networkInfo
     )
-    let gasLimit: string = miningFees.gasLimit
 
     //
     // Nonce:
@@ -832,7 +831,7 @@ export class EthereumEngine extends CurrencyEngine<
       otherParams = {
         from: [this.walletLocalData.publicKey],
         to: [publicAddress],
-        gas: gasLimit,
+        gas: miningFees.gasLimit,
         gasPrice: miningFees.gasPrice,
         gasUsed: '0',
         nonceUsed,
@@ -870,7 +869,7 @@ export class EthereumEngine extends CurrencyEngine<
       otherParams = {
         from: [this.walletLocalData.publicKey],
         to: [contractAddress],
-        gas: gasLimit,
+        gas: miningFees.gasLimit,
         gasPrice: miningFees.gasPrice,
         gasUsed: '0',
         tokenRecipientAddress: publicAddress,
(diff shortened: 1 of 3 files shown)


src/ethereum/ethereumTypes.ts
```diff
@@ -260,7 +260,7 @@ export interface EthereumTxOtherParams {
 export const asEthereumTxOtherParams = asObject<EthereumTxOtherParams>({
   from: asArray(asString),
   to: asArray(asString),
-  gas: asString,
+  gasLimit: asString,
```
peachbits, line 269: You'll need to update txRpcParamsToSpendInfo since the spec is to use 'gas'.

src/ethereum/EthereumEngine.ts
```diff
@@ -892,6 +893,66 @@ export class EthereumEngine extends CurrencyEngine<
...
+  getTxParameterInformation(
+    edgeSpendInfo: EdgeSpendInfo,
+    currencyCode: string,
+    currencyInfo: EdgeCurrencyInfo
+  ): EthereumTxParameterInformation {
+    const { spendTargets } = edgeSpendInfo
+    const spendTarget = spendTargets[0]
+    const { publicAddress, nativeAmount } = spendTarget
+
+    if (nativeAmount == null) throw new NoAmountSpecifiedError()
```
peachbits, line 929: This check is duplicated

peachbits (changes requested):

peachbits:
I think most of the renames were unnecessary but I won't ask you to change them. Just the one you renamed twice.

src/ethereum/EthereumEngine.ts
samholmes: It's an assertion to satisfy the type checker.

src/ethereum/EthereumEngine.ts
```diff
@@ -121,8 +121,7 @@ export class EthereumEngine extends CurrencyEngine<
     this.networkInfo = env.networkInfo
     this.ethNetwork = new EthereumNetwork(this)
     this.lastEstimatedGasLimit = {
-      publicAddress: '',
-      contractAddress: '',
+      toAddress: '',
```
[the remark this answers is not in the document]
samholmes: I'll revert this change. We want the publicAddress to trigger a recalculation/call to eth_estimateGas because it may be the case that the destination of the token change may impact the gas limit (the contract may do for different accounts).

src/ethereum/ethereumTypes.ts
samholmes: I'll revert to avoid potential regressions here. Given the level of how many ts-expect-errors may be present in this repo, this change is too risky.

peachbits (approved):

PR #581, opened 2023-06-20
Deprecate wallet connect v1 and add v2

peachbits:
Wallet Connect is no longer handled in these plugins. Instead, a utility function is provided in order to parse a smart contract payload to identify the amounts and token. Otherwise the host app is responsible for managing the connection. 

---
  - https://app.asana.com/0/0/1204857257229134

### DIFF
--- a/src/ethereum/ethereumTypes.ts
+++ b/src/ethereum/ethereumTypes.ts
@@ -15,13 +15,7 @@ import {
 } from 'cleaners'
 import { EdgeSpendInfo } from 'edge-core-js/types'
 
-import {
-  asSafeCommonWalletInfo,
-  Dapp,
-  WalletConnectPayload,
-  WcDappDetails,
-  WcProps
-} from '../common/types'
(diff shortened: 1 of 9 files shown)


src/ethereum/ethereumTypes.ts
```diff
@@ -18,6 +18,7 @@ import { EdgeSpendInfo } from 'edge-core-js/types'
 import {
   asSafeCommonWalletInfo,
   Dapp,
+  WalletConnectPayload,
```
swansontec: Import is unused. How did this commit pass the linter?

swansontec (changes requested):
The first commit has a dangling import, but otherwise this seems fine.

src/ethereum/ethereumTypes.ts
peachbits: That shouldn't have even been a commit. I combined the first two

swansontec (approved):

PR #590, opened 2023-07-18
Remove eosmadrid

peachbits:
---
  - https://app.asana.com/0/0/1205057984776312


swansontec (approved):
I am concerned about the second commit. What happens if a signature is invalid? Does the app take forever to error out now? Assuming there's a good answer, we can merge.

swansontec (approved):

PR #595, opened 2023-07-28
Only use `io.fetchCors`, never `io.fetch`

swansontec:
With the exception of when `io.fetchCors` doesn't exist. Then we have no choice.

### CHANGELOG

- changed: Use `EdgeIo.fetchCors` for all requests.

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205163504777914

### DIFF
--- a/src/common/CurrencyEngine.ts
+++ b/src/common/CurrencyEngine.ts
@@ -84,7 +84,6 @@ export class CurrencyEngine<
   currentSettings: any
   timers: any
   walletId: string
-  io: EdgeIo
   log: EdgeLog
   warn: (message: string, e?: Error) => void
   error: (message: string, e?: Error) => void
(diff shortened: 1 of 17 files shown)


src/common/CurrencyEngine.ts
```diff
@@ -84,7 +84,6 @@ export class CurrencyEngine<
   currentSettings: any
   timers: any
   walletId: string
-  io: EdgeIo
```
samholmes, line 87: How can this be removed? A separate commit for this one would have made it clearer perhaps?

samholmes (approved):
I like how cleaned up everything became. It's pretty straight-forward, I only had one question that wasn't apparent from the diffs.

src/common/CurrencyEngine.ts
swansontec: It is no longer used. The only property anybody ever accessed was `fetch`, and we don't do that anymore.

PR #598, opened 2023-08-09
Paul/xrp dex

paullinator:
### CHANGELOG

Add Ripple makeTx routine for xrp dex OfferCreate

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205228472265662

### DIFF
--- a/src/ripple/RippleEngine.ts
+++ b/src/ripple/RippleEngine.ts
@@ -517,7 +517,7 @@ export class XrpEngine extends CurrencyEngine<
         edgeToken.networkLocation
       )
       const networkFee = SET_TRUST_LINE_FEE
-      const trustSetTx: TrustSet = await this.tools.rippleApi.autofill({
+      const xrpTransaction: TrustSet = await this.tools.rippleApi.autofill({
         TransactionType: 'TrustSet',
         Account: this.walletLocalData.publicKey,
         Fee: networkFee,
@@ -542,7 +542,7 @@ export class XrpEngine extends CurrencyEngine<
         ourReceiveAddresses: [],
         signedTx: '',
         otherParams: {
-          trustSetTx
+          xrpTransaction
         },
         walletId: this.walletId
       }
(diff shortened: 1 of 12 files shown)


src/ripple/rippleTypes.ts
```diff
@@ -48,3 +49,35 @@ export type RipplePrivateKeys = ReturnType<typeof asRipplePrivateKeys>
 export const asRipplePrivateKeys = asObject({
   rippleKey: asString
 })
+
+export interface RippleNetworkLocation {
+  currency: string
+  issuer: string
+}
```
swansontec, line 56: You already have `asXrpNetworkLocation` a few lines above. You can do `type RippleNetworkLocation = ReturnType<typeof asXrpNetworkLocation>`

src/ripple/RippleEngine.ts
```diff
@@ -119,6 +138,92 @@ export class XrpEngine extends CurrencyEngine<
...
+    }
+  }
+
+  nativeToXrpAmount(nativeAmount: string, tokenId?: string): Amount {
+    if (tokenId == null) {
+      // We are sending XRP
+      return nativeAmount
+    } else {
+      const { networkLocation, denominations } = this.allTokensMap[tokenId]
+      const { currency, issuer } = networkLocation as RippleNetworkLocation
```
swansontec, line 220: Please use `asXrpNetworkLocation` instead of a type-cast.

src/ripple/RippleEngine.ts
```diff
@@ -119,6 +138,92 @@ export class XrpEngine extends CurrencyEngine<
...
+
+          const TakerGets = this.nativeToXrpAmount(
+            fromNativeAmount,
+            fromTokenId
+          )
+          const TakerPays = this.nativeToXrpAmount(toNativeAmount, toTokenId)
+
+          const Expiration = unixTimeToRippleTime(
+            (expiration ?? DEX_ORDER_EXPIRATION_TIME_DEFAULT_S) * 1000
+          )
```
swansontec, line 161: Optional: I see that you are using uppercase, because that is what the `OfferCreate` JSON expects. However, we generally reserve uppercase names for types. Would it make more sense to use lowercase for the variables, and then do `{ ... Expiration: exipration, ... }` and such to change case inside the object?

swansontec (changes requested):
One easy change and one optional change.

src/ripple/rippleTypes.ts
paullinator: Ah good catch. how'd i miss that.

src/ripple/RippleEngine.ts
paullinator: it crossed my mind but went for the convenient route. will do

swansontec (approved):

PR #601, opened 2023-08-11
Upgrade react-native-zcash

peachbits:
---
  - https://app.asana.com/0/0/1205252649096314


src/zcash/ZcashEngine.ts
```diff
@@ -42,13 +42,16 @@ export class ZcashEngine extends CurrencyEngine<
...
   initialNumBlocksToDownload!: number
   initializer!: ZcashInitializerConfig
-  alias!: string
   progressRatio!: number
   queryMutex: boolean
   makeSynchronizer: (
     config: ZcashInitializerConfig
   ) => Promise<ZcashSynchronizer>
 
+  // Synchronizer management
```
swansontec, line 52: Please move `synchronizer` to this section, and make it `?:` instead of `!:`. This will force you to check for `null` before accessing the synchronizer, since we have even less guarantee than before that the synchronizer will exist.

src/zcash/ZcashEngine.ts
```diff
@@ -42,13 +42,16 @@ export class ZcashEngine extends CurrencyEngine<
...
+  started: boolean
+  resolver?: (value: number | PromiseLike<number>) => void
```
swansontec, line 54: Please call this `stopSyncing` or something instead of `resolver`.

src/zcash/ZcashTools.ts
```diff
@@ -33,9 +30,7 @@ export class ZcashTools implements EdgeCurrencyTools {
   currencyInfo: EdgeCurrencyInfo
   io: EdgeIo
   networkInfo: ZcashNetworkInfo
-
-  KeyTool: typeof KeyToolType
-  AddressTool: typeof AddressToolType
+  Tools: typeof ToolsType
```
swansontec, line 33: Can we call this `nativeTools` or such? Capitalized names are always classes or types, and lowercase `tools` usually means `EdgeCurrencyTools` in this repo.

swansontec (commented):
Now that

swansontec (approved):

swansontec (approved):

swansontec (approved):
One small change needed.

PR #610, opened 2023-08-29
Paul/add pokt

paullinator:
### CHANGELOG

added: Pokt network RPC servers

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205228135402245

### DIFF
--- a/src/ethereum/EthereumNetwork.ts
+++ b/src/ethereum/EthereumNetwork.ts
@@ -31,6 +31,7 @@ import {
   asBlockChairAddress,
   asCheckBlockHeightBlockchair,
   asCheckTokenBalBlockchair,
+  asEthereumInitKeys,
   asEvmScancanTokenTransaction,
   asEvmScanInternalTransaction,
   asEvmScanTransaction,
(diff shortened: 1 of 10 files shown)


src/ethereum/EthereumNetwork.ts
```diff
@@ -506,30 +507,18 @@ export class EthereumNetwork {
...
+    const regex = /{{(.*?)}}/g
+    const match = regex.exec(url)
+    if (match != null) {
+      const key = match[1]
+      const cleanKey = asEthereumInitKeys(key)
+      const apiKey = this.ethEngine.initOptions[cleanKey]
+      if (typeof apiKey === 'string') {
+        url = url.replace(match[0], apiKey)
+      } else {
+        throw new Error('Incorrect apikey type for RPC')
```
samholmes, line 519: If the `apiKey` is one of the values in `initOptions` of `cleanKey`, and if `cleanKey` is a key in the `initOptions`, then the error message should state something like:

```js
`Missing ${apiKey} in 'initOptions' for ${pluginId}`
```

samholmes (changes requested):

src/ethereum/EthereumNetwork.ts
paullinator: apiKey can also be an array, that's what this clause was meant to catch. But i'll add another for the case of `null`

src/ethereum/EthereumNetwork.ts
```diff
@@ -515,6 +515,10 @@ export class EthereumNetwork {
       const apiKey = this.ethEngine.initOptions[cleanKey]
       if (typeof apiKey === 'string') {
         url = url.replace(match[0], apiKey)
+      } else if (apiKey == null) {
+        throw new Error(
+          `Missing ${cleanKey} in 'initOptions' for ${this.ethEngine.currencyInfo.pluginId}`
+        )
       } else {
         throw new Error('Incorrect apikey type for RPC')
```
samholmes, line 523: I looked at the type for `apiKey` and it looks like it could be an array of strings. Shouldn't we handle this case for `evmScanApiKey`? Maybe a simple random pick of one the keys

samholmes (changes requested):
One more thing

src/ethereum/EthereumNetwork.ts
paullinator: No. Evmscan servers are not an RPC servers but a completely different "type". This PR is only for RPC servers and I don't want to scope creep the PR. That would also require another hour of testing.

samholmes: Ah, didn’t catch that difference. Thanks for correcting me here

samholmes (approved):

PR #616, opened 2023-09-05
Disable Piratechain

peachbits:
---
  - https://app.asana.com/0/0/1205419516159280


swansontec (approved):
I would have just deleted the code, instead of commenting it, since we can always use `git revert` to get it back when we are ready. Either way, though, it seems like it would work.

PR #618, opened 2023-09-07
Fix filecoin spend saving

samholmes:
The date field was missing from the signed transactions so it wasn't being saved (I suppose that's a requirement by the CurrencyEngine implementation). Also, `isSend` had a bug that was copied from `ZcashEngine`; now it's fixed in both places.

### CHANGELOG

- fixed: Bug prevent Filecoin spend trasactions from being saved in the wallet (by `saveTx`)

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205428345589472

### DIFF
--- a/src/filecoin/FilecoinEngine.ts
+++ b/src/filecoin/FilecoinEngine.ts
@@ -189,15 +189,15 @@ export class FilecoinEngine extends CurrencyEngine<
     }
 
     const networkFee = mul(txJson.GasLimit.toString(), txJson.GasPremium) // TODO: Include base fee and burn fee somehow?
-    const totalTxAmount = add(nativeAmount, networkFee)
+    const txNativeAmount = mul(add(nativeAmount, networkFee), '-1')
 
     const edgeTransaction: EdgeTransaction = {
       txid: '',
       date: 0,
       currencyCode,
       blockHeight: 0,
-      nativeAmount: `-${totalTxAmount}`,
-      isSend: nativeAmount.startsWith('-'),
+      nativeAmount: txNativeAmount,
+      isSend: true,
       networkFee,
       ourReceiveAddresses: [],
       otherParams,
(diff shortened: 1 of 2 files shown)


src/filecoin/FilecoinEngine.ts
```diff
@@ -189,15 +189,15 @@ export class FilecoinEngine extends CurrencyEngine<
...
-    const totalTxAmount = add(nativeAmount, networkFee)
+    const txNativeAmount = mul(add(nativeAmount, networkFee), '-1')
 
     const edgeTransaction: EdgeTransaction = {
       txid: '',
       date: 0,
       currencyCode,
       blockHeight: 0,
-      nativeAmount: `-${totalTxAmount}`,
-      isSend: nativeAmount.startsWith('-'),
```
paullinator, line 200: should just be `true`

src/zcash/ZcashEngine.ts
```diff
@@ -338,13 +338,15 @@ export class ZcashEngine extends CurrencyEngine<
...
+
     const edgeTransaction: EdgeTransaction = {
       txid: '', // txid
       date: 0, // date
       currencyCode, // currencyCode
       blockHeight: 0, // blockHeight
-      nativeAmount: `-${totalTxAmount}`, // nativeAmount
-      isSend: nativeAmount.startsWith('-'),
+      nativeAmount: txNativeAmount, // nativeAmount
+      isSend: txNativeAmount.startsWith('-'),
```
paullinator, line 349: Should just be `true`

paullinator (changes requested):

src/filecoin/FilecoinEngine.ts
samholmes: Why do other engines detect sends based on the sign of the number?

paullinator (approved):

PR #619, opened 2023-09-07
Filecoin network query fixes

samholmes:
Heads up. This disables Filscan because I haven't found a convenient way to calculate fees from Filscan's API. I haven't quit wrapped my head around [how Filecoin fees are calculated](https://spec.filecoin.io/systems/filecoin_vm/gas_fee/), so rather than calculating the fee from the message fields, I use a hack to peer into the "transfer" data returned by Filfox.

### CHANGELOG

- fixed: Filecoin network fee query issue
- changed: Use Filfox exclusively for Filecoin transaction querying

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205419647873868

### DIFF
--- a/src/filecoin/FilecoinEngine.ts
+++ b/src/filecoin/FilecoinEngine.ts
@@ -87,9 +87,7 @@ export class FilecoinEngine extends CurrencyEngine<
   }
 
   initData(): void {
-    // Initialize walletLocalData:
-    // ...
-
+    this.tokenCheckTransactionsStatus[this.currencyInfo.currencyCode] = 0
     // Engine variables
     this.availableAttoFil = '0'
   }
--- a/src/filecoin/FilecoinEngine.ts
+++ b/src/filecoin/FilecoinEngine.ts
@@ -48,6 +48,7 @@ export class FilecoinEngine extends CurrencyEngine<
 > {
   address: Address
   availableAttoFil: string
+  isScanning: boolean
   networkInfo: FilecoinNetworkInfo
   otherData!: FilecoinWalletOtherData
   pluginId: string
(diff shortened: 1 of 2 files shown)


src/filecoin/FilecoinEngine.ts
```diff
@@ -331,6 +331,8 @@ export class FilecoinEngine extends CurrencyEngine<
     }): void => {
       this.addTransaction(this.currencyInfo.currencyCode, tx)
       this.onUpdateTransactions()
+      this.otherData.knownTxids[tx.txid] = true
```
paullinator, line 334: There is already an `engine.txIdMap` you should use. It's automatically kept up  to date by `addTransaction`

src/filecoin/FilecoinEngine.ts
```diff
@@ -382,13 +384,10 @@ export class FilecoinEngine extends CurrencyEngine<
...
-        const idx = this.findTransaction(this.currencyInfo.currencyCode, txid)
 
-        if (idx >= 0) {
-          // Exit early because we reached transaction history from previous
-          // check
-          return
-        }
+        // Exit early because we reached transaction history from previous
+        // check
+        if (this.otherData.knownTxids[txid]) return
```
paullinator, line 390: Use

```
if (this.findTransaction('FIL', normalizeAddress(txid)) !== -1) return
```

paullinator (changes requested):

src/filecoin/FilecoinEngine.ts
samholmes: I cannot use this because it's not available after loading the engine.

src/filecoin/FilecoinEngine.ts
samholmes: Is it available after loading the engine? If so, I'll use that instead of the `otherData` which does get loaded.

paullinator (approved):

PR #622, opened 2023-09-11
Add WithdrawExpireUnfreeze support

peachbits:
---
  - https://app.asana.com/0/0/1205419517607650

### DIFF
--- a/src/tron/TronEngine.ts
+++ b/src/tron/TronEngine.ts
@@ -54,6 +54,7 @@ import {
   asTRXTransferContract,
   asUnfreezeBalanceContract,
   asUnfreezeV2BalanceContract,
+  asWithdrawExpireUnfreezeContract,
   CalcTxFeeOpts,
   ReferenceBlock,
   SafeTronWalletInfo,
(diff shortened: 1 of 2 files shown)


src/tron/TronEngine.ts
```diff
@@ -1242,6 +1244,55 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
...
+      '0'
+    )
+
+    const edgeTransaction: EdgeTransaction = {
+      txid: '',
+      date: 0,
+      currencyCode: this.currencyInfo.currencyCode,
+      blockHeight: 0,
+      nativeAmount: sub(claimedAmount, networkFee),
+      isSend: false,
```
swansontec, line 1327: This should be true, since it's a transaction we ourselves have constructed. `false` is for stuff we detect off the network.

src/tron/TronEngine.ts
```diff
@@ -705,6 +706,50 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
...
+        if (sentTx == null) break
+        const nativeAmount = sentTx.nativeAmount
+
+        const edgeTransaction: EdgeTransaction = {
+          txid,
+          date: Math.floor(timestamp / 1000),
+          currencyCode,
+          blockHeight: blockNumber,
+          nativeAmount,
+          isSend: false,
```
swansontec, line 743: Also `true` here. Presumably we ourselves did the unstaking.

swansontec (changes requested):
One small change. I could be wrong, so if you disagree, just leave a comment and I'll take another look.

src/tron/TronEngine.ts
peachbits: I marked it false since the net effect is positive to the wallet (staked amount returning to spendable balance minus fee). There's no other incoming transaction to show the funds have been added back to the wallet

swansontec: I see. That means the bug is in the GUI, since the "isSend" concept should be independent of whether money was added or removed from the balance. This is getting into refactor territory, so let's merge as-is.

swansontec (approved):

PR #624, opened 2023-09-13
Fix Zcash memo handling

peachbits:
---
  - https://app.asana.com/0/0/1205041005078327

### DIFF
--- a/src/zcash/ZcashEngine.ts
+++ b/src/zcash/ZcashEngine.ts
@@ -261,20 +261,20 @@ export class ZcashEngine extends CurrencyEngine<
       ourReceiveAddresses.push(this.walletInfo.keys.publicKey)
     }
 
-    const memos: EdgeMemo[] = []
-    if (tx.memo != null) {
-      memos.push({
+    const edgeMemos: EdgeMemo[] = tx.memos
+      .filter(text => text !== '')
+      .map(text => ({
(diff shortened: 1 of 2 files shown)


peachbits:
Draft until https://github.com/EdgeApp/react-native-zcash/pull/38 is published

src/zcash/ZcashEngine.ts
```diff
@@ -261,20 +261,19 @@ export class ZcashEngine extends CurrencyEngine<
...
     }
 
-    const memos: EdgeMemo[] = []
-    if (tx.memo != null) {
-      memos.push({
+    const edgeMemos: EdgeMemo[] = tx.memos
+      .filter(text => text !== '')
+      .map(text => ({
         type: 'text',
-        value: tx.memo
```
swansontec, line 268: The default is `memoName: 'memo'`, which is what we want to show in the GUI. However, we could add this field to be explicit about what this thing is called.

swansontec (approved):

src/zcash/ZcashEngine.ts
peachbits: Zcash calls them memo so the default is fine

peachbits: changed it anyway

PR #629, opened 2023-09-20
Use sapling address from deriveUnifiedAddresses

peachbits:
---
  - https://app.asana.com/0/0/1205534330730646

### DIFF
--- a/src/zcash/zcashTypes.ts
+++ b/src/zcash/zcashTypes.ts
@@ -64,6 +64,12 @@ export interface ZcashInitializerConfig {
   birthdayHeight: number
 }
 
+export interface ZcashAddresses {
+  unifiedAddress: string
+  saplingAddress: string
+  transparentAddress: string
+}
+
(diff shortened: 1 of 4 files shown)


peachbits:
In draft until https://github.com/EdgeApp/react-native-zcash/pull/40 is merged and published

src/zcash/zcashTypes.ts
```diff
@@ -64,6 +64,12 @@ export interface ZcashInitializerConfig {
   birthdayHeight: number
 }
 
+export interface ZcashAddresses {
+  unifiedAddress: string
+  saplingAddress: string
+  transparentAddress: string
+}
+
```
swansontec, line 72: Can you do `import type {ZcashAddresses } from 'react-native-zcash'`? The `import type` syntax ensures that there are no runtime consequences for the import, only type information.

swansontec (approved):

PR #637, opened 2023-10-05
XLM Mnemonic Phrase Support

Jon-edge:
### CHANGELOG

- added: Support for 12/24-word mnemonic XLM wallet imports

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204800868907796

### DIFF
--- a/src/stellar/StellarTools.ts
+++ b/src/stellar/StellarTools.ts
@@ -1,4 +1,6 @@
 import { add, div } from 'biggystring'
+import { mnemonicToSeed, validateMnemonic } from 'bip39'
+import * as ed25519 from 'ed25519-hd-key'
 import {
   EdgeCurrencyInfo,
   EdgeCurrencyTools,
@@ -19,7 +21,8 @@ import { getLegacyDenomination } from '../common/utils'
 import {
   asSafeStellarWalletInfo,
   asStellarPrivateKeys,
-  StellarNetworkInfo
+  StellarNetworkInfo,
+  StellarPrivateKeys
 } from './stellarTypes'
 
 const URI_PREFIX = 'web+stellar'
@@ -81,11 +84,33 @@ export class StellarTools implements EdgeCurrencyTools {
     return { stellarKey: keypair.secret() }
   }
 
-  async importPrivateKey(privateKey: string): Promise<{ stellarKey: string }> {
-    privateKey.replace(/ /g, '')
-    stellarApi.Keypair.fromSecret(privateKey)
-    if (privateKey.length !== 56) throw new Error('Private key wrong length')
-    return await Promise.resolve({ stellarKey: privateKey })
+  async importPrivateKey(userInput: string): Promise<StellarPrivateKeys> {
+    let stellarKey
+    let stellarMnemonic
+
+    if (validateMnemonic(userInput)) {
+      const seed = await mnemonicToSeed(userInput)
+      const derivedSeed = ed25519.derivePath(
+        "m/44'/148'/0'",
+        seed.toString('hex')
+      ).key
+      const keypair = stellarApi.Keypair.fromRawEd25519Seed(
+        Uint8Array.from(derivedSeed) as unknown as number[]
+      )
+
+      stellarKey = keypair.secret()
+      stellarMnemonic = userInput
+    } else {
+      userInput.replace(/ /g, '')
+      stellarApi.Keypair.fromSecret(userInput)
+      if (userInput.length !== 56) throw new Error('Private key wrong length')
+      stellarKey = userInput
+    }
+
+    return await Promise.resolve({
+      stellarKey,
+      stellarMnemonic
+    })
   }
 
   async derivePublicKey(walletInfo: EdgeWalletInfo): Promise<Object> {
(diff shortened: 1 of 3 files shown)


src/stellar/StellarTools.ts
```diff
@@ -28,16 +32,18 @@ export class StellarTools implements EdgeCurrencyTools {
...
   highestTxHeight: number = 0
   stellarApiServers: StellarServer[]
 
   constructor(env: PluginEnvironment<StellarNetworkInfo>) {
-    const { builtinTokens, currencyInfo, io, networkInfo } = env
+    const { builtinTokens, currencyInfo, io, log, networkInfo } = env
     this.builtinTokens = builtinTokens
     this.currencyInfo = currencyInfo
     this.io = io
+    this.log = log
```
peachbits, line 46: Looks like a testing remnant. Not unwelcome but it's unused and doesn't fit with the commit

src/stellar/StellarTools.ts
```diff
@@ -81,11 +87,33 @@ export class StellarTools implements EdgeCurrencyTools {
     return { stellarKey: keypair.secret() }
   }
 
-  async importPrivateKey(privateKey: string): Promise<{ stellarKey: string }> {
-    privateKey.replace(/ /g, '')
-    stellarApi.Keypair.fromSecret(privateKey)
-    if (privateKey.length !== 56) throw new Error('Private key wrong length')
-    return await Promise.resolve({ stellarKey: privateKey })
+  async importPrivateKey(userInput: string): Promise<StellarPrivateKeys> {
+    if (/^(0x)?[0-9a-fA-F]{64}$/.test(userInput)) {
```
peachbits, line 91: The current private key format is an all-caps alphanumeric string and this checks for hex. We can support hex if that's a thing but we have to maintain support for the keys we create now.

src/stellar/StellarTools.ts
```diff
@@ -81,11 +87,33 @@ export class StellarTools implements EdgeCurrencyTools {
...
+        stellarMnemonic: ''
+      })
+    } else {
+      // it looks like a mnemonic, so validate that way:
+      if (!validateMnemonic(userInput)) {
+        // "input" instead of "mnemonic" in case private key
+        // was just the wrong length
+        throw new Error('Invalid input')
+      }
+      const derivedSeed = ed25519.derivePath("m/44'/314159'/0'", userInput).key
```
peachbits, line 108: Where did you get this path? The XLM SLIP44 path is 148.

src/stellar/StellarTools.ts
```diff
@@ -81,11 +87,33 @@ export class StellarTools implements EdgeCurrencyTools {
...
+  async importPrivateKey(userInput: string): Promise<StellarPrivateKeys> {
+    if (/^(0x)?[0-9a-fA-F]{64}$/.test(userInput)) {
+      // It looks like a private key, so validate the hex:
+      userInput.replace(/ /g, '')
+      stellarApi.Keypair.fromSecret(userInput)
+      if (userInput.length !== 56) throw new Error('Private key wrong length')
+
+      return await Promise.resolve({
+        stellarKey: userInput,
+        stellarMnemonic: ''
```
peachbits, line 99: Don't save this as an empty string. The cleaner says it can be undefined so don't save it at all

peachbits (changes requested):

src/stellar/StellarTools.ts
Jon-edge: https://github.com/limitpointinf0/pi-wallet-cli/blob/e9c016e0d686aa04fb9ddf2411d49b52d4b82511/src/createAsset.js#L44

src/stellar/StellarTools.ts
```diff
@@ -81,11 +84,28 @@ export class StellarTools implements EdgeCurrencyTools {
...
+  async importPrivateKey(userInput: string): Promise<StellarPrivateKeys> {
+    let stellarKey
+    let stellarMnemonic
+
+    if (validateMnemonic(userInput)) {
+      const derivedSeed = ed25519.derivePath("m/44'/314159'/0'", userInput).key
+      const keypair = stellarApi.Keypair.fromRawEd25519Seed(
+        Uint8Array.from(derivedSeed) as unknown as number[]
+      )
+      stellarKey = keypair.secret().replace(/\s/g, '')
```
peachbits, line 96: Why are you editing the key returned from secret()? We use this to instantiate a KeyPair later so it needs to be the exact value

src/stellar/StellarTools.ts
peachbits: And the code doesn't work. `derivePath` doesn't accept a mnemonic directly it takes the hex encoded version. You need to use bip39's `mnemonicToSeed` like pi-wallet-cli and SolanaTools does

peachbits (changes requested):

src/stellar/StellarTools.ts
Jon-edge: I'm confused, pi-wallet-cli also uses the `m/44'/314159'/0'`` path when deriving
from mnemonic:
https://github.com/limitpointinf0/pi-wallet-cli/blob/e9c016e0d686aa04fb9ddf2411d49b52d4b82511/src/piLib.js#L22

peachbits (approved):
Approved but please update changelog

PR #639, opened 2023-10-06
Upgrade react-native-zcash

peachbits:
Upgrade to latest rnzcash and implement spend-before-sync support. Remove transaction and balance polling in favoring of listening to events.

---
  - https://app.asana.com/0/0/1205664938290920

### DIFF
--- a/src/zcash/ZcashEngine.ts
+++ b/src/zcash/ZcashEngine.ts
@@ -40,10 +40,8 @@ export class ZcashEngine extends CurrencyEngine<
   otherData!: ZcashWalletOtherData
   synchronizerStatus!: ZcashSynchronizerStatus
   availableZatoshi!: string
-  initialNumBlocksToDownload!: number
   initializer!: ZcashInitializerConfig
   progressRatio!: number
-  queryMutex: boolean
   makeSynchronizer: (
     config: ZcashInitializerConfig
   ) => Promise<ZcashSynchronizer>
@@ -65,7 +63,6 @@ export class ZcashEngine extends CurrencyEngine<
     this.pluginId = this.currencyInfo.pluginId
     this.networkInfo = networkInfo
     this.makeSynchronizer = makeSynchronizer
-    this.queryMutex = false
 
     this.started = false
   }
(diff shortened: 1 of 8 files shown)


peachbits:
Blocked by https://github.com/EdgeApp/react-native-zcash/pull/41

src/zcash/ZcashEngine.ts
```diff
@@ -72,15 +73,9 @@ export class ZcashEngine extends CurrencyEngine<
...
-    const { birthdayHeight } = this.initializer
-
     // walletLocalData
-    if (this.otherData.blockRange.first === 0) {
-      this.otherData.blockRange = {
-        first: birthdayHeight,
-        last: birthdayHeight
-      }
-    }
+    this.otherData.isSdkInitializedOnDisk = true
```
swansontec, line 81: Shouldn't this default to false, to match the cleaner?

src/zcash/ZcashEngine.ts
```diff
@@ -42,7 +42,11 @@ export class ZcashEngine extends CurrencyEngine<
   synchronizerStatus!: ZcashSynchronizerStatus
   availableZatoshi!: string
   initializer!: ZcashInitializerConfig
-  progressRatio!: number
+  progressRatio!: {
+    percent: number
+    lastUpdate: number
```
swansontec, line 47: Should we store this as an ISO date string?

swansontec (approved):
A few small questions, but this looks pretty good overall.

src/zcash/ZcashEngine.ts
peachbits: The purpose of this field is to track if we've ever initialized a synchronizer on this device before. The flag should be false the very first time the synchronizer is started but true for every other synchronizer start (including across resyncs where the zcash dbs are still present on disk). 

So it defaults to false the first time the synchronizer starts but will always be true after that.

src/zcash/ZcashEngine.ts
peachbits: I think a number is appropriate since we're just seeing if a second has passed to prevent to many callbacks

swansontec (approved):

PR #641, opened 2023-10-11
Implement XRP DEX order parsing -> EdgeTxAction's

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-core-js/pull/567

### Description

Parse XRP DEX orders into EdgeTxActions. Supports OfferCreate and OfferCancel type order tx's

---
  - https://app.asana.com/0/0/1205412691849558

### DIFF
--- a/src/ripple/RippleEngine.ts
+++ b/src/ripple/RippleEngine.ts
@@ -15,6 +15,7 @@ import {
   EdgeActivationApproveOptions,
   EdgeActivationQuote,
   EdgeActivationResult,
+  EdgeAssetAmount,
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
   EdgeEngineActivationOptions,
@@ -22,6 +23,8 @@ import {
   EdgeGetActivationAssetsResults,
   EdgeSpendInfo,
   EdgeTransaction,
+  EdgeTxActionSwap,
+  EdgeTxActionSwapType,
   EdgeWalletInfo,
   InsufficientFundsError,
   JsonObject,
@@ -29,7 +32,11 @@ import {
 } from 'edge-core-js/types'
 import { base16 } from 'rfc4648'
 import {
+  DeletedNode,
   getBalanceChanges,
+  isCreatedNode,
+  isDeletedNode,
+  isModifiedNode,
   OfferCreate,
   Payment as PaymentJson,
   rippleTimeToUnixTime,
@@ -38,7 +45,10 @@ import {
   Wallet
 } from 'xrpl'
 import { Amount } from 'xrpl/dist/npm/models/common'
-import { AccountTxResponse } from 'xrpl/dist/npm/models/methods/accountTx'
+import {
+  AccountTxResponse,
+  AccountTxTransaction
+} from 'xrpl/dist/npm/models/methods/accountTx'
 import { validatePayment } from 'xrpl/dist/npm/models/transactions/payment'
 
 import { CurrencyEngine } from '../common/CurrencyEngine'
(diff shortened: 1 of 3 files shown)


src/ripple/RippleEngine.ts
```diff
@@ -275,6 +303,155 @@ export class XrpEngine extends CurrencyEngine<
     }
   }
 
+  /**
+   * Parse TakerGets or TakerPays into an EdgeAssetAmount
+   * */
+  parseRippleDexTxAmount = (
```
paullinator, line 309: Replace most of parseRippleDexTxAmount and processRippleDexTx with getBalances to determine src/dest amounts of order fulfillments.

paullinator (changes requested):

src/ripple/RippleEngine.ts
Jon-edge: The way I had it done is more accurate and complete according to [xrpscan.com](http://xrpscan.com/). 

The balances object is good for including mainnet fees in the xrp amount total, but for some reason underreports the actual amounts involved in the trades (regardless of fees - token amounts are underreported). Also there's no way to tell unfilled order amounts from the `Balances` obj either.

Jon-edge: Comparing against xrpscan.com shows alignment between my amounts and the summary amounts given by api queries to xrp explorers

src/ripple/RippleEngine.ts
```diff
@@ -275,6 +303,155 @@ export class XrpEngine extends CurrencyEngine<
...
+      // Order was fully filled
+      const isFullyFilled = hasDeletedNodes && !isOpenOrder
+
+      // Don't care about partial fills - counting them as general fills
+      type =
+        isFullyFilled || isPartiallyFilled ? 'swapOrderFill' : 'swapOrderPost'
+
+      // Parse amounts
+      const { TakerPays, TakerGets } = tx
+      sourceAsset = this.parseRippleDexTxAmount(TakerGets)
```
paullinator, line 409: This always get the amounts from the `tx` object. But in the case of a partial order fulfillment, the correct amounts are in the Nodes.

paullinator (changes requested):

paullinator (approved):

src/ripple/RippleEngine.ts
Jon-edge: We don't actually display the amounts from here at this time. 

This implementation is tracking 2 things consistently across all order states: the state of the order, and the posted order amount. 

Fill amount is a completely separate thing and I think is a mistake to conflate into the `source/destAsset` amounts described here.

Also if we only track fill amount, we actually are mis-reporting later down the road when the order status changes. We won't have knowledge of that unless we go back and modify related transactions as we process newer transactions. Imo, fill amount should be a completely separate field.

At this point in time, the only thing we are certain of is the total amounts of the original offer. Even if we have a partial fill immediately upon posting, we don't know if a later tx from someone else further modified the fill amount.

src/ripple/RippleEngine.ts
```diff
@@ -275,6 +303,155 @@ export class XrpEngine extends CurrencyEngine<
...
+    const {
+      currency,
+      issuer,
+      value
+      // Taker pays/gets XRP if 'TakerPays/Gets' is a plain string
+    } =
+      typeof takerAmount === 'string'
+        ? { currency: 'XRP', issuer: undefined, value: takerAmount }
+        : takerAmount
+    const isTakerToken = currency !== 'XRP' && issuer != null
```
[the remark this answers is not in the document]
Jon-edge: I'll add the assert, but we still need to keep the entire `isTakerToken` statement, or else `makeTokenId` right after complains

PR #645, opened 2023-10-16
Add Calibration Filecoin Testnet

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205724542492733

### DIFF
--- a/src/filecoin/filecoinInfo.ts
+++ b/src/filecoin/filecoinInfo.ts
@@ -8,6 +8,7 @@ const networkInfo: FilecoinNetworkInfo = {
   filfoxUrl: 'https://filfox.info/api/v1',
   filscanUrl: 'https://api-v2.filscan.io/api/v1',
   hdPathCoinType: 461,
+  networkPrefix: 'Mainnet',
   rpcNode: {
     networkName: 'Mainnet',
     url: 'https://api.node.glif.io/'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- added: Filecoin Testnet (Calibration)
+
 ## 2.7.2 (2023-10-20)
 
 - added: Add deprecated `memoType` to zcashInfo for backwards compatibility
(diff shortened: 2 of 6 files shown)


src/filecoin/filecoinInfo.ts
```diff
@@ -1,3 +1,4 @@
+import { NetworkPrefix } from '@zondax/izari-filecoin'
```
swansontec, line 1: We cannot currency libraries from the info files, either directly or transitively. However, `import type` statements are safe to use, because those get compiled out.

So, put this definition in `filecoinTypes.ts`:

```ts
import type { Network, NetworkPrefix } from '@zondax/izari-filecoin'

export interface FilecoinNetworkInfo {
   ...
  networkPrefix: keyof typeof NetworkPrefix
}
```

Now this will be equivalent to `networkPrefix: 'Mainnet' | 'Testnet'`, which means `filecoinInfo.ts` can just do `networkPrefix: 'Mainnet'` with no run-time imports.

Finally, when you go to use this flag in `Wallet.deriveAccount`, you will have to do `import { NetworkPrefix } from '@zondax/izari-filecoin'`, and then you can do `NetworkPrefix[this.networkInfo.networkPrefix]` to translate the enum key into the enum value.

swansontec (changes requested):
This will crash our customers on older phones. We need to fix the import structure.

swansontec (approved):
Use fixup commits! Our assumption is that already-reviewed commits will never change, so there's no need to re-visit them. The reviewer only needs to check the fixups.

By failing to use fixup commits, even if the changes are trivial, you are forcing me to re-review the entire PR, since something that was supposed to be immutable (the already-reviewed stuff) is now open to any and all changes.

samholmes:
> Use fixup commits! Our assumption is that already-reviewed commits will never change, so there's no need to re-visit them. The reviewer only needs to check the fixups.
> 
> By failing to use fixup commits, even if the changes are trivial, you are forcing me to re-review the entire PR, since something that was supposed to be immutable (the already-reviewed stuff) is now open to any and all changes.

True. I got lazy. There is also the compare button for force-pushes however: https://github.com/EdgeApp/edge-currency-accountbased/compare/8142b0c0083a587827af043b3b4716e12a03937c..8ff31b4e05a05ae8189d4ca494e23e29eb180502

PR #647, opened 2023-10-17
Fix Zcash memos

peachbits:


swansontec (approved):
We can also merge this for the sake of legacy clients, but our new code should hopefully just ignore this old stuff.

PR #652, opened 2023-10-24
Fix getMaxSpendable for Filecoin

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205794728237032

### DIFF
--- a/src/filecoin/FilecoinEngine.ts
+++ b/src/filecoin/FilecoinEngine.ts
@@ -8,7 +8,7 @@ import {
   Transaction,
   Wallet
 } from '@zondax/izari-filecoin'
-import { add, gt, gte, lte, mul, sub } from 'biggystring'
+import { add, eq, gt, mul, sub } from 'biggystring'
 import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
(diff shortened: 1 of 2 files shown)


src/filecoin/FilecoinEngine.ts
```diff
@@ -153,29 +163,40 @@ export class FilecoinEngine extends CurrencyEngine<
...
     spendInfo.skipChecks = true
 
-    // Continuously query networkFees using `makeSpend` until `tx.networkFee`
-    // is stable before returning the `spendableBalance` amount.
-    // This allows us to do a double-check on the network fee.
-    while (maxTries-- > 0) {
-      spendInfo.spendTargets[0].nativeAmount = spendableBalance
-      const tx = await this.makeSpend(spendInfo)
+    // Probe for an amount to use in our transaction:
+    const txForAmount = await this.makeSpend(spendInfo)
```
peachbits, line 174: makeSpend checks for a null spendTarget nativeAmount so we should make sure it has a value before calling it

src/filecoin/FilecoinEngine.ts
```diff
@@ -204,6 +225,14 @@ export class FilecoinEngine extends CurrencyEngine<
 
     const txJson = transaction.toJSON()
 
+    // Override fee parameters if this is a max-spend:
+    if (this.lastMaxSpendable != null) {
+      const { nativeAmount, params } = this.lastMaxSpendable
+      if (eq(nativeAmount, nativeAmount)) {
+        Object.assign(txJson, params)
+      }
```
peachbits, line 233: Is it worth keeping around if the values aren't equal?

peachbits (changes requested):

peachbits (approved):

PR #657, opened 2023-11-01
XRP EdgeTxActionSwap Direction Fix

Jon-edge:
Fix for only the bugged src/dest direction fix.

Partial fill amounts that happen after the initial order tx will require
a future fix and api change.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205787716339747

### DIFF
--- a/src/ripple/RippleEngine.ts
+++ b/src/ripple/RippleEngine.ts
@@ -37,6 +37,7 @@ import {
   isCreatedNode,
   isDeletedNode,
   isModifiedNode,
+  ModifiedNode,
   OfferCreate,
   Payment as PaymentJson,
   rippleTimeToUnixTime,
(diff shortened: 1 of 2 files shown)


src/ripple/RippleEngine.ts
```diff
@@ -338,18 +339,23 @@ export class XrpEngine extends CurrencyEngine<
...
-        ? this.currencyInfo.denominations[0]
-        : this.builtinTokens[tokenId].denominations[0]
+
+    let takerDenom
+    if (tokenId == null) {
+      takerDenom = this.currencyInfo.denominations[0]
+    } else {
+      const builtinToken = this.builtinTokens[tokenId]
+      if (builtinToken == null) return
+      takerDenom = this.builtinTokens[tokenId]?.denominations[0]
```
paullinator, line 349: change to
`takerDenom = builtInToken.denominations[0]`

This will fix typescript issue

paullinator, line 349: change to
`takerDenom = builtInToken.denominations[0]`

This will fix typescript issue

src/ripple/RippleEngine.ts
```diff
@@ -338,18 +339,23 @@ export class XrpEngine extends CurrencyEngine<
...
+    }
     if (takerDenom == null) {
-      this.error(`parseRippleDexTxAmount: Unknown denom ${currency}`)
+      // Shouldn't happen
```
paullinator, line 352: This should be able to be removed after the change above

paullinator (approved):
Approved but just requesting a couple small changes to clean up the code

PR #658, opened 2023-11-02
Support type 2 EVM transaction with EIP1559 parameters

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205751906499650

### DIFF
--- a/src/ethereum/fees/ethMiningFees.ts
+++ b/src/ethereum/fees/ethMiningFees.ts
@@ -1,7 +1,8 @@
-import Common from '@ethereumjs/common'
-import { Transaction } from '@ethereumjs/tx'
+import { Common } from '@ethereumjs/common'
+import { TransactionFactory } from '@ethereumjs/tx'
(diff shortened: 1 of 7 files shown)


src/ethereum/fees/ethMiningFees.ts
```diff
@@ -277,3 +277,32 @@ function flatMap<T>(items: NestedArray<T>, destinationItems: T[] = []): T[] {
...
+
+export type GasParams =
+  | { gasPrice: string }
+  | {
+      maxPriorityFeePerGas: string
+      maxFeePerGas: string
+    }
+
+export async function getFeeParamsByTransactionType(
+  type: number,
```
swansontec, line 289: This variable name, `type`, doesn't give much info. Can we rename it, or put a comment saying "The type comes from as `TransactionFactory.fromTxData`" or such.

swansontec (changes requested):
Small changes needed.

swansontec (approved):

PR #659, opened 2023-11-03
On-Chain Batch Balance Query

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

Use eth-balance-checker contracts for batch token balance queries for ETH, BSC, and MATIC

---
  - https://app.asana.com/0/0/1205654034220624

### DIFF
--- a/src/ethereum/EthereumNetwork.ts
+++ b/src/ethereum/EthereumNetwork.ts
@@ -1,8 +1,10 @@
 import { add, div, mul, sub } from 'biggystring'
 import { EdgeTransaction, JsonObject } from 'edge-core-js/types'
+import { ethers } from 'ethers'
 import { FetchResponse } from 'serverlet'
 import parse from 'url-parse'
 
+import { asMaybeContractLocation } from '../common/tokenHelpers'
 import {
   asyncWaterfall,
   cleanTxLogs,
@@ -17,6 +19,7 @@ import {
   shuffleArray,
   snooze
 } from '../common/utils'
+import ETH_BAL_CHECKER_ABI from './abi/ETH_BAL_CHECKER_ABI.json'
 import { WEI_MULTIPLIER } from './ethereumConsts'
 import { EthereumEngine } from './EthereumEngine'
 import {
@@ -60,6 +63,7 @@ const NUM_TRANSACTIONS_TO_QUERY = 50
 interface EthereumNeeds {
   blockHeightLastChecked: number
   nonceLastChecked: number
+  tokenBalsLastChecked: number
   tokenBalLastChecked: { [currencyCode: string]: number }
   tokenTxsLastChecked: { [currencyCode: string]: number }
 }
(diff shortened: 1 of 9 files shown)


src/ethereum/EthereumNetwork.ts
```diff
@@ -1656,13 +1755,30 @@ export class EthereumNetwork {
...
           preUpdateBlockHeight,
-          async () => await this.check('tokenBal', tk)
+          async () => await this.check('tokenBal')
         )
+      }
+
+      for (const tk of currencyCodes) {
+        // Only check each code individually if this engine does not support
+        // batch token balance queries.
+        if (ethBalCheckerContract == null) {
```
peachbits, line 1774: If you move this for loop block into an else condition above you can remove this duplicate ethBalCheckerContract check

src/ethereum/EthereumNetwork.ts
```diff
@@ -1502,6 +1510,97 @@ export class EthereumNetwork {
...
+    )
+    if (ethBalCheckerContract != null && jsonRpcUrl != null) {
+      const ethProvider = new ethers.providers.JsonRpcProvider(
+        jsonRpcUrl,
+        chainParams.chainId
+      )
+
+      const mainnetTokenAddr = '0x0000000000000000000000000000000000000000'
+      const tokenAddrs = [
+        ...Object.values(builtinTokens)
```
peachbits, line 1538: Have to use allTokensMap instead of builtinTokens otherwise you won't lookup custom tokens

peachbits (changes requested):

src/ethereum/EthereumNetwork.ts
```diff
@@ -1502,6 +1514,88 @@ export class EthereumNetwork {
...
+    })
+
+    // Randomize provider priority to distribute RPC provider load
+    funcs = shuffleArray(funcs)
+    const balances = await asyncWaterfall(funcs).catch(e => {
+      this.logError(
+        'checkEthBalChecker',
+        new Error(`All rpc servers failed eth balance checks: ${e}`)
+      )
+      return []
```
peachbits, line 1567: Throw the error instead of returning an array. If this fails we shouldn't pretend that it didn't.

src/ethereum/EthereumNetwork.ts
```diff
@@ -1502,6 +1514,88 @@ export class EthereumNetwork {
...
+
+      const contract = new ethers.Contract(
+        ethBalCheckerContract,
+        ETH_BAL_CHECKER_ABI,
+        ethProvider
+      )
+
+      funcs.push(
+        async () =>
+          await contract.balances(
```
peachbits, line 1553: Do some sort of data validation within this callback before returning the response from `balances`. Something like checking the array length compared to the list of tokens you originally provided would suffice. throw if it's unexpected

peachbits (changes requested):

peachbits (approved):

PR #661, opened 2023-11-07
Update bridged USDC currency code to USDC.e

peachbits:
---
  - https://app.asana.com/0/0/1205865331204026


swansontec (approved):
Approved, but let's get the exchange fixes in place before we merge this.

PR #666, opened 2023-11-09
ENS Name Resolution

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205920007236365

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@
 - added: Add `MakeTxDeposit` to support THORChain swap transactions
 - changed: Move `makeTx` params types to common
 - fixed: Possible error in ETH transactions when calculated gasPrice is below baseFee
-
-## 2.12.0 (2023-11-14)
+- added: ENS name resolution support
 
 ## 2.11.2 (2023-11-13)
 
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -11,6 +11,7 @@ import {
   EdgeTransaction,
   JsonObject
 } from 'edge-core-js/types'
+import { ethers } from 'ethers'
 import { base16 } from 'rfc4648'
 
 export function normalizeAddress(address: string): string {
@@ -476,4 +477,24 @@ export const objectCheckOneWay = (obj1: any, obj2: any): boolean => {
     }
   }
   return true
+}
+
+/**
+ * Calls `func` on ethers JsonRpcProviders initialized with configured
+ * RPC servers. Randomizes order priority to distribute load.
+ */
+export const multicastEthProviders = async <
+  R,
+  P extends ethers.providers.Provider
+>(props: {
+  func: (ethProvider: P) => Promise<R>
+  providers: P[]
+}): Promise<R> => {
+  const { func, providers } = props
+  const funcs: Array<() => Promise<any>> = providers.map(
+    provider => async () => {
+      return await func(provider)
+    }
+  )
+  return await asyncWaterfall(shuffleArray(funcs))
 }
--- a/src/ethereum/EthereumTools.ts
+++ b/src/ethereum/EthereumTools.ts
@@ -14,15 +14,21 @@ import {
 } from 'edge-core-js/types'
 import EthereumUtil from 'ethereumjs-util'
 import hdKey from 'ethereumjs-wallet/hdkey'
+import { ethers } from 'ethers'
 
 import { PluginEnvironment } from '../common/innerPlugin'
 import { asMaybeContractLocation, validateToken } from '../common/tokenHelpers'
 import { encodeUriCommon, parseUriCommon } from '../common/uriHelpers'
-import { biggyScience, getLegacyDenomination } from '../common/utils'
+import {
+  biggyScience,
+  getLegacyDenomination,
+  multicastEthProviders
+} from '../common/utils'
 import { ethereumPlugins } from './ethereumInfos'
 import {
   asEthereumPrivateKeys,
   asSafeEthWalletInfo,
+  EthereumInitOptions,
   EthereumNetworkInfo
 } from './ethereumTypes'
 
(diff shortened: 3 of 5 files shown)


src/common/utils.ts
```diff
@@ -477,3 +483,52 @@ export const objectCheckOneWay = (obj1: any, obj2: any): boolean => {
...
+    rpcServer => async () => {
+      const ethProvider = new ethers.providers.JsonRpcProvider(
+        addRpcApiKey(initOptions, pluginId, rpcServer),
+        chainId
+      )
+      return await func(ethProvider)
+    }
+  )
+  return await asyncWaterfall(shuffleArray(funcs))
+}
```
samholmes, line 534: This should be in its own commit.

src/common/utils.ts
```diff
@@ -141,7 +147,7 @@ export const snoozeReject: Function = async (ms: number) =>
 export const snooze: Function = async (ms: number) =>
   await new Promise((resolve: Function) => setTimeout(resolve, ms))
 
-export async function promiseAny<T>(promises: Array<Promise<T>>): Promise<T> {
+export async function promiseAny(promises: Array<Promise<any>>): Promise<any> {
```
samholmes, line 150: Regression.

src/ethereum/networkAdapters/RpcAdapter.ts
```diff
@@ -396,7 +377,11 @@ export class RpcAdapter
       params
     }
 
-    url = this.addRpcApiKey(url)
+    url = addRpcApiKey(
+      this.ethEngine.initOptions,
+      this.ethEngine.currencyInfo.pluginId,
+      url
+    )
```
samholmes, line 384: I think I've discovered this to be a bug. We're adding API key to the server urls in the constructor, so the url passed into this call should already include the api key. Make sure this private method is always called with the a `baseUrl` from the RpcAdapter config, and if so, remove this as a separate fix commit.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-## 2.12.0 (2023-11-14)
```
samholmes, line 5: This should have been fixed because the version was published with some real changes. Removing it is not the right thing. I added a fix commit to master, but be sure to resolve your conflict here correctly.

src/ethereum/EthereumTools.ts
```diff
@@ -14,15 +14,21 @@ import {
...
+import { ethers } from 'ethers'
 
 import { PluginEnvironment } from '../common/innerPlugin'
 import { asMaybeContractLocation, validateToken } from '../common/tokenHelpers'
 import { encodeUriCommon, parseUriCommon } from '../common/uriHelpers'
-import { biggyScience, getLegacyDenomination } from '../common/utils'
+import {
+  biggyScience,
+  getLegacyDenomination,
+  multicastEthProviders
```
samholmes, line 25: This is the commit which introduces `multicastEthProviders`

src/common/utils.ts
```diff
@@ -477,3 +483,52 @@ export const objectCheckOneWay = (obj1: any, obj2: any): boolean => {
...
+  rpcServers: string[]
+  initOptions: EthereumInitOptions
+  pluginId: string
+  chainId?: number
+}): Promise<any> => {
+  const { func, rpcServers, initOptions, pluginId, chainId } = props
+  const funcs: Array<() => Promise<any>> = rpcServers.map(
+    rpcServer => async () => {
+      const ethProvider = new ethers.providers.JsonRpcProvider(
+        addRpcApiKey(initOptions, pluginId, rpcServer),
```
samholmes, line 527: Add the API key before calling `multicastEthProviders` and now `multicastEthProviders` doesn't need the extra parameters.

Also, instead of `rpcServers: string[]` you can paramterize `providers: ethers.providers.JsonRpcProvider[]`. In the end the function signature could look like:

```ts
export const multicastEthProviders = async <R, P extends ethers.providers.Provider>(props: {
  func: (ethProvider: P) => Promise<R>
  providers: P[]
}): Promise<R>
```

If you want to be more narrow then `Providers` type then change the subtype for `P` to `ethers.providers.JsonRpcProvider`.

src/ethereum/EthereumTools.ts
```diff
@@ -357,8 +365,35 @@ export class EthereumTools implements EdgeCurrencyTools {
     }
     return cleanLocation.contractAddress.toLowerCase().replace(/^0x/, '')
   }
+
+  // #region otherMethods
+
+  /**
+   * Resolve an ENS name, for example: "bob.eth"
+   */
+  async resolveEnsName(ensName: string): Promise<string> {
```
samholmes, line 374: Why not make this called `resolveName(name: string): Promise<string>` just incase other name services are deployed besides ENS?

samholmes (changes requested):

src/ethereum/EthereumTools.ts
Jon-edge: This implementation is specific to ENS since it only works for ENS on ETH. If we deploy other name services that can be generic (including support in ETH), then we can do this

samholmes: Fair enough, though I'd imagine the function signature would be the same. I only thought that maybe if we make it generic and just fail on the input if it's not an ENS name (or whatever the implementation supports), then there wouldn't be a need for a new method later on. It's forward thinking, so not necessary.

src/ethereum/networkAdapters/RpcAdapter.ts
Jon-edge: No the constructor requires an `EthereumEngine`, so it won't work.

Jon-edge: `addRpcApiKey `isn't needed. The api keys are always populated in the `networkAdapterConfig`

samholmes (approved):

PR #668, opened 2023-11-12
Fix failed sync of evm wallets

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205935097933115

### DIFF
--- a/src/ethereum/EthereumNetwork.ts
+++ b/src/ethereum/EthereumNetwork.ts
@@ -1537,7 +1537,13 @@ export class EthereumNetwork {
 
     let funcs: Array<() => Promise<any>> = []
     rpcServers.forEach(rpcServer => {
-      const rpcServerWithApiKey = this.addRpcApiKey(rpcServer)
+      let rpcServerWithApiKey: string
+      try {
+        rpcServerWithApiKey = this.addRpcApiKey(rpcServer)
+      } catch (e) {
+        // addRpcApiKey can throw if there's a missing apikey. skip this rpcServer
+        return
+      }
       const ethProvider = new ethers.providers.JsonRpcProvider(
         rpcServerWithApiKey,
         chainParams.chainId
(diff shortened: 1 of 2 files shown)


src/ethereum/EthereumNetwork.ts
```diff
@@ -1733,14 +1739,22 @@ export class EthereumNetwork {
         this.ethNeeds.blockHeightLastChecked,
         BLOCKHEIGHT_POLL_MILLISECONDS,
         preUpdateBlockHeight,
-        async () => await this.check('blockheight')
+        async () => {
+          console.log(`nonce`)
```
samholmes, line 1743: Cleanup console.logs here. Or if you want a real log for the checker, then move it into the check implementation?

src/ethereum/EthereumNetwork.ts
```diff
@@ -1537,7 +1537,13 @@ export class EthereumNetwork {
...
     let funcs: Array<() => Promise<any>> = []
     rpcServers.forEach(rpcServer => {
-      const rpcServerWithApiKey = this.addRpcApiKey(rpcServer)
+      let rpcServerWithApiKey: string
+      try {
+        rpcServerWithApiKey = this.addRpcApiKey(rpcServer)
+      } catch (e) {
+        // addRpcApiKey can throw if there's a missing apikey. skip this rpcServer
+        return
+      }
```
samholmes, line 1546: Hmm, if all RPC servers fail on this adding of key step, then this means balance checking via RPC is not possible. There are other methods to get balances which wont be used because of the condition on line 1787. This can surface itself as a bug as well, so this only partly fixes the problem.

samholmes (changes requested):

samholmes (approved):

src/ethereum/EthereumNetwork.ts
paullinator: cleaned up

PR #673, opened 2023-11-21
Add Osmosis

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

Add Osmosis
Add chain data, dynamic fees, and optimize archive/validator rpc usage
Cleanup various exports across all currencies

---
  - https://app.asana.com/0/0/1205856665924152

### DIFF
--- a/src/cosmos/CosmosEngine.ts
+++ b/src/cosmos/CosmosEngine.ts
@@ -6,7 +6,7 @@ import {
   encodePubkey,
   makeAuthInfoBytes
 } from '@cosmjs/proto-signing'
-import { parseCoins, StargateClient } from '@cosmjs/stargate'
+import { coin, parseCoins, StargateClient } from '@cosmjs/stargate'
 import { parseRawLog } from '@cosmjs/stargate/build/logs'
 import {
   fromRfc3339WithNanoseconds,
@@ -137,7 +137,7 @@ export class CosmosEngine extends CurrencyEngine<
         this.walletInfo.keys.bech32Address
       )
       const mainnetBal = balances.find(
-        bal => bal.denom === this.currencyInfo.currencyCode.toLowerCase()
+        bal => bal.denom === this.networkInfo.nativeDenom
       )
       this.updateBalance(
         this.currencyInfo.currencyCode,
@@ -294,19 +294,19 @@ export class CosmosEngine extends CurrencyEngine<
     }
     const networkFee = networkFeeCoin.amount
 
-    const { coin, recipient, sender } = event
+    const { coin: eventCoin, recipient, sender } = event
 
     if (
       (this.walletInfo.keys.bech32Address !== recipient &&
         this.walletInfo.keys.bech32Address !== sender) ||
-      this.currencyInfo.currencyCode !== coin.denom.toUpperCase()
+      this.networkInfo.nativeDenom !== eventCoin.denom
     ) {
       return
     }
     const isSend = this.walletInfo.keys.bech32Address === sender
     const ourReceiveAddresses: string[] = []
 
-    let nativeAmount = coin.amount
+    let nativeAmount = eventCoin.amount
     const currencyCode = this.currencyInfo.currencyCode
     if (isSend) {
       nativeAmount = `-${add(nativeAmount, networkFee)}`
@@ -403,7 +403,7 @@ export class CosmosEngine extends CurrencyEngine<
 
     // Encode a send message.
     const msg = this.tools.methods.transfer({
-      amount: nativeAmount,
+      amount: [coin(nativeAmount, this.networkInfo.nativeDenom)],
       fromAddress: this.walletInfo.keys.bech32Address,
       toAddress: publicAddress
     })
--- a/src/cosmos/CosmosEngine.ts
+++ b/src/cosmos/CosmosEngine.ts
@@ -1,3 +1,4 @@
+import { getGasPriceStep } from '@chain-registry/utils'
 import { decodeSignature, encodeSecp256k1Pubkey } from '@cosmjs/amino'
 import { toHex } from '@cosmjs/encoding'
 import {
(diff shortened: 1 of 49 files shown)


src/cosmos/CosmosTools.ts
```diff
@@ -53,11 +56,22 @@ export class CosmosTools implements EdgeCurrencyTools {
     )
     this.methods = methods
     this.registry = registry
+    const { data, name, url } = this.networkInfo.chainData
+    this.chainData = data
```
paullinator, line 60: This is inconsistent naming. I'd recommend renaming `this.networkInfo.chainData` to `this.networkInfo.chainInfo`

src/cosmos/CosmosTools.ts
```diff
@@ -53,11 +56,22 @@ export class CosmosTools implements EdgeCurrencyTools {
...
+    const chainUpdater = new ChainRegistryFetcher()
+    chainUpdater
+      .fetch(url)
+      .then(() => {
+        this.chainData = chainUpdater.getChain(name)
```
paullinator, line 65: Something makes me uneasy about knowing key derivation can be dependent on a network call.  What would we need updated from the network that would change from the hardcoded json?

src/cosmos/CosmosEngine.ts
```diff
@@ -203,6 +222,7 @@ export class CosmosEngine extends CurrencyEngine<
         progress
       this.updateOnAddressesChecked()
     }
+    this.otherData.archivedTxLastCheckTime = Date.now()
```
paullinator, line 225: Just a comment for future consideration:

`CurrencyEngine.walletLocalData` already has (but poorly named) `lastTransactionQueryHeight`. To avoid having custom values in `otherData`, this could be used. I would suggest renaming it, but it's already saved on disk for EVM users.

src/cosmos/CosmosTools.ts
```diff
@@ -196,14 +196,11 @@ export class CosmosTools implements EdgeCurrencyTools {
...
-      const stargateClient = await createStargateClient(
-        this.io.fetchCors,
-        rpcWithApiKey(this.networkInfo, this.initOptions)
-      )
-      // eslint-disable-next-line @typescript-eslint/dot-notation
-      const queryClient = stargateClient['forceGetQueryClient']()
-      // eslint-disable-next-line @typescript-eslint/dot-notation
-      const tendermintClient = stargateClient['forceGetTmClient']()
+      const { queryClient, stargateClient, tendermintClient } =
+        await createStargateClient(
```
paullinator, line 200: This now seems misnamed. `createCosmosClients` instead?

Why bother destructuring? Just assign `this.clients = await createStargateClients()`

src/cosmos/CosmosEngine.ts
```diff
@@ -192,8 +197,22 @@ export class CosmosEngine extends CurrencyEngine<
...
+            this.fetchCors,
+            this.networkInfo.archiveNode
+          )
+        : // Otherwise, uses regular rpc
+          this.getClients()
+
     for (const query of txQueryStrings) {
-      const newestTxid = await this.queryTransactionsInner(query)
+      const newestTxid = await this.queryTransactionsInner(
+        query,
```
paullinator, line 212: Just pass the full `CosmosClients` to `queryTransactionsInner` and remove the destructuring

src/cosmos/cosmosRegistry.ts
```diff
@@ -1,16 +1,34 @@
 import { fromBech32 } from '@cosmjs/encoding'
 import { EncodeObject, Registry } from '@cosmjs/proto-signing'
+import { MsgSend } from 'cosmjs-types/cosmos/bank/v1beta1/tx'
 
 import { DepositOpts, TransferOpts, UpgradedRegistry } from './cosmosTypes'
 import { assetFromString } from './cosmosUtils'
 import { MsgDeposit } from './info/proto/thorchainrune/thorchain/v1/x/thorchain/types/msg_deposit'
-import { MsgSend } from './info/proto/thorchainrune/thorchain/v1/x/thorchain/types/msg_send'
+import { MsgSend as thorchainruneMsgSend } from './info/proto/thorchainrune/thorchain/v1/x/thorchain/types/msg_send'
```
paullinator, line 8: Since this seems like a class, name should be camel-case `ThorchainRuneMsgSend`

paullinator (changes requested):

src/cosmos/CosmosEngine.ts
```diff
@@ -290,12 +291,15 @@ export class CosmosEngine extends CurrencyEngine<
       body: { memo }
     } = decodeTxRaw(txRaw)
 
-    let networkFeeCoin = this.networkInfo.defaultTransactionFee
+    let networkFee = this.networkInfo.defaultTransactionFee?.amount ?? '0'
     if (fee != null) {
       const { amount } = fee
-      networkFeeCoin = safeAddCoins([networkFeeCoin, ...amount])
+      const networkFeeCoin = safeAddCoins([
+        coin('0', this.networkInfo.nativeDenom),
```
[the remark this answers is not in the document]
peachbits: safeAddCoins will add up coins with the same denom and this forces the calculation to use the native denom instead of some other currency. It's the same behavior as filtering the amount array first

src/cosmos/CosmosTools.ts
peachbits: The minimum fees were the main concern. I can see hardcoding the derivation as preferred I'll re-add that to the infos

src/cosmos/cosmosRegistry.ts
peachbits: done

src/cosmos/CosmosEngine.ts
peachbits: done

src/cosmos/CosmosTools.ts
peachbits: done

src/cosmos/CosmosTools.ts
```diff
@@ -56,22 +56,22 @@ export class CosmosTools implements EdgeCurrencyTools {
     )
     this.methods = methods
     this.registry = registry
-    const { data, name, url } = this.networkInfo.chainData
-    this.chainData = data
+    const { data, name, url } = this.networkInfo.chainInfo
+    this.chainInfo = data
```
paullinator: this.chainInfo should stay as `chainData`

paullinator (changes requested):

PR #679, opened 2023-11-29
Solana: Add token support

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205407182344883

### DIFF
--- a/src/solana/SolanaEngine.ts
+++ b/src/solana/SolanaEngine.ts
@@ -1,6 +1,6 @@
 import * as solanaWeb3 from '@solana/web3.js'
 import { add, gt, mul } from 'biggystring'
-import { asNumber } from 'cleaners'
+import { asArray, asNumber, asString } from 'cleaners'
 import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
@@ -29,6 +29,7 @@ import {
   asRecentBlockHash,
   asRpcBalance,
   asRpcGetTransaction,
+  asRpcSignatureForAddress,
   asSafeSolanaWalletInfo,
   asSolanaPrivateKeys,
   asSolanaWalletOtherData,
@@ -84,8 +85,7 @@ export class SolanaEngine extends CurrencyEngine<
     this.otherData = asSolanaWalletOtherData(raw)
   }
 
-  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-  async fetchRpc(method: string, params: any = []) {
+  async fetchRpc(method: string, params: any = []): Promise<unknown> {
     const body = {
       jsonrpc: '2.0',
       id: 1,
@@ -115,8 +115,7 @@ export class SolanaEngine extends CurrencyEngine<
     return response.result
   }
 
-  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-  async queryBalance() {
+  async queryBalance(): Promise<void> {
     try {
       const response = await this.fetchRpc('getBalance', [
         this.base58PublicKey,
@@ -130,8 +129,7 @@ export class SolanaEngine extends CurrencyEngine<
     }
   }
 
-  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-  async queryBlockheight() {
+  async queryBlockheight(): Promise<void> {
     try {
       const blockheight = asNumber(await this.fetchRpc('getSlot'))
       if (blockheight > this.walletLocalData.blockHeight) {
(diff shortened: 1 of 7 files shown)


src/solana/SolanaEngine.ts
```diff
@@ -361,33 +369,133 @@ export class SolanaEngine extends CurrencyEngine<
...
+    let parentNetworkFee: string | undefined
+    const balance = this.walletLocalData.totalBalances[currencyCode] ?? '0'
+
+    // pubkeys
+    const PAYER = new PublicKey(this.base58PublicKey)
+    const PAYEE = new PublicKey(publicAddress)
+    const TOKEN_PROGRAM_ID = new PublicKey(this.networkInfo.tokenPublicKey)
+    const ASSOCIATED_TOKEN_PROGRAM_ID = new PublicKey(
+      this.networkInfo.associatedTokenPublicKey
+    )
```
swansontec, line 502: These variable names don't fit our convention. Is there some reason?

src/solana/SolanaEngine.ts
```diff
@@ -448,6 +557,12 @@ export class SolanaEngine extends CurrencyEngine<
     })
     edgeTransaction.signedTx = solTx.serialize().toString('base64')
     this.warn(`signTx\n${cleanTxLogs(edgeTransaction)}`)
+
+    // Remove all false values from addressCache
+    this.addressCache = new Map(
+      Array.from(this.addressCache.entries()).filter(([_, value]) => value)
+    )
```
swansontec, line 681: Optional: This code is creating 2 extra arrays and 1 extra map. Doing it in a loop won't create any extra data structures, so it should be a lot simpler & faster:

```ts
for (const [key, value] of this.addressCache.entries()) {
  if (!value) this.addressCache.delete(key)
}
```

src/solana/SolanaEngine.ts
```diff
@@ -273,52 +278,47 @@ export class SolanaEngine extends CurrencyEngine<
...
+      )
+      this.otherData.newestTxid = txids[i].signature
+
+      // Update progress
+      const percent = 1 - i / txids.length
+      if (percent !== this.progressRatio) {
+        if (Math.abs(percent - this.progressRatio) > 0.25 || percent === 1) {
+          this.progressRatio = percent
+          this.tokenCheckTransactionsStatus[this.chainCode] = this.progressRatio
+          this.updateOnAddressesChecked()
```
swansontec: It seems weird to maintain a fractional progress ratio in here. Since it's only one network query now, wouldn't we just report 0 (not queried) or 1 (queried) for each token?

Even if there is some paging going on, I would expect the progress code to move outside the loop, at the same level as the network request, instead of inside the result-processing loop.

src/solana/SolanaEngine.ts
```diff
@@ -121,13 +124,41 @@ export class SolanaEngine extends CurrencyEngine<
     }
 
     const funcs = this.networkInfo.rpcNodes.map(serverUrl => async () => {
+      const apiKeys = asSolanaInitOptions(this.initOptions) as {
+        [key: string]: string
+      }
+      const regex = /{{(.*?)}}/g
```
swansontec, line 130: I don't see why the `?` is needed. This should do the exact same thing: `/{{(.*)}}/g`

src/solana/SolanaEngine.ts
```diff
@@ -151,13 +151,7 @@ export class SolanaEngine extends CurrencyEngine<
...
       const out = await res.json()
 
-      // look for jsonrpc error
-      if (out[0].error != null) {
-        throw new Error(
-          `fetchRpc ${options.method} failed error: ${out.error.message}`
-        )
-      }
-
+      // this.log.warn(serverUrl)
```
swansontec, line 154: Leftover comment?

swansontec (changes requested):
Found a few small things, but this makes sense overall.

src/solana/SolanaEngine.ts
```diff
@@ -85,19 +89,24 @@ export class SolanaEngine extends CurrencyEngine<
...
     const options = {
       method: 'POST',
       headers: {
         Accept: 'application/json',
         'Content-Type': 'application/json'
       },
-      body: JSON.stringify(body)
+      body: JSON.stringify(
+        requests.map(req => ({ jsonrpc: '2.0', id: 1, ...req }))
+      )
```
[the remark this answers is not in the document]
peachbits: Yup, despite that it was working this should be changed

src/solana/SolanaEngine.ts
peachbits: No good reason, will fix

src/solana/SolanaEngine.ts
swansontec: Oh, that makes sense.

PR #683, opened 2023-12-01
(cosmos-based) Add unconfirmed nonce handling

peachbits:
---
  - https://app.asana.com/0/0/1206053210000460


src/cosmos/CosmosEngine.ts
```diff
@@ -377,6 +391,48 @@ export class CosmosEngine extends CurrencyEngine<
...
+          txidsToRemove.push(txid)
+        }
+      }
+      for (const txid of txidsToRemove) {
+        this.unconfirmedTransactionCache.txids.delete(txid)
+      }
+
+      return (
+        this.unconfirmedTransactionCache.cacheSequence +
+        this.unconfirmedTransactionCache.txids.size
```
paullinator, line 431: if any txs got removed assume all are expired

paullinator (changes requested):

paullinator (approved):

PR #698, opened 2023-12-16
Add BONK (Solana)

peachbits:
---
  - https://app.asana.com/0/0/1206185605352700

### DIFF
--- a/src/solana/solanaInfo.ts
+++ b/src/solana/solanaInfo.ts
@@ -6,6 +6,14 @@ import type { SolanaTools } from './SolanaTools'
 import type { SolanaNetworkInfo } from './solanaTypes'
 
 const builtinTokens: EdgeTokenMap = {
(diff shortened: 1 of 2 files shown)


swansontec (approved):

src/solana/solanaInfo.ts
```diff
@@ -6,6 +6,14 @@ import type { SolanaTools } from './SolanaTools'
 import type { SolanaNetworkInfo } from './solanaTypes'
 
 const builtinTokens: EdgeTokenMap = {
+  DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263: {
```
swansontec, line 9: Are Solana tokenId's not lower case? I know on Ethereum we forced these to be lowercase, but maybe this chain does it differently.

peachbits: Didn't bother to lowercase it when originally implemented. Other chains outside of evms are the same

PR #703, opened 2023-12-22
Paul/fio stake fix

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206227159432533


samholmes (approved):
This seems to make sense. Does Tron also use the first two indexes as the locked/staked balances?

paullinator:
> This seems to make sense. Does Tron also use the first two indexes as the locked/staked balances?

They do not but Tron has its own staking plugin that deciphers the stakingStatus. 

Ideally FIO will also have its own plugin

PR #717, opened 2024-02-09
Add Coreum tokens

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206548993975489

### DIFF
--- a/src/cosmos/cosmosUtils.ts
+++ b/src/cosmos/cosmosUtils.ts
@@ -224,7 +224,7 @@ export const extendedParseCoins = (input: string): Coin[] => {
     .map(part => {
       // Denom regex from Stargate (https://github.com/cosmos/cosmos-sdk/blob/v0.42.7/types/coin.go#L599-L601)
       const match = part.match(/^([0-9]+)([a-zA-Z][a-zA-Z0-9/-]{2,127})$/)
-      if (match == null) throw new Error('Got an invalid coin string')
+      if (match == null) throw new Error(`Got an invalid coin string: ${input}`)
       return {
         amount: match[1].replace(/^0+/, '') ?? '0',
         denom: match[2]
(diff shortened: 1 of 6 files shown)


src/cosmos/info/coreumInfo.ts
```diff
@@ -1,11 +1,132 @@
...
+  'usara-core1r9gc0rnxnzpq33u82f44aufgdwvyxv4wyepyck98m9v2pxua6naqr8h03z': {
+    currencyCode: 'SARA',
+    displayName: 'Pulsara',
+    denominations: [{ name: 'SARA', multiplier: '1000000' }],
+    networkLocation: {
+      contractAddress:
+        'usara-core1r9gc0rnxnzpq33u82f44aufgdwvyxv4wyepyck98m9v2pxua6naqr8h03z'
+    }
+  },
+  'ibc/E1E3674A0E4E1EF9C69646F9AF8D9497173821826074622D831BAB73CCB99A2D': {
```
paullinator, line 20: Ensure that these denoms actually need to be upper case. If not then use the lowercase for the `tokenId`

src/cosmos/cosmosUtils.ts
```diff
@@ -210,3 +202,20 @@ export const reduceCoinEventsForAddress = (
   })
   return out
 }
+
+// The helper function parseCoins from the @cosmjs sdk doesn't handle denoms with hyphens.
+const extendedParseCoins = (input: string): Coin[] => {
```
paullinator, line 207: This needs some unit tests. Also testing this with `usara-core1r9gc0rnxnzpq33u82f44aufgdwvyxv4wyepyck98m9v2pxua6naqr8h03z` causes the routine to throw. 

Should this work or is `amount` not the denom.

paullinator (changes requested):

src/cosmos/cosmosUtils.ts
peachbits: This parses amount/denom strings from transactions ie. 12345usara-core1r9gc0rnxnzpq33u82f44aufgdwvyxv4wyepyck98m9v2pxua6naqr8h03z

paullinator (approved):

PR #722, opened 2024-02-14
Add fallback balance query for unstake funds chain bug

peachbits:
This allows the wallet to still get a balance if they are affected by the unstake bug. The balance is necessary so they can take the necessary actions to fix their account's onchain data.

https://fioprotocol.atlassian.net/wiki/spaces/DAO/pages/852688908/2024-02-07+Token+Locking+Issue+on+Unstake

---
  - https://app.asana.com/0/0/1206526184999703

### DIFF
--- a/src/ethereum/fees/feeProviders.ts
+++ b/src/ethereum/fees/feeProviders.ts
@@ -261,12 +261,13 @@ export const fetchFeesFromEvmGasStation = async (
 
 export const fetchFeesFromInfoServer = async (
   fetch: EdgeFetchFunction,
-  { pluginId }: EdgeCurrencyInfo
+  { pluginId }: EdgeCurrencyInfo,
+  opts: { timeout?: number } = {}
 ): Promise<EthereumFees> => {
   const result = await fetchInfo(
     `v1/networkFees/${pluginId}`,
     undefined,
-    undefined,
+    opts.timeout,
     fetch
   )
   const json = await result.json()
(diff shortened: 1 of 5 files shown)


src/ethereum/fees/feeProviders.ts
```diff
@@ -261,12 +261,13 @@ export const fetchFeesFromEvmGasStation = async (
 
 export const fetchFeesFromInfoServer = async (
   fetch: EdgeFetchFunction,
-  { pluginId }: EdgeCurrencyInfo
+  { pluginId }: EdgeCurrencyInfo,
+  timeout?: number
```
swansontec, line 265: I would prefer `opts: { timeout?: number } = {}`.

src/fio/FioEngine.ts
```diff
@@ -952,6 +969,42 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
...
+            await timeout(
+              this.fioApiRequest(apiUrl, 'getCurrencyBalance', {
+                code: 'fio.token',
+                account: this.actor,
+                symbol: 'FIO'
+              }),
+              10000
+            )
+        ),
+        (result: any) => {
```
swansontec, line 985: I would prefer `(result: unknown)`

swansontec (approved):
Good work getting this fixed. I see that this is a partial solution, so we can at least see the balance.

PR #723, opened 2024-02-19
Update Pulsechain integrations for fetching transaction

samholmes:
### Note about this PR

The old API which is evmscan/blockscout compatible was added as a commit after I did the integration even though that commit proceeds the new integration commit. The new API commit is to include all the new API for PLS as a fallback adaptor now. I figured it’s worth merging this integration into the codebase because I spent the time working on it. The new API isn’t as good as the evmscan adaptor because I am unable to figure out how to get the “next page” parameters to match the passed “startBlock” parameter.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206163473998245

### DIFF
--- a/src/ethereum/networkAdapters/PulsechainScanAdapter.ts
+++ b/src/ethereum/networkAdapters/PulsechainScanAdapter.ts
@@ -1,0 +1,255 @@
+import { mul, sub } from 'biggystring'
+import {
+  asArray,
+  asBoolean,
+  asJSON,
+  asNumber,
+  asObject,
+  asOptional,
+  asString
+} from 'cleaners'
+import { EdgeTransaction } from 'edge-core-js/types'
+
+import { EthereumNetworkUpdate, getFeeRateUsed } from '../EthereumNetwork'
+import { EthereumTxOtherParams } from '../ethereumTypes'
+import { GetTxsParams, NetworkAdapter } from './types'
+
+export interface PulsechainScanAdapterConfig {
+  type: 'pulsechain-scan'
+  servers: string[]
+}
+
+export class PulsechainScanAdapter extends NetworkAdapter<PulsechainScanAdapterConfig> {
+  fetchBlockheight = null
+  broadcast = null
+  getBaseFeePerGas = null
+  multicastRpc = null
+  fetchNonce = null
+  fetchTokenBalance = null
+  fetchTokenBalances = null
+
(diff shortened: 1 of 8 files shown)


src/ethereum/networkAdapters/PulsechainScanAdapter.ts
```diff
@@ -0,0 +1,262 @@
...
+
+      // Exit if a transaction block height is less than/equal to the startBlock:
+      if (responseData.items.some(scanTx => scanTx.block <= startBlock)) break
+
+      // Exit if there is no next page of transactions:
+      if (nextPageParams == null) break
+    }
+
+    // Convert the transaction data into EdgeTransactions:
+
```
swansontec, line 74: Extra blank line here.

src/ethereum/networkAdapters/PulsechainScanAdapter.ts
```diff
@@ -0,0 +1,262 @@
...
+        [currencyCode]: {
+          blockHeight: startBlock,
+          edgeTransactions
+        }
+      },
+      server: this.config.servers.join(',')
+    }
+  }
+
+  private processScanTransaction(
```
swansontec, line 90: This method could be a standalone function, instead of a class member. It never touches `this`

src/ethereum/networkAdapters/PulsechainScanAdapter.ts
```diff
@@ -0,0 +1,262 @@
...
+      }
+    } else {
+      // Receive
+      if (tokenTx) {
+        nativeAmount = scanTx.value
+        networkFee = '0'
+      } else {
+        nativeAmount = scanTx.value
+        networkFee = '0'
+      }
```
swansontec, line 137: Both cases look the same here.

swansontec (approved):
Approved, but there are some weird things to look at / clean up.

src/ethereum/networkAdapters/PulsechainScanAdapter.ts
samholmes: It uses `this` four times.

src/ethereum/networkAdapters/PulsechainScanAdapter.ts
samholmes: Fixed, and fixed the source of this copy-pasta

PR #726, opened 2024-02-20
Force Zcash and Piratechain imports to have 24 words

peachbits:
This is enforced by the sdk. Throwing here prevents the wallet from being created

---
  - https://app.asana.com/0/0/1206270926523609

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - added: (PLS) Integrate new v2 API for Pulsechain as a pulsechain-scan adapter
+- changed: (Zcash/Piratechain) Enforce 24 word seed length for imports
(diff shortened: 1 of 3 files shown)


swansontec (changes requested):
There is no native try / catch being added in the Kotlin / Swift side!!!!!!! This means that any other formatting errors could still crash the app. This isn't a normal crash where you get a drop-down and the wallet doesn't appear - it makes the whole account unusable forever because the app dies any time you log in. If should not be possible for the Kotlin / Swift code to take down the whole app.

The check you have here will certainly improve things, but it's not very thorough. What if the user imports a string like "a           b"? This would pass your test, but I'm guessing zcash would still crash on it.

swansontec (approved):
New plan: We can fix the other stuff later. This is fine for now.

PR #728, opened 2024-02-21
Ignore base currency in internal transactions

peachbits:
---
  - https://app.asana.com/0/0/1206654853147961


peachbits:
Not a safe assumption, closing

PR #729, opened 2024-02-23
Cache and fuzz Arbitrum fees

peachbits:
---
  - https://app.asana.com/0/0/1206670220740547

### DIFF
--- a/src/ethereum/fees/ethMiningFees.ts
+++ b/src/ethereum/fees/ethMiningFees.ts
@@ -329,6 +329,12 @@ export async function getFeeParamsByTransactionType(
   }
 }
 
+let baseFeeCache = {
+  destinationAddress: '',
+  l1Gas: '0',
+  l1GasPrice: '10000000000'
+}
+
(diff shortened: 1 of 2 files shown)


src/ethereum/fees/ethMiningFees.ts
```diff
@@ -329,6 +329,12 @@ export async function getFeeParamsByTransactionType(
   }
 }
 
+let baseFeeCache = {
+  destinationAddress: '',
+  l1Gas: '0',
+  l1GasPrice: '10000000000'
+}
+
```
swansontec, line 337: I don't really like global variables like this. I think I see what the the problem is - you are trying to keep the fee stable as you loop around the max spend code?

I wonder if there is another way to do this. I am approving the PR for now because this is the right approach at the 11'th hour, but maybe we should create a follow-on task to clean things up.

swansontec (approved):
I'm not super happy, but I think I see how it works.

PR #732, opened 2024-02-23
Enable eip1559 on zksync

peachbits:
---
  - https://app.asana.com/0/0/1206673671805801


samholmes (commented):
Hold on this change and investigate the source issue more before concluding that this is the fix.

samholmes:
Closing this PR because the real fix is: https://github.com/EdgeApp/edge-currency-accountbased/pull/742

PR #738, opened 2024-03-05
Replace deprecated rpc methods

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206671189992541

### DIFF
--- a/src/solana/SolanaEngine.ts
+++ b/src/solana/SolanaEngine.ts
@@ -4,7 +4,7 @@ import {
   getAssociatedTokenAddress
 } from '@solana/spl-token'
 import * as solanaWeb3 from '@solana/web3.js'
-import { add, eq, gt, gte, lt, mul, sub } from 'biggystring'
+import { add, eq, gt, gte, lt, max, mul, sub } from 'biggystring'
 import { asMaybe, asNumber, asString } from 'cleaners'
 import {
   EdgeCurrencyEngine,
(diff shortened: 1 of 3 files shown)


src/solana/SolanaEngine.ts
```diff
@@ -583,6 +598,36 @@ export class SolanaEngine extends CurrencyEngine<
...
+        microLamports = max('0', mul(this.priorityFee, '0.75'))
+        break
+      case 'standard':
+        microLamports = this.priorityFee
+        break
+      case 'high':
+        microLamports = max('1', mul(this.priorityFee, '1.25'))
+        break
+      case 'custom': {
+        if (customNetworkFee == null) {
```
swansontec, line 615: Why not use a cleaner for the custom fee, instead of opening it up by hand?

src/solana/SolanaEngine.ts
```diff
@@ -252,11 +255,12 @@ export class SolanaEngine extends CurrencyEngine<
...
     try {
-      const response = await this.fetchRpc('getRecentBlockhash')
-      const {
-        feeCalculator: { lamportsPerSignature }
-      } = asRecentBlockHash(response).value
-      this.feePerSignature = lamportsPerSignature.toString()
+      const response = await this.fetchRpc('getRecentPrioritizationFees')
+      const recentPriorityFees = asGetRecentPrioritizationFees(response)
+      const latestPriorityFee = recentPriorityFees.sort(
+        (a, b) => a.slot - b.slot
```
swansontec, line 261: This could crash if the array is empty, but since we are in a try block anyhow, I guess it doesn’t matter. Might warrant a comment, though.

swansontec (approved):
Two optional changes.

PR #743, opened 2024-03-07
Refactor FIO processTransaction for unit testing

paullinator:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206771367992292

### DIFF
--- a/src/fio/FioEngine.ts
+++ b/src/fio/FioEngine.ts
@@ -13,6 +13,7 @@ import { asMaybe, asString, asTuple } from 'cleaners'
 import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
+  EdgeDenomination,
   EdgeEnginePrivateKeyOptions,
   EdgeFetchFunction,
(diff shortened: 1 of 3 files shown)


src/fio/FioEngine.ts
```diff
@@ -384,212 +377,38 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
...
-    if (action.block_num <= this.otherData.highestTxHeight) {
-      return action.block_num
+    const denom = getDenomination(
+      currencyCode,
+      this.currencyInfo,
+      this.allTokensMap
+    )
+    if (denom == null) {
+      this.error(`Received unsupported currencyCode: ${currencyCode}`)
+      return 0
```
peachbits, line 387: Unrelated to your changes and just something I noticed, but if FIO ever adds tokens this routine will break early because this number is a block height and it's compared against a local saved value.

peachbits (approved):

PR #745, opened 2024-03-15
Add Cardano

peachbits:
---
  - https://app.asana.com/0/0/1206574606763075

### DIFF
--- a/src/cardano/CardanoEngine.ts
+++ b/src/cardano/CardanoEngine.ts
@@ -1,4 +1,6 @@
-import { add, sub } from 'biggystring'
+import * as Cardano from '@emurgo/cardano-serialization-lib-nodejs'
+import { add, gt, mul, sub } from 'biggystring'
+import { asMaybe, asString, asTuple } from 'cleaners'
 import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
@@ -8,21 +10,30 @@ import {
   EdgeTokenId,
   EdgeTransaction,
   EdgeWalletInfo,
-  JsonObject
+  InsufficientFundsError,
+  JsonObject,
+  NoAmountSpecifiedError
 } from 'edge-core-js/types'
+import { base16 } from 'rfc4648'
 
 import { CurrencyEngine } from '../common/CurrencyEngine'
 import { PluginEnvironment } from '../common/innerPlugin'
-import { getFetchCors } from '../common/utils'
+import { cleanTxLogs, getFetchCors } from '../common/utils'
 import { CardanoTools } from './CardanoTools'
 import {
+  asCardanoInitOptions,
+  asCardanoPrivateKeys,
+  asCardanoTxOtherParams,
   asCardanoWalletOtherData,
   asKoiosAddressTransactions,
   asKoiosBalance,
   asKoiosBlockheight,
+  asKoiosNetworkParameters,
   asKoiosTransactionsRes,
   asSafeCardanoWalletInfo,
+  CardanoInitOptions,
   CardanoNetworkInfo,
+  CardanoTxOtherParams,
   CardanoWalletOtherData,
   KoiosNetworkTx,
   SafeCardanoWalletInfo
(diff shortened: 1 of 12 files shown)


src/cardano/CardanoEngine.ts
```diff
@@ -0,0 +1,556 @@
...
+    // Add output to the tx
+    txBuilder.add_output(
+      Cardano.TransactionOutput.new(
+        outputAddr,
+        Cardano.Value.new(Cardano.BigNum.from_str(outputAmount))
+      )
+    )
+
+    // Filter out multi asset utxo to keep this simple
+    const lovelaceUtxos = this.utxos.filter(u => u.value !== '0')
```
swansontec, line 321: Could customers get stuck funds, if they import the wallet from another app?

src/cardano/CardanoEngine.ts
```diff
@@ -0,0 +1,556 @@
...
+        break
+      default:
+        throw new Error('Unknown networkId')
+    }
+
+    return txBody
+  }
+
+  async makeSpend(edgeSpendInfoIn: EdgeSpendInfo): Promise<EdgeTransaction> {
+    edgeSpendInfoIn = upgradeMemos(edgeSpendInfoIn, this.currencyInfo)
```
swansontec, line 361: Maybe we could drop the legacy memo support for new coins going forward? Dropping it for Ethereum or whatever would be a breaking change, but for new stuff, there's no legacy support to break.

src/cardano/CardanoEngine.ts
```diff
@@ -0,0 +1,556 @@
...
+  // Do any async initialization necessary for the engine
+  await engine.loadEngine()
+
+  return engine
+}
+
+export const processCardanoTransaction = (opts: {
+  currencyCode: string
+  publicKey: string
+  tx: ReturnType<typeof asKoiosTransactionsRes>[number]
```
swansontec, line 513: Should you define a `type KoiosNetworkTx = ReturnType<typeof asKoiosTransactionsRes>[number]` in the types file?

src/cardano/CardanoTools.ts
```diff
@@ -0,0 +1,191 @@
...
+    privateKey: Cardano.Bip32PrivateKey
+  } {
+    const mnemonicEntropy = mnemonicToEntropy(mnemonic)
+    const privateKey = Cardano.Bip32PrivateKey.from_bip39_entropy(
+      base16.parse(mnemonicEntropy),
+      new Uint8Array()
+    )
+
+    function harden(num: number): number {
+      return 0x80000000 + num
```
swansontec, line 50: I would do `0x80000000 | num` instead of `+`. It doesn't matter in this particular case, but it's just more correct.

src/cardano/CardanoTools.ts
```diff
@@ -0,0 +1,191 @@
...
+    const keys = asCardanoPrivateKeys(pluginId)(privateWalletInfo.keys)
+    return keys.mnemonic
+  }
+
+  async getDisplayPublicKey(publicWalletInfo: EdgeWalletInfo): Promise<string> {
+    const { keys } = asSafeCardanoWalletInfo(publicWalletInfo)
+    return keys.publicKey
+  }
+
+  async importPrivateKey(mnemonic: string): Promise<Object> {
```
swansontec, line 77: The return type should be `Promise<JsonObject>`.

src/cardano/CardanoTools.ts
```diff
@@ -0,0 +1,191 @@
...
+    // Validate input
+    Cardano.Bip32PrivateKey.from_bip39_entropy(
+      base16.parse(mnemonicEntropy),
+      new Uint8Array()
+    )
+
+    return { [`${pluginId}Mnemonic`]: mnemonic }
+  }
+
+  async createPrivateKey(walletType: string): Promise<Object> {
```
swansontec, line 92: The return type should be `Promise<JsonObject>`.

src/cardano/CardanoTools.ts
```diff
@@ -0,0 +1,191 @@
...
+    const type = walletType.replace('wallet:', '')
+
+    if (type === this.currencyInfo.pluginId) {
+      const entropy = Buffer.from(this.io.random(32)).toString('hex')
```
swansontec, line 96: Use `base16.stringify(this.io.random(32))`

src/cardano/CardanoTools.ts
```diff
@@ -0,0 +1,191 @@
...
+    Cardano.Bip32PrivateKey.from_bip39_entropy(
+      base16.parse(mnemonicEntropy),
+      new Uint8Array()
+    )
+
+    return { [`${pluginId}Mnemonic`]: mnemonic }
+  }
+
+  async createPrivateKey(walletType: string): Promise<Object> {
+    const type = walletType.replace('wallet:', '')
```
swansontec, line 93: Just put this check at the top, and then you can get rid of the string parsing and the extra indentation level:

```ts
if (walletType !== this.currencyInfo.walletType) {
 throw new Error('InvalidWalletType')
}
```

swansontec (approved):
Generally awesome PR! I have a few questions and some cleanup suggestions, but nothing major.

swansontec (approved):

src/cardano/CardanoTools.ts
swansontec: We can skip this one, since we want to match what the sample code is doing.

swansontec (approved):

PR #746, opened 2024-03-15
Update eslint config

peachbits:
---
  - https://app.asana.com/0/0/1206854385865126

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,4 @@
 /dist/
 /coverage/
 /lib/
+/scripts/cosmos-ts-protos/tmp/
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "clean": "rimraf dist lib android/src/main/assets/edge-currency-accountbased",
     "cli": "node -r sucrase/register cli/cli.ts",
     "fix": "eslint . --fix",
-    "lint": "eslint .",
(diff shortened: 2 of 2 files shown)


.eslintignore
```diff
@@ -2,3 +2,4 @@
 /dist/
 /coverage/
 /lib/
+/scripts/
```
swansontec, line 5: We definitely want linting in here, since this is code we write ourselves. Yes, these are not part of the final build, but the code should still be of high quality.

package.json
```diff
@@ -37,7 +37,7 @@
     "clean": "rimraf dist lib android/src/main/assets/edge-currency-accountbased",
     "cli": "node -r sucrase/register cli/cli.ts",
     "fix": "eslint . --fix",
-    "lint": "eslint .",
+    "lint": "NODE_OPTIONS=--max-old-space-size=8192 eslint .",
```
swansontec, line 40: Ah, this makes sense.

swansontec (changes requested):

.eslintignore
peachbits: I had ignored the whole folder because it wasn't in the build but that's a good point. Only ignoring the tmp folder now

swansontec (approved):

PR #747, opened 2024-03-19
Upgrade edge-core-js to v1.14.0

peachbits:
---
  - https://app.asana.com/0/0/1206872405547651

### DIFF
--- a/src/algorand/AlgorandEngine.ts
+++ b/src/algorand/AlgorandEngine.ts
@@ -255,6 +255,7 @@ export class AlgorandEngine extends CurrencyEngine<
     } = tx
 
     let currencyCode: string
+    let tokenId: EdgeTokenId
     let nativeAmount: string
     let networkFee: string
     let parentNetworkFee: string | undefined
@@ -279,6 +280,7 @@ export class AlgorandEngine extends CurrencyEngine<
         }
 
         currencyCode = this.currencyInfo.currencyCode
+        tokenId = null
         break
       }
       case 'axfer': {
@@ -304,6 +306,7 @@ export class AlgorandEngine extends CurrencyEngine<
         const edgeToken: EdgeToken | undefined = this.allTokensMap[assetId]
         if (edgeToken == null) return
         currencyCode = edgeToken.currencyCode
+        tokenId = assetId.toString()
 
         break
       }
@@ -327,6 +330,7 @@ export class AlgorandEngine extends CurrencyEngine<
         }
 
         currencyCode = this.currencyInfo.currencyCode
+        tokenId = null
         break
       }
       default: {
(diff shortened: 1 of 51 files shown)


src/ethereum/networkAdapters/EvmScanAdapter.ts
```diff
@@ -349,7 +349,14 @@ export class EvmScanAdapter extends NetworkAdapter<EvmScanAdapterConfig> {
...
       this.ethEngine.walletLocalData.publicKey.toLowerCase()
     const tokenTx = currencyCode !== this.ethEngine.currencyInfo.currencyCode
-
+    const knownTokenId = Object.keys(this.ethEngine.allTokensMap).find(
+      tokenId =>
+        this.ethEngine.allTokensMap[tokenId].currencyCode === currencyCode
+    )
+    if (tokenTx && knownTokenId === undefined) {
+      throw new Error('Unknown token')
+    }
```
swansontec, line 358: This isn't really safe, because of the general currencyCode confusion. However, I see that we would need to change `GetTxsParams` from currencyCode to tokenId, which is a pretty deep refactor.

If will approve this for now, IF you create a follow-on task to actually make this change to `GetTxsParams`.

src/tron/TronEngine.ts
```diff
@@ -1575,7 +1587,7 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
     const { edgeSpendInfo, currencyCode } = super.makeSpendCheck(
       edgeSpendInfoIn
     )
-    const { memos = [] } = edgeSpendInfo
+    const { memos = [], tokenId } = edgeSpendInfo
```
swansontec, line 1590: We should probably refactor `makeSpendCheck` to return a `tokenId` as well.

This can also be a follow-up task.

swansontec (approved):
Approved with a question.

src/algorand/AlgorandEngine.ts
```diff
@@ -304,6 +306,7 @@ export class AlgorandEngine extends CurrencyEngine<
         const edgeToken: EdgeToken | undefined = this.allTokensMap[assetId]
         if (edgeToken == null) return
         currencyCode = edgeToken.currencyCode
+        tokenId = assetId.toString()
```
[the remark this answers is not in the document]
peachbits: yup

src/ethereum/networkAdapters/EvmScanAdapter.ts
peachbits: https://app.asana.com/0/0/1206891509258292/f

src/tron/TronEngine.ts
peachbits: https://app.asana.com/0/1200382638405084/1206891509258298/f

PR #749, opened 2024-03-19
Upgrade edge-core-js to v2.2.1

peachbits:
---
  - https://app.asana.com/0/0/1206872405547660


cli/commands/makeSpend.ts
```diff
@@ -17,7 +17,7 @@ export class MakeSpend extends Command<CurrencyContext> {
     const { engine } = await getCliEngine(this.context)
 
     const spend: EdgeSpendInfo = {
-      tokenId,
+      tokenId: tokenId ?? null,
```
swansontec, line 20: Can we do `const {address, amount,tokenId = null} = this` up above? We don't want the `undefined` floating around - we should kill it right away.

swansontec (commented):
We should add one more commit, to search and replace `tokenId ?? null` with just `tokenId`, since the v2 core should no longer pass `undefined` for tokenId's.

src/ripple/RippleEngine.ts
```diff
@@ -895,18 +846,19 @@ export class XrpEngine extends CurrencyEngine<
...
+        throw new InsufficientFundsError({ tokenId })
       const parentBalance =
         this.walletLocalData.totalBalances[parentCurrencyCode] ?? '0'
 
       if (gt(parentNetworkFee, parentBalance)) {
         throw new InsufficientFundsError({
-          currencyCode: parentCurrencyCode,
-          networkFee: parentNetworkFee
+          networkFee: parentNetworkFee,
+          tokenId
```
swansontec, line 861: This should also be `tokenId: null`, because it's for the parent currency.

src/ripple/RippleEngine.ts
```diff
@@ -1092,31 +1045,36 @@ export class XrpEngine extends CurrencyEngine<
...
-    paymentTokenId,
-    paymentWallet
+    paymentInfo
   }: EdgeEngineActivationOptions): Promise<EdgeActivationQuote> => {
     if (activateTokenIds == null)
       throw new Error(
         `Must specify activateTokenIds for ${this.currencyInfo.currencyCode}`
       )
-    if (paymentTokenId != null)
+    const { wallet, tokenId } = paymentInfo ?? {}
```
swansontec, line 1054: Do `paymentInfo ?? { tokenId: null}` so tokenId can't be `undefined`

src/solana/SolanaEngine.ts
```diff
@@ -532,31 +532,31 @@ export class SolanaEngine extends CurrencyEngine<
...
-        currencyCode: this.currencyInfo.currencyCode
+        tokenId: null
       })
       const solRequired = sub(solBalance, edgeTx.networkFee)
       if (lt(sub(solRequired, this.minimumAddressBalance), '0')) {
         throw new InsufficientFundsError({
-          currencyCode: this.currencyInfo.currencyCode,
-          networkFee: this.feePerSignature
+          networkFee: this.feePerSignature,
+          tokenId
```
swansontec, line 554: This one should be `tokenId: null`, because we are out of the parent currency.

src/solana/SolanaEngine.ts
```diff
@@ -682,15 +682,15 @@ export class SolanaEngine extends CurrencyEngine<
...
       }
 
       const balanceSol =
         this.walletLocalData.totalBalances[this.chainCode] ?? '0'
       if (gt(add(parentNetworkFee, this.minimumAddressBalance), balanceSol)) {
         throw new InsufficientFundsError({
-          currencyCode: this.chainCode,
-          networkFee: parentNetworkFee
+          networkFee: parentNetworkFee,
+          tokenId
```
swansontec, line 693: Also `tokenId: null` here, because the parent currency is the low.

src/tron/TronEngine.ts
```diff
@@ -1422,14 +1365,8 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
 
     const unstakeNativeAmount = sub(claimedAmount, networkFee)
     const edgeTransaction: EdgeTransaction = {
-      action: {
-        type: 'unstake',
-        stakeAssets: [
-          {
-            pluginId: this.currencyInfo.pluginId,
-            nativeAmount: unstakeNativeAmount
-          }
```
swansontec, line 1420: We split `action` into `assetAction` and `chainAction`. The `chainAction` represents the total effect of this transaction, independent of which tokenId is being queried, while `assetAction` would be different if you are querying one asset vs another.

So for a swap, querying the "to" asset or the "from" asset in "getTransactions" would return different `assetAction` values, but both transactions would have the same `chainAction` with detains about the swap in general.

So, once we fully adopt this new architecture, a lot of this deleted code will end up coming back. I'm not saying we need to change anything now, but just a heads up for the near future.

src/tron/TronEngine.ts
```diff
@@ -1490,9 +1427,13 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
 
   async getMaxSpendable(spendInfo: EdgeSpendInfo): Promise<string> {
     spendInfo = upgradeMemos(spendInfo, this.currencyInfo)
-    const { memos = [] } = spendInfo
+    const { memos = [], tokenId } = spendInfo
+    const currencyCode =
+      tokenId === null
+        ? this.currencyInfo.currencyCode
+        : this.allTokensMap[tokenId]
```
swansontec, line 1434: I see this block in a bunch of places, and it seems like it will return `string | EdgeToken`? Did you mean `const {currencyCode} = tokenId == null ? this.currencyInfo : this.allTokensMap[tokenId]`, to actually return a `string` in all cases?

src/tron/TronEngine.ts
```diff
@@ -1662,8 +1603,8 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
       this.walletLocalData.totalBalances[this.currencyInfo.currencyCode] ?? '0'
     if (gt(transactionCostSUN, balanceSUN)) {
       throw new InsufficientFundsError({
-        currencyCode: this.currencyInfo.currencyCode,
-        networkFee: totalFeeSUN
+        networkFee: totalFeeSUN,
+        tokenId: edgeSpendInfoIn.tokenId
```
swansontec, line 1607: This should be `tokenId: null`

swansontec (changes requested):
I found a few things that we need to fix here.

src/common/uriHelpers.ts
```diff
@@ -16,6 +17,7 @@ export function parseUriCommon(
   currencyInfo: EdgeCurrencyInfo,
   uri: string,
   networks: { [network: string]: boolean },
+  builtinTokens: EdgeTokenMap,
```
swansontec, line 20: Nitpick: In the `getLegacyDenomination`, you put the `builtinTokens` last, but here you put it in the middle. Maybe it should go at the end here too, to match what `getLegacyDenomination` does?

swansontec (approved):
This cleaned up nicely. Go ahead and squash down the three fixups (including "Fix tokenId ?? null"), so it's just one commit.

You can then either merge it, or you can fix the one nit-picky thing I found.

src/ripple/RippleEngine.ts
```diff
@@ -205,6 +191,7 @@ export class XrpEngine extends CurrencyEngine<
             },
             ourReceiveAddresses: [],
             signedTx: '',
+            tokenId: fromTokenId ?? null,
```
swansontec, line 194: This one is still `?? null`

cli/commands/getBalance.ts
```diff
@@ -22,7 +22,7 @@ export class GetBalance extends Command<CurrencyContext> {
     const allTokens = { ...builtinTokens, ...customTokens }
     const { currencyCode } = tokenId == null ? currencyInfo : allTokens[tokenId]
 
-    const nativeBalance = await engine.getBalance({ currencyCode })
+    const nativeBalance = await engine.getBalance({ tokenId: tokenId ?? null })
```
swansontec, line 25: You can do `const { tokenId = null } = this` up above to get rid of the `?? null` here

cli/commands/makeSpend.ts
swansontec: This one is still here.

PR #751, opened 2024-03-21
Add EIP-4844 fee calculation

peachbits:
---
  - https://app.asana.com/0/0/1206844514368991

### DIFF
--- a/src/ethereum/fees/ethMiningFees.ts
+++ b/src/ethereum/fees/ethMiningFees.ts
@@ -1,6 +1,6 @@
 import { Common } from '@ethereumjs/common'
 import { TransactionFactory } from '@ethereumjs/tx'
-import { add, ceil, div, gte, lt, lte, mul, sub } from 'biggystring'
+import { add, div, gte, lt, lte, mul, sub } from 'biggystring'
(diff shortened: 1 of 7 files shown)


src/ethereum/fees/ethMiningFees.ts
```diff
@@ -253,23 +256,26 @@ export const calcOptimismRollupFees = (
...
-      rawTxCost += 16 // cost for non-zero byte
+      txCompressedSize += 16 // cost for non-zero byte
     }
   }
 
-  const gasUsed = add(
-    add(rawTxCost.toString(), fixedOverhead),
-    MAX_SIGNATURE_COST
-  )
+  txCompressedSize = (txCompressedSize + MAX_SIGNATURE_COST) / 16
```
swansontec, line 268: So the spec postpones the division to the very end, but I see that we are using biggystring's floating-point support to shuffle this earlier, while still getting the same result. I guess that works.

swansontec (approved):

src/ethereum/fees/ethMiningFees.ts
peachbits: I got it mixed up because I used two sources which had the math different as you described. I'll make it match the commented source.

PR #757, opened 2024-04-05
Unit-test the `validateMemos` function

swansontec:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Description

We had reports that `makeSpend` would succeed with "0x" in our memos, but that the transactions would fail on-chain. Well, no, `makeSpend` won't actually allow this situation, at least in the current release.

---
  - https://app.asana.com/0/0/1206985639130703

### DIFF
--- a/test/utils/validateMemos.test.ts
+++ b/test/utils/validateMemos.test.ts
@@ -1,0 +1,149 @@
+import { expect } from 'chai'
+import { EdgeCurrencyInfo, EdgeMemo } from 'edge-core-js'
+import { describe, it } from 'mocha'
+
+import { validateMemos } from '../../src/common/validateMemos'
+
+describe('validateMemos', function () {
+  function callValidateMemos(
+    memos: EdgeMemo[],
+    extraInfo: Partial<EdgeCurrencyInfo>
+  ): void {
+    return validateMemos(
+      {
+        spendTargets: [],
+        tokenId: null,
+        memos
+      },
+      {
+        addressExplorer: '',
+        currencyCode: 'MTC',
+        denominations: [],
+        displayName: 'Memo Test Coin',
+        pluginId: 'memoTest',
+        transactionExplorer: '',
+        walletType: 'wallet:memoTest',
+        ...extraInfo
+      }
+    )
+  }
+
+  it('rejects multiple memos', function () {
+    expect(() =>
+      callValidateMemos(
+        [
+          { type: 'hex', value: '' },
+          { type: 'hex', value: '' }
+        ],
+        { memoOptions: [{ type: 'hex' }] }
+      )
+    ).to.throw('Memo Test Coin only supports one memo')
+  })
+
+  it('accepts multiple memos', function () {
+    callValidateMemos(
+      [
+        { type: 'text', value: '' },
+        { type: 'hex', value: '' }
+      ],
+      {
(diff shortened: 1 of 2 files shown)


test/utils/validateMemos.test.ts
```diff
@@ -0,0 +1,135 @@
...
+          type: 'hex',
+          memoName: 'data',
+          minBytes: 2,
+          maxBytes: 4
+        }
+      ]
+    }
+
+    // Good:
+    callValidateMemos([{ type: 'hex', value: '12345678' }], extraInfo)
```
samholmes, line 69: add happy cases for `1234` and `123456`

test/utils/validateMemos.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    // Bad:
+    expect(() =>
+      callValidateMemos([{ type: 'number', value: '257' }], extraInfo)
+    ).throws('Memo Test Coin tag: cannot be greater than 256')
+    expect(() =>
+      callValidateMemos([{ type: 'number', value: '1b' }], extraInfo)
+    ).throws('Memo Test Coin tag: is not a valid number')
+    expect(() =>
+      callValidateMemos([{ type: 'hex', value: '12345678' }], extraInfo)
+    ).throws('Memo Test Coin tag: cannot be type hex')
```
samholmes, line 110: Are negatives okay? Are decimals okay?

test/utils/validateMemos.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        {
+          type: 'text',
+          memoName: 'message',
+          maxLength: 6
+        }
+      ]
+    }
+
+    // Good:
+    callValidateMemos([{ type: 'text', value: 'hello' }], extraInfo)
```
samholmes, line 125: Are empty strings okay?

samholmes, line 125: I'd also add a case for a `hex` here. That way we can catch when implementation confuses strings with hex. I'm thinking `"0x123"` and `cbdc` as a couple of cases that are valid hex and also valid strings

samholmes (changes requested):

test/utils/validateMemos.test.ts
```diff
@@ -122,12 +130,16 @@ describe('validateMemos', function () {
...
+    callValidateMemos([{ type: 'text', value: '' }], extraInfo)
     callValidateMemos([{ type: 'text', value: 'hello' }], extraInfo)
 
     // Bad:
     expect(() =>
       callValidateMemos([{ type: 'text', value: 'hello!!' }], extraInfo)
     ).throws('Memo Test Coin message: cannot be longer than 6 characters')
+    expect(() =>
+      callValidateMemos([{ type: 'hex', value: '1234' }], extraInfo)
+    ).throws('Memo Test Coin message: cannot be type hex')
```
samholmes, line 142: What about `callValidateMemos([{ type: 'text', value: '1234' }], extraInfo)`? That's more what I meant. Having a case where the data is hex but the type is specified as `'text'` could be good incase implementation incorrectly assumes because the data looks like hex that it will be assumed a hex type. Better regression coverage me thinks

samholmes (approved):
One more change, but otherwise approved.

PR #758, opened 2024-04-05
Test processEvmScanTransaction

samholmes:
### CHANGELOG

- [ ] Yes
- [x] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206966606672147

### DIFF
--- a/test/ethereum/network/evmScanTxProcessing.cases.ts
+++ b/test/ethereum/network/evmScanTxProcessing.cases.ts
@@ -1,0 +1,269 @@
+import { EdgeTransaction } from 'edge-core-js'
+
+import { TransactionProcessingContext } from '../../../src/ethereum/networkAdapters/EvmScanAdapter'
+import { allTokensMapFixture } from './allTokensMapFixture'
+import { currencyInfoFixture } from './currencyInfoFixture'
+
+export interface TestCase {
+  input: {
+    context: TransactionProcessingContext
+    rawTx: unknown
+    l1RollupFee: string
+  }
+  output: EdgeTransaction
+}
+
+export const testCases = [
+  {
+    input: {
+      context: {
+        allTokensMap: allTokensMapFixture,
+        currencyInfo: currencyInfoFixture,
+        forWhichAddress: '0x036639F209f2Ebcde65a3f7896d05a4941d20373',
+        forWhichCurrencyCode: 'ETH',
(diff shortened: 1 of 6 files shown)


test/ethereum/network/evmScanTxProcessing.cases.ts
```diff
@@ -0,0 +1,273 @@
+import { EdgeTransaction } from 'edge-core-js'
+
+import { TransactionProcessingContext } from '../../../src/ethereum/networkAdapters/EvmScanAdapter'
+import { allTokensMapFixture } from './allTokensMapFixture'
+import { currencyInfoFixture } from './currencyInfoFixture'
```
swansontec, line 5: Can we use the actual Ethereum info, instead of copy-pasting everything? Alternatively, can we trim down the fixtures cases to just the tokens / currency info we care about?

test/ethereum/network/evmScanTxProcessing.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import { describe, it } from 'mocha'
+
+import {
+  asEvmScanInternalTransaction,
+  asEvmScanTransaction
+} from '../../../src/ethereum/ethereumTypes'
+import { processEvmScanTransaction } from '../../../src/ethereum/networkAdapters/EvmScanAdapter'
+import { testCases as processEvmScanTransactionTestCases } from './evmScanTxProcessing.cases'
+
+describe(`processEvmScanTransaction`, function () {
```
swansontec, line 12: Can we put the test cases in the same file as the logic? I don't think separating them makes the code more readable.

test/ethereum/network/evmScanTxProcessing.cases.ts
```diff
@@ -0,0 +1,273 @@
...
+  output: Output
+}
+
+interface Input {
+  context: TransactionProcessingContext
+  rawTx: unknown
+  l1RollupFee: string
+}
+
+type Output = EdgeTransaction
```
swansontec, line 18: Can we flatten this a bit:

```ts
export interface TestCase {
  input: {
    context: TransactionProcessingContext
    rawTx: unknown
    l1RollupFee: string
  }
  output: EdgeTransaction
}
```

swansontec (approved):
Approved, but I feel like the unit tests need to be tightened up so we can maintain them going forward.

PR #764, opened 2024-04-19
Hedera - Add auto activation

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206745835636212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- added: (Hedera) Replace account purchase with auto account creation
 - added: Ethereum default tokens AXS, COTI, and GMT
 - added: Binance Smart Chain default token AAVE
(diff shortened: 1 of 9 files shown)


swansontec (approved):
The code looks great.

Does this work correctly with our weird derivation path? It looks like the derivation path just impacts how we get the public key, so all the auto-activation stuff happens on the other side of this problem.

peachbits:
> The code looks great.
> 
> Does this work correctly with our weird derivation path? It looks like the derivation path just impacts how we get the public key, and all the auto-activation stuff happens after this point.

Yes, the issue is the direvation path from a mnemonic to the private key changed between sdk versions. The method `toEd25519PrivateKey` is marked deprecated but is the same as the previous sdk derivation method

PR #765, opened 2024-04-24
Merge regular and internal EvmScanAdapter transaction queries

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207126281144803

### DIFF
--- a/src/ethereum/networkAdapters/EvmScanAdapter.ts
+++ b/src/ethereum/networkAdapters/EvmScanAdapter.ts
@@ -210,10 +210,10 @@ export class EvmScanAdapter extends NetworkAdapter<EvmScanAdapterConfig> {
         { searchRegularTxs: false }
       )
       server = txsRegularResp.server ?? txsInternalResp.server ?? ''
-      allTransactions = [
+      allTransactions = mergeEdgeTransactions([
         ...txsRegularResp.allTransactions,
         ...txsInternalResp.allTransactions
-      ]
+      ])
     } else {
       const tokenInfo = this.ethEngine.getTokenInfo(currencyCode)
       if (tokenInfo != null && typeof tokenInfo.contractAddress === 'string') {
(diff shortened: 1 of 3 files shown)


src/ethereum/networkAdapters/EvmScanAdapter.ts
```diff
@@ -486,3 +486,47 @@ export function processEvmScanTransaction(
   return edgeTransaction
   // or should be this.addTransaction(currencyCode, edgeTransaction)?
 }
+
+export function mergeEdgeTransactions(
+  transactions: EdgeTransaction[]
+): EdgeTransaction[] {
+  // The uniqueKey is txid and tokenId concatenated
+  const txidToTransaction: { [uniqueKey: string]: EdgeTransaction } = {}
```
swansontec, line 494: Could you use a `Map<string, EdgeTransaction>` here?

test/ethereum/network/evmScanTxMerging.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      networkFee: '42',
+      ourReceiveAddresses: [bobAddress]
+    }
+    const internalTx = {
+      ...templateTx,
+      isSend: true,
+      nativeAmount: '100',
+      networkFee: '0'
+    }
+    assert.throws(() => mergeEdgeTransactions([regularTx, internalTx]))
```
swansontec, line 130: `expect(()=> mergeEdgeTransactions(...)).throws("Failed to merge transaction 0xabcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789: Mismatch parentNetworkFee")`

The `expect` syntax is generally nicer, so we prefer it in most cases.

In this case, the `expect` syntax lets us say what error message we want. This is important, because it means the test will fail if the function starts throwing a `TypeError` or some other error besides the one we want.

swansontec (changes requested):
We should be using a `Map` instead of a JS object, since this is safer.

I also have a tiny suggestion for the unit tests. I am sooo happy to see unit tests, though! Thank you thank you thank you.

swansontec (approved):

PR #768, opened 2024-04-26
Algorand, Cosmos, Solana, XRP, and Tron - Add bulk balance fetching and onNewTokens callback 

peachbits:
---
  - https://app.asana.com/0/0/1206096604395365

### DIFF
--- a/src/cosmos/CosmosEngine.ts
+++ b/src/cosmos/CosmosEngine.ts
@@ -22,7 +22,7 @@ import {
 } from '@cosmjs/stargate'
 import { longify } from '@cosmjs/stargate/build/queryclient'
 import { fromRfc3339WithNanoseconds, toSeconds } from '@cosmjs/tendermint-rpc'
-import { add, ceil, lt, mul, sub } from 'biggystring'
+import { add, ceil, gt, lt, mul, sub } from 'biggystring'
 import {
   AuthInfo,
   Fee,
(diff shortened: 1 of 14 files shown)


src/cosmos/CosmosEngine.ts
```diff
@@ -410,14 +410,24 @@ export class CosmosEngine extends CurrencyEngine<
...
-      this.enabledTokenIds.forEach(tokenId => {
+      const detectedTokenIds: string[] = []
+      Object.keys(this.allTokensMap).forEach(tokenId => {
         const token = this.allTokensMap[tokenId]
         const tokenBal = balances.find(
           bal => bal.denom === token.networkLocation?.contractAddress
         )
-        this.updateBalance(token.currencyCode, tokenBal?.amount ?? '0')
+        const balance = tokenBal?.amount ?? '0'
+        this.updateBalance(token.currencyCode, balance)
```
swansontec, line 420: Would we want to guard the `updateBalance` call with `this.enabledTokenIds.includes(tokenId)`, or does it not matter?

swansontec (approved):

src/cosmos/CosmosEngine.ts
peachbits: As we discussed IRL, updateBalance doesn't fire the expensive callback unless the balance changes so there's it's not necessary to guard it.

PR #773, opened 2024-05-10
Fix DOT max spending caused by incorrectly checking `spendInfo.tokenId`

samholmes:
It looks like this bug slipped in during the edge-core-js v2 upgrade.
The logic used to compare spend info currency code inequality with
engine currency code.
This change is the correct condition for the refactor.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207258762926303

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fixed: (DOT) Fix incorrect insufficient funds error for max spend of native DOT currency in Polkadot engine
+
 ## 4.3.2 (2024-05-09)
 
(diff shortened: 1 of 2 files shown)


swansontec (approved):

swansontec:
Specifically commit 4931fb55f5cedd58ee7c3187ab81a7df1eec35cb was the source of this bug. The logic changed from `currencyCode !== currencyInfo.currencyCode` (so it's a token) to `token == null` (so it's not a token).

Yup, that would break it.

PR #777, opened 2024-05-29
Upgrade Solana dependencies

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207288905528221


swansontec (approved):
This all looks reasonable, but I'm not sure about merging this so close to the release. If Solana is completely dead, then I guess we could take the risk.

PR #780, opened 2024-05-30
Remove Thorchain synth assets

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207355978231281

### DIFF
--- a/src/cosmos/info/thorchainruneInfo.ts
+++ b/src/cosmos/info/thorchainruneInfo.ts
@@ -1,77 +1,9 @@
-import { EdgeCurrencyInfo, EdgeTokenMap } from 'edge-core-js/types'
+import { EdgeCurrencyInfo } from 'edge-core-js/types'
 
 import { makeOuterPlugin } from '../../common/innerPlugin'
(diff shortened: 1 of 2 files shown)


src/cosmos/info/thorchainruneInfo.ts
```diff
@@ -98,7 +29,6 @@ const networkInfo: CosmosNetworkInfo = {
 
 const currencyInfo: EdgeCurrencyInfo = {
   currencyCode: 'RUNE',
-  customTokenTemplate: cosmosCustomTokenTemplate,
```
swansontec, line 101: I would comment this one out instead of deleting it. There are a few other networks where we don't have tokens, even though the broader family supports it. In all those cases, I left this line as a comment.

swansontec (approved):

src/cosmos/info/thorchainruneInfo.ts
peachbits: good point, re-added as comment

PR #784, opened 2024-06-10
Allow Piratechain `derivePublicKey` to fail in case native code is missing

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207485024420879

### DIFF
--- a/src/piratechain/PiratechainTools.ts
+++ b/src/piratechain/PiratechainTools.ts
@@ -54,7 +54,7 @@ export class PiratechainTools implements EdgeCurrencyTools {
   ): Promise<string> {
     const { pluginId } = this.currencyInfo
     const keys = asPiratechainPrivateKeys(pluginId)(privateWalletInfo.keys)
-    return keys.mnemonic
+    return `Seed Phrase:\n${keys.mnemonic}\n\nBirthday Height:\n${keys.birthdayHeight}`
   }
 
   async getDisplayPublicKey(publicWalletInfo: EdgeWalletInfo): Promise<string> {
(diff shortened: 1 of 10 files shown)


src/zcash/zcashTypes.ts
```diff
@@ -82,13 +82,8 @@ export type ZcashBalances = Omit<
   'totalZatoshi'
 >
 
-export const asZecPublicKey = asObject({
-  birthdayHeight: asNumber,
```
swansontec, line 87: The reason we included the birthday height in the public keys is so we could potentially, in the future, have watch-only Zcash wallets. The SDK's don't presently support this use-case, but perhaps they could in the future. I guess we could put back the birthday height if / when that happens.

Otherwise, we could leave the birthday height here, but put an `asOptional` around it so we don't blow up. Another option is to return the birthday height from `derivePublicKey`, but simply fail to mention it in the cleaner. Let me know which direction we want to go with.

src/piratechain/PiratechainTools.ts
```diff
@@ -131,11 +130,18 @@ export class PiratechainTools implements EdgeCurrencyTools {
...
+
+    let unifiedViewingKey: string
+    try {
+      unifiedViewingKey = await this.nativeTools.deriveViewingKey(
         mnemonic,
         this.networkInfo.rpcNode.networkName
       )
+    } catch (e) {
+      // Return a failsafe value if derivation fails (e.g., native code for platform not implemented)
+      unifiedViewingKey = 'failed_to_call_deriveViewingKey'
```
swansontec, line 142: In the previous commit, I talked about leaving the birthday height in the public keys, but making it `asOptional`. What if we do the same thing to the `publicKey`? That way we can return `undefined` here, instead of some weird string "failed_to_call_deriveViewingKey".

swansontec (changes requested):

swansontec (approved):

PR #785, opened 2024-06-17
Solana: Allow unsignedTx passthrough to makeSpend

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207570211348323

### DIFF
--- a/src/solana/SolanaEngine.ts
+++ b/src/solana/SolanaEngine.ts
@@ -42,7 +42,8 @@ import {
   asyncWaterfall,
   cleanTxLogs,
   getFetchCors,
-  getOtherParams
+  getOtherParams,
+  promiseAny
 } from '../common/utils'
 import { SolanaTools } from './SolanaTools'
 import {
@@ -822,10 +823,10 @@ export class SolanaEngine extends CurrencyEngine<
     if (edgeTransaction.signedTx == null) throw new Error('Missing signedTx')
 
     try {
-      const funcs = this.tools.connections.map(connection => async () => {
+      const promises = this.tools.connections.map(async connection => {
         return await connection.sendEncodedTransaction(edgeTransaction.signedTx)
       })
-      const txid: TransactionSignature = await asyncWaterfall(funcs)
+      const txid: TransactionSignature = await promiseAny(promises)
       edgeTransaction.txid = txid
       edgeTransaction.date = Date.now() / 1000
       this.warn(`SUCCESS broadcastTx\n${cleanTxLogs(edgeTransaction)}`)
--- a/src/solana/SolanaEngine.ts
+++ b/src/solana/SolanaEngine.ts
@@ -5,7 +5,6 @@ import {
 } from '@solana/spl-token'
 import {
   AccountInfo,
-  BlockhashWithExpiryBlockHeight,
   ComputeBudgetProgram,
   ConfirmedSignatureInfo,
   Keypair,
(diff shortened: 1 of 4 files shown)


src/solana/SolanaEngine.ts
```diff
@@ -822,10 +823,10 @@ export class SolanaEngine extends CurrencyEngine<
     if (edgeTransaction.signedTx == null) throw new Error('Missing signedTx')
 
     try {
-      const funcs = this.tools.connections.map(connection => async () => {
+      const funcs = this.tools.connections.map(async connection => {
```
swansontec, line 853: These aren't functions anymore - they are promises. Update the name?

src/solana/SolanaEngine.ts
```diff
@@ -746,6 +765,7 @@ export class SolanaEngine extends CurrencyEngine<
       instructions.push(memoInstruction)
     }
 
+    if (this.recentBlockhash === '') await this.queryBlockhash()
```
swansontec, line 770: It seems like we are only going to load the block hash once at boot now. Wouldn't that make it even more stale?

Based on the commit message, I would have expected `startEngine` to ignore the blockhash entirely, and then `signTx` to call it anew for each signature, perhaps with a `Date.now()` timestamp so we can skip the query if we have a value from the last X seconds ago.

src/solana/solanaTypes.ts
```diff
@@ -113,3 +113,7 @@ export type SolanaInitOptions = ReturnType<typeof asSolanaInitOptions>
 export const asSolanaCustomFee = asObject({
   microLamports: asString
 })
+
+export const asSolanaSpendInfoOtherParams = asObject({
+  unsignedTx: asOptional(asString)
```
swansontec, line 118: This should be `asBase64`, so we document what it is an also validate that.

src/solana/SolanaEngine.ts
```diff
@@ -747,10 +747,14 @@ export class SolanaEngine extends CurrencyEngine<
...
 
-    const otherParams: JsonObject = {
-      unsignedSerializedSolTx: solTx.serialize({
+    const unsignedSerializedSolTx = base64.stringify(
+      solTx.serialize({
         requireAllSignatures: false
       })
+    )
+
+    const otherParams: JsonObject = {
```
swansontec, line 781: Does Solana use a cleaner for `otherParams`? That cleaner should be responsible for serializing / unserializing the payload using `asBase64`. Any time we have raw `base64.parse` or `base64.serialize` calls, that's a big red flag that suggests we aren't cleaning stuff properly.

swansontec (changes requested):
This makes sense overall, but we should be more careful with the cleaners.

src/solana/SolanaEngine.ts
```diff
@@ -775,13 +775,15 @@ export class SolanaEngine extends CurrencyEngine<
...
     const versionedTx = new VersionedTransaction(versionedMessage)
 
-    const cleanSpendInfoOtherParams =
-      asSolanaSpendInfoOtherParams(spendInfoOtherParams)
+    let unsignedTx = versionedTx.serialize()
+    if (spendInfoOtherParams?.unsignedTx != null) {
+      const cleanSpendInfoOtherParams =
+        asSolanaTxOtherParams(spendInfoOtherParams)
+      unsignedTx = cleanSpendInfoOtherParams.unsignedTx
+    }
```
swansontec, line 783: We need to do this:

```ts
const asSolanaSpendInfoOtherParams = asObject({
  unsignedTx: asOptional(asBase64)
})

export const asSolanaTxOtherParams = asObject({
  unsignedTx: asBase64
})
```

Then use `asSolanaSpendInfoOtherParams` here to clean the input from the GUI. Since the `unsignedTx` is optional, we can safely clean the input before we open the field. The `asSolanaTxOtherParams` has the `unsignedTx` as mandatory, since we need that to sign and send.

src/solana/SolanaEngine.ts
```diff
@@ -775,13 +775,15 @@ export class SolanaEngine extends CurrencyEngine<
...
 
     const otherParams: JsonObject = {
```
swansontec, line 785: Can we do `const otherParams = wasSolanaTxOtherParams({ unsignedTx })`, with `const wasSolanaTxOtherParams = uncleaner(asSolanaTxOtherParams)`

swansontec (changes requested):

swansontec (approved):

PR #788, opened 2024-06-19
Add `lightMode` support to Solana and EVM engines

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-core-js/pull/599

### Description

 none

---
  - https://app.asana.com/0/0/1207585261612937

### DIFF
--- a/src/algorand/AlgorandTools.ts
+++ b/src/algorand/AlgorandTools.ts
@@ -92,7 +92,7 @@ export class AlgorandTools implements EdgeCurrencyTools {
     const { pluginId } = this.currencyInfo
     const networks = { [pluginId]: true }
 
-    const { parsedUri, edgeParsedUri } = parseUriCommon({
+    const { parsedUri, edgeParsedUri } = await parseUriCommon({
       currencyInfo: this.currencyInfo,
       uri,
       networks,
--- a/src/algorand/AlgorandTools.ts
+++ b/src/algorand/AlgorandTools.ts
@@ -92,14 +92,14 @@ export class AlgorandTools implements EdgeCurrencyTools {
     const { pluginId } = this.currencyInfo
     const networks = { [pluginId]: true }
 
-    const { parsedUri, edgeParsedUri } = parseUriCommon(
-      this.currencyInfo,
+    const { parsedUri, edgeParsedUri } = parseUriCommon({
+      currencyInfo: this.currencyInfo,
       uri,
       networks,
-      this.builtinTokens,
-      currencyCode ?? this.currencyInfo.currencyCode,
+      builtinTokens: this.builtinTokens,
+      currencyCode: currencyCode ?? this.currencyInfo.currencyCode,
       customTokens
-    )
+    })
 
     let address = ''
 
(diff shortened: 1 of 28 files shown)


src/algorand/AlgorandTools.ts
```diff
@@ -91,14 +91,14 @@ export class AlgorandTools implements EdgeCurrencyTools {
...
-      this.currencyInfo,
-      uri,
-      networks,
-      this.builtinTokens,
-      currencyCode ?? this.currencyInfo.currencyCode,
-      customTokens
-    )
+    const { parsedUri, edgeParsedUri } = parseUriCommon({
+      currencyInfo: this.currencyInfo,
+      uri: uri,
```
swansontec, line 96: For this whole commit, do a search & replace for all `uri: uri,` to just `uri,`, and similarly for `network` and `customTokens`.

src/common/uriHelpers.ts
```diff
@@ -70,6 +79,7 @@ export function parseUriCommon(
   }
 
   const amountStr = parsedUri.query.amount
+  let currencyCode = currencyCodeArg
```
swansontec, line 85: Maybe move this up to the top, and make it `let {currencyCode} = opts`. Then we can drop the weird `currencyCodeArg` thing.

src/common/tokenHelpers.ts
```diff
@@ -35,8 +40,10 @@ export function makeMetaTokens(tokens: EdgeTokenMap): EdgeMetaToken[] {
 
 export const getTokenIdFromCurrencyCode = (
   currencyCode: string,
+  mainnetCurrencyCode: string,
   allTokensMap: EdgeTokenMap
-): string | undefined => {
+): EdgeTokenId | undefined => {
+  if (currencyCode === mainnetCurrencyCode) return null
```
swansontec, line 46: The ripple plugin uses `!= null` to deal with the return value, which is a big problem now that we have a separation between `null` and `undefined`. Each of the 4 places needs to be updated to `!== undefined` or whatever other logic would be needed.

src/solana/SolanaEngine.ts
```diff
@@ -502,7 +503,8 @@ export class SolanaEngine extends CurrencyEngine<
   // // Public methods
   // // ****************************************************************************
 
-  async startEngine(): Promise<void> {
+  async startEngine(opts?: StartEngineOpts): Promise<void> {
+    const { lightMode = false } = opts ?? {}
```
swansontec, line 507: Since we are moving `lightMode` to the `makeCurrencyEngine` call, this will become `this.lightMode` instead of being an option.

swansontec (changes requested):
A few tweaks & clean-ups.

src/common/tokenHelpers.ts
peachbits: I only found one spot to change. A few usages of != null should only allow strings through.

swansontec (approved):

PR #794, opened 2024-07-03
Reduce polling

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207721444734363


peachbits (changes requested):
I tested my own branch with the same changes (but also changed more timers than you did) and it caused worse performance.

peachbits (approved):

PR #800, opened 2024-07-12
Temporarily omit FIO tpid

Jon-edge:
Per a known bug on their chain, this is a temporary workaround.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207730139556448

### DIFF
--- a/src/fio/FioEngine.ts
+++ b/src/fio/FioEngine.ts
@@ -8,7 +8,7 @@ import {
 } from '@fioprotocol/fiosdk/lib/transactions/queries'
 import { Query } from '@fioprotocol/fiosdk/lib/transactions/queries/Query'
 import { Transactions } from '@fioprotocol/fiosdk/lib/transactions/Transactions'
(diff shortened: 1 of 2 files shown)


src/fio/FioEngine.ts
```diff
@@ -168,7 +168,10 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
   ) {
     super(env, tools, walletInfo, opts)
     this.fetchCors = getFetchCors(env.io)
-    this.tpid = tpid
+    // Removed tpid as a temporary fix for connecting wallets using bundled tx's
+    // until FIO fixes
+    // their chain
+    // this.tpid = tpid
```
peachbits, line 174: The network bug only affects zero balance wallets so we should still set it in the constructor but make it undefined in the balance loop if necessary

peachbits (changes requested):

peachbits (approved):

peachbits (approved):

PR #805, opened 2024-07-25
Downgrade FIO back to v1.8.0

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207873405885398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


peachbits (changes requested):
If the only issue this was meant to resolve is FIO name updates (new name, resync, fresh install, etc) then we should close in favor of https://github.com/EdgeApp/edge-currency-accountbased/pull/806

PR #811, opened 2024-07-31
Cosmos hardfork support

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207925866929800

### DIFF
--- a/src/cosmos/CosmosEngine.ts
+++ b/src/cosmos/CosmosEngine.ts
@@ -56,6 +56,7 @@ import { EdgeTokenId, MakeTxParams } from '../common/types'
 import { cleanTxLogs } from '../common/utils'
 import { CosmosTools } from './CosmosTools'
 import {
+  asChainIdUpdate,
   asCosmosPrivateKeys,
   asCosmosTxOtherParams,
   asCosmosWalletOtherData,
@@ -110,6 +111,7 @@ export class CosmosEngine extends CurrencyEngine<
   feeCache: Map<string, CosmosFee>
   stakedBalanceCache: string
   stakingSupported: boolean
+  chainId: string
 
   constructor(
     env: PluginEnvironment<CosmosNetworkInfo>,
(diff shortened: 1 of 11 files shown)


src/cosmos/CosmosEngine.ts
```diff
@@ -520,38 +520,50 @@ export class CosmosEngine extends CurrencyEngine<
...
+        const { newestTxid, lastTimestamp } = await this.queryTransactionsInner(
+          query,
+          clients
+        )
+        if (
+          newestTxid != null &&
+          this.otherData[query]?.newestTxid !== newestTxid
+        ) {
+          this.otherData[query] = { newestTxid }
+          this.otherData.archivedTxLastCheckTime = lastTimestamp
```
paullinator, line 558: Setting `archivedTxLastCheckTime` inside the for loop of txQueryStrings could cause the `receiver` query type to not execute since `archivedTxLastCheckTime` is global to both send and receive queries

src/cosmos/CosmosEngine.ts
```diff
@@ -520,38 +520,50 @@ export class CosmosEngine extends CurrencyEngine<
...
-        (newestTxid != null &&
-          this.otherData[query]?.newestTxid !== newestTxid) ||
-        lastPage !== this.otherData[query]?.lastPage
-      ) {
-        this.otherData[query] = { newestTxid, lastPage }
-        this.walletLocalDataDirty = true
+      for (const clients of clientsList) {
+        const { newestTxid, lastTimestamp } = await this.queryTransactionsInner(
+          query,
+          clients
```
paullinator, line 551: Removing the starting page seems to mean you'll be querying all transactions every time and not continuing from where you left off.

paullinator (changes requested):

src/cosmos/CosmosEngine.ts
```diff
@@ -559,7 +557,10 @@ export class CosmosEngine extends CurrencyEngine<
           this.otherData[query]?.newestTxid !== newestTxid
         ) {
           this.otherData[query] = { newestTxid }
-          this.otherData.archivedTxLastCheckTime = lastTimestamp
+          this.otherData.archivedTxLastCheckTime = Math.max(
```
paullinator, line 560: approved with requested small change of only setting archivedTxLastCheckTime once both send and recv transactions sycned.

paullinator (approved):

PR #815, opened 2024-08-13
Add evmscan fallback for eth_getTransactionReceipt in getL1RollupFee

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207882588906997

### DIFF
--- a/src/ethereum/networkAdapters/EvmScanAdapter.ts
+++ b/src/ethereum/networkAdapters/EvmScanAdapter.ts
@@ -1,4 +1,5 @@
 import { add, max, mul, sub } from 'biggystring'
+import { asMaybe } from 'cleaners'
 import {
   EdgeConfirmationState,
   EdgeCurrencyInfo,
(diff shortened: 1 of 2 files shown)


src/ethereum/networkAdapters/EvmScanAdapter.ts
```diff
@@ -367,11 +368,18 @@ export class EvmScanAdapter extends NetworkAdapter<EvmScanAdapterConfig> {
...
       )
-      const json = asGetTransactionReceipt(response.result.result)
+      let json = asMaybe(asGetTransactionReceipt)(rpcResponse.result.result)
+      if (json == null) {
+        const path = `?module=proxy&action=eth_getTransactionReceipt&txhash=${txid}`
+        const evmScanResponse = await this.serialServers(
+          async server => await this.fetchGetEtherscan(server, path)
+        )
+        json = asGetTransactionReceipt(evmScanResponse.result)
+      }
```
samholmes, line 382: Why not exclusively use this method to get the receipt since this is a EvmScanAdapter. I suppose it's nice to have a fallback, but perhaps the fallback should be `multicastRpc` instead of the evmscan connection. By removing `multicastRpc` from EvmScanAdapter, we've decoupled the evmscan adapter from the rpc adapter entirely. Is there any reason why we should keep multicastRpc around other then some redundancy?

samholmes (approved):
Approved, but with consideration to change.

src/ethereum/networkAdapters/EvmScanAdapter.ts
peachbits: Besides redundancy the other consideration is potentially cost since rpcs are generally cheaper to use but the potential benefit/detriment would vary chain by chain. This method isn't public on the adaptor so I don't think it matters what node list is actually used under the hood.

PR #820, opened 2024-08-26
Expose non-cached getters for addresses, names, and domains

Jon-edge:
We need the ability to provide fresh data on demand to the GUI. Previously, refreshing this data was only done in `checkAccountInnerLoop` .

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208153057785487

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- added: Public FIO fetch fns to return non-cached addresses and domains
+
 ## 4.21.0 (2024-08-29)
 
 - added: POL token on Ethereum.
--- a/src/fio/FioEngine.ts
+++ b/src/fio/FioEngine.ts
@@ -198,6 +198,14 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
       fioAction: async (actionName: string, params: any): Promise<any> => {
         return await this.multicastServers(actionName, params)
       },
+      fetchFioAddresses: async (): Promise<FioAddress[]> => {
+        await this.refreshFioAddresses()
+        return this.otherData.fioAddresses
+      },
+      fetchFioDomains: async (): Promise<FioDomain[]> => {
+        await this.refreshFioDomains()
+        return this.otherData.fioDomains
+      },
       getFioAddresses: async (): Promise<FioAddress[]> => {
         return this.otherData.fioAddresses
       },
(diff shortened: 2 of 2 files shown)


src/fio/FioEngine.ts
```diff
@@ -198,6 +198,18 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
       fioAction: async (actionName: string, params: any): Promise<any> => {
         return await this.multicastServers(actionName, params)
       },
+      fetchFioAddresses: async (): Promise<FioAddress[]> => {
+        await this.refreshFioAddresses()
```
swansontec, line 202: Since we are doing the fetch right away, should we delete the `try/catch` block from the `refreshFioAddresses` method? That way, if you call `fetchFioAddresses`, and something goes wrong, you get a normal error instead of a `console.warn` and an empty return value.

src/fio/FioEngine.ts
```diff
@@ -198,6 +198,18 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
...
+        return this.otherData.fioAddresses
+      },
+      fetchFioAddressNames: async (): Promise<string[]> => {
+        await this.refreshFioAddresses()
+        return this.otherData.fioAddresses.map(fioAddress => fioAddress.name)
```
swansontec, line 207: This seems redundant with `fetchFioAddressNames`, since the client can just do the `map` itself. These methods are pretty heavy-weight now, since they query the network.

src/fio/FioEngine.ts
```diff
@@ -198,6 +198,18 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
...
+      },
+      fetchFioDomains: async (): Promise<FioDomain[]> => {
+        await this.refreshFioDomains()
+        return this.otherData.fioDomains
```
swansontec, line 211: Same concern about error handling here too.

src/fio/FioEngine.ts
```diff
@@ -1098,18 +1059,76 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
...
+        this.localDataDirty()
+        this.updateStakingStatus()
+      }
+    } catch (e: any) {
+      this.log.warn('checkAccountInnerLoop getFioBalance error: ', e)
+    }
+
+    // Fio Addresses and Domains
+    const isAddressChanged = await this.refreshFioAddresses()
+    const isDomainChanged = await this.refreshFioDomains()
```
swansontec, line 1129: If you still want error handling in the inner loop, you can just put a try / catch around these two guys specifically.

src/fio/FioEngine.ts
```diff
@@ -939,62 +951,7 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
...
-    try {
-      const balances = { staked: '0', locked: '0' }
-      let balanceRes: BalanceResponse
-      try {
-        balanceRes = asGetFioBalanceResponse(
-          await this.multicastServers('getFioBalance')
-        )
-      } catch (e: any) {
-        if (e?.json?.message === MAINNET_LOCKS_ERROR) {
-          // This is a fallback query for this chain bug https://fioprotocol.atlassian.net/wiki/spaces/DAO/pages/852688908/2024-02-07+Token+Locking+Issue+on+Unstake
```
swansontec, line 963: I noticed that this comment didn't get copied over. Is it somehow irrelevant now?

swansontec (changes requested):

src/fio/FioEngine.ts
Jon-edge: Still relevant, will add back

src/fio/FioEngine.ts
Jon-edge: I did consider going that route, that doesn't count as changing public APIs?

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- added: Public FIO fetch fns to return non-cached addresses, names, and domains
+- added: Public FIO fetch fns to return non-cached addresses domains
```
swansontec, line 5: "addresses *and* domains"

swansontec (approved):
Approved, but one typo in the changelog.

PR #821, opened 2024-08-27
Cardano Staking Infra

samholmes:
In addition to the changes needed for Cardano staking, there is a private key optimization included in this PR to improve signing performance.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207289569255345

### DIFF
--- a/src/cardano/CardanoEngine.ts
+++ b/src/cardano/CardanoEngine.ts
@@ -76,27 +76,19 @@ export class CardanoEngine extends CurrencyEngine<
     this.otherData = asCardanoWalletOtherData(_raw)
   }
 
-  async fetchGet(method: string): Promise<unknown> {
-    const res = await this.fetchCors(
-      `${this.networkInfo.koiosServer}/api/v1/${method}`
-    )
-    if (!res.ok) {
-      const message = await res.text()
-      throw new Error(`Koios error: ${message}`)
-    }
-    const json = await res.json()
-    return json
-  }
-
-  async fetchPost(method: string, body: JsonObject): Promise<unknown> {
-    const opts = {
-      method: 'POST',
-      headers: { 'content-type': 'application/json' },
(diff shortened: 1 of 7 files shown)


src/cardano/CardanoEngine.ts
```diff
@@ -566,6 +567,77 @@ export class CardanoEngine extends CurrencyEngine<
...
+
+  decodeStakingTx = async (encodedTx: string): Promise<EdgeTransaction> => {
+    const { bech32Address } = asSafeCardanoWalletInfo(this.walletInfo).keys
+
+    const tx = Cardano.Transaction.from_hex(encodedTx)
+    const txHash = Cardano.hash_transaction(tx.body())
+
+    const txJson = tx.to_js_value()
+
+    // Validate the transaction is a staking transaction:
```
peachbits, line 604: Personal preference: I would put the actual comparison of the addresses in outputs and our bech32address in this function instead of buried in a cleaner

src/cardano/CardanoEngine.ts
```diff
@@ -521,6 +519,7 @@ export class CardanoEngine extends CurrencyEngine<
     vkeyWitnesses.add(vkeyWitness)
 
     if (isStakeTx) {
+      const stakeKey = accountKey.derive(2).derive(0)
```
peachbits, line 522: This belongs in "Add isStakeTx otherParams field for signing with stake key" commit

peachbits (approved):
Approved with minor optional considerations.

src/cardano/CardanoEngine.ts
samholmes: Interesting preference. My thought is that asStakingTx.ts is concerned with validation, so why move that concern up to the caller?

PR #824, opened 2024-08-28
Rename MATIC to POL on Polygon native chains

swansontec:
The token needs to be swapped on Ethereum, so also add the new POL contract address. On the native chain, it's just a name change.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none


src/ethereum/info/polygonInfo.ts
```diff
@@ -216,7 +216,7 @@ const networkInfo: EthereumNetworkInfo = {
 
 const currencyInfo: EdgeCurrencyInfo = {
   canReplaceByFee: true,
-  currencyCode: 'MATIC',
+  currencyCode: 'POL',
```
peachbits, line 219: We should force a resync so the user doesn't have to. Search for "zksyncForceResyncUSDC" as an example

peachbits (changes requested):

src/ethereum/info/polygonInfo.ts
swansontec: Do we really need this? The `lastTransactionQueryHeight.POL` will simply be 0, because we have never seen POL before, and the user will automatically get the correct transactions. This would only be an issue if the user adds a custom token on Polygon called "MATIC", in which case they might see the wrong history.

peachbits: Yup you're right not necessary, my mistake

peachbits (approved):

PR #825, opened 2024-08-30
(ZcashPiratechain) Add error listeners

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/react-native-piratechain/pull/14
https://github.com/EdgeApp/react-native-zcash/pull/56

### Description

 none

---
  - https://app.asana.com/0/0/1208178889823047

### DIFF
--- a/src/piratechain/piratechainTypes.ts
+++ b/src/piratechain/piratechainTypes.ts
@@ -8,6 +8,7 @@ import {
   asValue,
   Cleaner
 } from 'cleaners'
+import type { ErrorEvent as PiratechainErrorEvent } from 'react-native-piratechain'
 import type { Subscriber } from 'yaob'
 
 import { asWalletInfo } from '../common/types'
@@ -112,6 +113,7 @@ export interface PiratechainSynchronizer {
   on: Subscriber<{
     statusChanged: PiratechainStatusEvent
     update: PiratechainUpdateEvent
+    error: PiratechainErrorEvent
   }>
   start: () => Promise<void>
   stop: () => Promise<void>
--- a/src/piratechain/piratechainTypes.ts
+++ b/src/piratechain/piratechainTypes.ts
@@ -8,7 +8,18 @@ import {
   asValue,
   Cleaner
 } from 'cleaners'
-import type { ErrorEvent as PiratechainErrorEvent } from 'react-native-piratechain'
+import type {
+  Addresses,
+  BlockRange,
+  ConfirmedTransaction,
+  ErrorEvent as PiratechainErrorEvent,
+  InitializerConfig,
+  SpendInfo,
+  SpendSuccess,
+  StatusEvent,
+  UpdateEvent,
+  WalletBalance
+} from 'react-native-piratechain'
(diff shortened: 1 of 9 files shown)


src/zcash/zcashTypes.ts
```diff
@@ -13,6 +13,7 @@ import {
 import type {
   Addresses,
   BalanceEvent,
+  ErrorEvent as ZcashErrorEvent,
```
samholmes, line 16: I'm getting `Module '"react-native-zcash"' has no exported member 'ErrorEvent'`

src/zcash/ZcashEngine.ts
```diff
@@ -168,6 +168,13 @@ export class ZcashEngine extends CurrencyEngine<
       })
       this.onUpdateTransactions()
     })
+    this.synchronizer.on('error', async payload => {
```
samholmes, line 171: I want to make sure the payload is not really an `unknown` type, but the library doesn't appear to have the type information that this change assume (when I checkout at least).

src/piratechain/piratechainTypes.ts
```diff
@@ -8,7 +8,18 @@ import {
...
+  BlockRange,
+  ConfirmedTransaction,
+  ErrorEvent as PiratechainErrorEvent,
+  InitializerConfig,
+  SpendInfo,
+  SpendSuccess,
+  StatusEvent,
+  UpdateEvent,
+  WalletBalance
+} from 'react-native-piratechain'
```
samholmes, line 22: I'm not able to reproduce these exports from `react-native-piratechain`. Is there a package.json change somewhere that's missing from this branch?

src/piratechain/piratechainTypes.ts
```diff
@@ -25,83 +36,18 @@ export interface PiratechainNetworkInfo {
...
-  lastScannedHeight: number
-  scanProgress: number // 0 - 100
-  networkBlockHeight: number
-}
-
-// Block range is inclusive
-export const asPiratechainBlockRange = asObject({
-  first: asNumber,
-  last: asNumber
-})
```
samholmes, line 94: You can keep this and pass the generic type argument to `asObject<BlockRange>` to ensure the cleaner matches the type.

samholmes (changes requested):

src/zcash/zcashTypes.ts
peachbits: That makes sense because this PR is dependent on new versions of rnzcash and rnpirate that don't exist yet

src/piratechain/piratechainTypes.ts
peachbits: Yes, to versions of rnzcash and rnpirate that don't exist yet

src/zcash/ZcashEngine.ts
peachbits: You can checkout and updot the dependent PRs for types or test again once those are published and updated in the package.json in this PR

samholmes (approved):

PR #828, opened 2024-09-03
Handle FIO 404 errors for getFioAddresses

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208215298507086

### DIFF
--- a/src/fio/FioEngine.ts
+++ b/src/fio/FioEngine.ts
@@ -199,11 +199,16 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
         return await this.multicastServers(actionName, params)
       },
       fetchFioAddresses: async (): Promise<FioAddress[]> => {
-        await this.refreshFioAddresses()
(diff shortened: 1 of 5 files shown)


src/fio/FioEngine.ts
Jon-edge, line 1: I'm not so sure about hiding all errors here. We're assuming the caller would never be interested in knowing why something went wrong. There's no way to differentiate between no addresses/domains vs an error. I could imagine a future manual refresh UI wanting to know if these calls failed for some reason

Jon-edge (changes requested):

src/fio/FioEngine.ts
samholmes: Hmm.. No that I think about; I'm not sure about calling `refreshFioAddresses` on ever call to `getFioAddresses`. Why are we doing what was a background task on ever call to `getFioAddresses`? What was the motivation for your change again?

PR #832, opened 2024-09-11
Merge duplicate token transaction data from evmscan

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208217866090168

### DIFF
--- a/src/ethereum/networkAdapters/EvmScanAdapter.ts
+++ b/src/ethereum/networkAdapters/EvmScanAdapter.ts
@@ -287,7 +287,7 @@ export class EvmScanAdapter extends NetworkAdapter<EvmScanAdapterConfig> {
     const address = this.ethEngine.walletLocalData.publicKey
     let page = 1
 
-    const allTransactions: EdgeTransaction[] = []
+    let allTransactions: EdgeTransaction[] = []
     let server: string | undefined
     while (true) {
       const offset = NUM_TRANSACTIONS_TO_QUERY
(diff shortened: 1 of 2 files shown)


src/ethereum/networkAdapters/EvmScanAdapter.ts
```diff
@@ -557,6 +567,11 @@ export function mergeEdgeTransactions(
       isSend: nativeAmount.startsWith('-'),
       nativeAmount,
       networkFee: mergedNetworkFee,
+      ...(mergedParentNetworkFee != null
```
paullinator, line 570: I get it. but this is a good example of high cognitive load to understand. Maybe in the future declare the tx object outside of the `.set()`, then add the `parentNetworkFee` to it if needed. Then call the .set() with the tx object

```
const edgeTransaction: EdgeTransaction =  {
      ...existingTransaction,
      isSend: nativeAmount.startsWith('-'),
      nativeAmount,
      networkFee: mergedNetworkFee,
      ourReceiveAddresses: [
        ...existingTransaction.ourReceiveAddresses,
        ...transaction.ourReceiveAddresses
      ]
}

if (mergedParentNetworkFee != null) {
  edgeTransaction.parentNetworkFee = mergedParentNetworkFee
}

txidToTransaction.set(uniqueKey, edgeTransaction)
```

paullinator (approved):
Optional change for clarity

PR #841, opened 2024-09-30
Add redundancy to Cardano queries

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208401832264547

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- changed: Add redundancy to Cardano queries
+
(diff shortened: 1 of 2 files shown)


swansontec (approved):

swansontec (approved):
I don't know about the second commit either, but psychologically people are OK with things being slow as long as they get progress. It's easier to wait a week with tracking info that wait 4-5 days without tracking information, so I'm generally in favor of this sort of thing.

PR #845, opened 2024-10-04
Await synchronizer initialization in public methods instead of throwing `synchronizer undefined`error

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208287065968694

### DIFF
--- a/src/piratechain/PiratechainEngine.ts
+++ b/src/piratechain/PiratechainEngine.ts
@@ -43,7 +43,7 @@ export class PiratechainEngine extends CurrencyEngine<
   synchronizerStatus!: StatusEvent['name']
   availableZatoshi!: string
   initialNumBlocksToDownload!: number
-  initializer!: InitializerConfig
+  birthdayHeight: number
   progressRatio!: number
   queryMutex: boolean
   makeSynchronizer: (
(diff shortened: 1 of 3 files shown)


src/piratechain/PiratechainEngine.ts
```diff
@@ -295,12 +300,14 @@ export class PiratechainEngine extends CurrencyEngine<
...
     try {
-      this.synchronizer = await this.makeSynchronizer({
+      this.synchronizerPromise = this.makeSynchronizer({
         mnemonicSeed: piratechainPrivateKeys.mnemonic,
         birthdayHeight: piratechainPrivateKeys.birthdayHeight,
         alias: base16.stringify(base64.parse(this.walletId)),
         ...rpcNode
       })
+      this.synchronizer = await this.synchronizerPromise
+      this.synchronizerResolver(this.synchronizer)
```
swansontec, line 310: This logic was hard to understand, but I finally figured it out. Maybe it needs a comment:

1. Replace `this.synchronizerPromise` with a fresh promise. The old promise might have already been resolved.
2. People might be waiting on the old promise, so resolve that.

swansontec (approved):
Confusing, but it works.

PR #847, opened 2024-10-24
Add TON

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208326136610301


swansontec (commented):
Where did the rest of the comments go?

src/ton/TonEngine.ts
```diff
@@ -0,0 +1,470 @@
...
+      fees.source_fees.storage_fee
+    const networkFee = totalFee.toString()
+
+    const total = add(nativeAmount, networkFee)
+    const balance = this.getBalance({ tokenId })
+    if (
+      edgeSpendInfoIn.skipChecks !== true &&
+      gt(add(total, this.networkInfo.minimumAddressBalance), balance)
+    ) {
+      throw new Error('InsufficientFundsError')
```
swansontec, line 326: This is incorrect. Please do `new InsufficientFundsError({ tokenId })`

src/ton/TonTools.ts
```diff
@@ -0,0 +1,266 @@
...
+        .then(() => {
+          clientsLastUpdate = Date.now()
+        })
+        .catch(this.log.error)
+    }
+    return shuffleArray(this.clients)
+  }
+
+  async updateClients(): Promise<void> {
+    const res = await this.io.fetchCors(this.networkInfo.orbsNetworkUrl)
```
swansontec, line 117: This seems like a job for the info server. The info server payload for TON should include an `orbsUrls` field, and then we should have a background process on the info server that fetches this information and shoves it in CouchDB. Then, the core info endpoint would merge the `orbsUrls` field into the payload on it way out the door.

swansontec (changes requested):
I had about 6 comments on this PR, but GitHub lost them, and now I don't remember what they were. These are the important ones.

swansontec (approved):
Nice.

PR #852, opened 2024-11-06
Implement LLD tx history

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207238792534941

### DIFF
--- a/src/polkadot/PolkadotEngine.ts
+++ b/src/polkadot/PolkadotEngine.ts
@@ -7,9 +7,12 @@ import { abs, add, div, gt, lte, mul, sub } from 'biggystring'
 import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
+  EdgeCurrencyInfo,
   EdgeFetchFunction,
   EdgeFreshAddress,
   EdgeSpendInfo,
+  EdgeTokenId,
+  EdgeTokenMap,
   EdgeTransaction,
   EdgeWalletInfo,
   InsufficientFundsError,
@@ -32,12 +35,15 @@ import {
 } from '../common/utils'
 import { PolkadotTools } from './PolkadotTools'
 import {
+  asLiberlandMeritsResponse,
+  asLiberlandTransfersResponse,
   asPolkadotWalletOtherData,
   asPolkapolkadotPrivateKeys,
   asSafePolkadotWalletInfo,
   asSubscanResponse,
   asTransactions,
   asTransfer,
+  LiberlandTransfer,
   PolkadotNetworkInfo,
   PolkadotWalletOtherData,
   SafePolkadotWalletInfo,
@@ -86,6 +92,35 @@ export class PolkadotEngine extends CurrencyEngine<
       // API returns valid response for empty accounts. For other errors just assume the recipient's account is sufficient.
       return this.minimumAddressBalance
     }
+  }
+
+  async fetchLiberlandScan(
+    endpoint: string,
+    operationName: string,
+    query: string,
+    variables: JsonObject
+  ): Promise<any> {
+    const body = {
+      operationName,
+      variables,
+      query
+    }
+
+    const options = {
+      method: 'POST',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify(body)
+    }
+
+    const response = await this.fetchCors(endpoint, options)
+    if (!response.ok) {
+      throw new Error(`Liberland API failed with status ${response.status}`)
+    }
+    const out = await response.json()
+    return out
   }
 
   async fetchSubscan(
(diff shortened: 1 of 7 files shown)


src/polkadot/PolkadotEngine.ts
```diff
@@ -637,3 +787,54 @@ export async function makeCurrencyEngine(
...
+    blockHeight,
+    currencyCode: context.currencyInfo.currencyCode,
+    date,
+    isSend: nativeAmount.startsWith('-'),
+    memos: [],
+    nativeAmount,
+    networkFee: '0', // Fee data not provided
+    ourReceiveAddresses,
+    signedTx: '',
+    tokenId: null,
```
peachbits, line 835: Token transaction support?

src/polkadot/PolkadotEngine.ts
```diff
@@ -210,6 +243,116 @@ export class PolkadotEngine extends CurrencyEngine<
...
+        // Update progress
+        const totalCount = cleanResponse.data.transfers.totalCount
+        processedCount += transfers.length
+
+        this.tokenCheckTransactionsStatus[this.currencyInfo.currencyCode] =
+          totalCount === 0 ? 1 : processedCount / totalCount
+        this.updateOnAddressesChecked()
+      } catch (e: any) {
+        this.warn(`Error fetching Liberland transactions: ${e.message}`)
+        break
```
peachbits, line 340: Breaking here would exit the while loop and set tokenCheckTransactionsStatus to 1 which wouldn't be accurate.

src/polkadot/PolkadotEngine.ts
```diff
@@ -637,3 +787,54 @@ export async function makeCurrencyEngine(
...
+
+  let nativeAmount = value
+  const denom = getDenomination(
+    context.currencyInfo.currencyCode,
+    context.currencyInfo,
+    context.allTokensMap
+  )
+  if (denom == null) return
+
+  nativeAmount = value
```
peachbits, line 817: This variable is initialized twice

peachbits (changes requested):

src/polkadot/PolkadotEngine.ts
peachbits: Issue still exists. throw the errors instead of breaking so you don't update the sync status

src/polkadot/PolkadotEngine.ts
```diff
@@ -210,6 +243,116 @@ export class PolkadotEngine extends CurrencyEngine<
     this.addTransaction(this.currencyInfo.currencyCode, edgeTransaction)
   }
 
+  async queryLiberlandTransactions(): Promise<void> {
+    return await queryTxMutex(
```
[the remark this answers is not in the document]
peachbits: Unaddressed. Thoughts?

src/polkadot/polkadotTypes.ts
```diff
@@ -19,6 +19,7 @@ export interface PolkadotNetworkInfo {
   subscanBaseUrl: string | undefined
   subscanQueryLimit: number
   lengthFeePerByte: string
+  liberlandScanUrl: string | undefined
```
peachbits, line 22: Set this in the liberlandTestnetInfo.ts or make it optional

src/polkadot/PolkadotEngine.ts
```diff
@@ -637,3 +890,62 @@ export async function makeCurrencyEngine(
...
+          context.allTokensMap
+        )
+      : context.allTokensMap[tokenId].denominations[0]
+
+  if (denom == null) return
+
+  if (fromId === context.walletInfo.keys.publicKey) {
+    nativeAmount = `-${nativeAmount}`
+  } else {
+    ourReceiveAddresses.push(toId)
```
peachbits, line 932: Replace toId with context.walletInfo.keys.publicKey

peachbits (changes requested):

src/polkadot/PolkadotEngine.ts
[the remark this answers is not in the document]
Jon-edge: Somehow didn't push my actual changes, sorry about that

src/polkadot/PolkadotEngine.ts
Jon-edge: Somehow didn't push my actual changes, sorry about that

src/polkadot/info/polkadotInfo.ts
```diff
@@ -14,7 +14,8 @@ const networkInfo: PolkadotNetworkInfo = {
   subscanBaseUrl: 'https://polkadot.api.subscan.io/api',
   subscanQueryLimit: 100,
   partialFeeOffsetMultiplier: '1',
-  lengthFeePerByte: '1000000'
+  lengthFeePerByte: '1000000',
+  liberlandScanUrl: undefined
```
peachbits, line 18: Wrong info file changed

src/polkadot/PolkadotEngine.ts
```diff
@@ -637,3 +890,62 @@ export async function makeCurrencyEngine(
...
+
+export function processLiberlandTransaction(
+  context: LiberlandTxProcessingContext,
+  tx: LiberlandTransfer
+): EdgeTransaction | undefined {
+  const { id, fromId, value, block } = tx
+  const { tokenId } = context
+
+  const blockHeight = parseInt(block.number)
+  const date = new Date(block.timestamp).getTime() / 1000 // Convert timestamp to epoch seconds
```
peachbits, line 910: I looked at the return object and the timestamp appears to be UTC but doesn't include the Z at the end. You can add it to the end to ensure it's handled correctly

src/polkadot/PolkadotEngine.ts
```diff
@@ -210,6 +245,217 @@ export class PolkadotEngine extends CurrencyEngine<
...
+              walletInfo: this.walletInfo,
+              currencyInfo: this.currencyInfo,
+              allTokensMap: this.allTokensMap,
+              tokenId
+            },
+            tx
+          )
+          if (edgeTransaction != null) {
+            if (
+              this.findTransaction(currencyCode, edgeTransaction.txid) !== -1
```
peachbits, line 425: For whatever reasons this isn't exiting early on the first loop which means the entire history of the wallet is queried every time the user logs in. Save the latest txid as a variable on otherData (lots of other engines do this and you'll need to track a txid each for lld and llm) and exit early on that.

peachbits (changes requested):

src/polkadot/PolkadotEngine.ts
Jon-edge: Cause it wasn't normalized. Even then, searching this way makes this break on an unconfirmed tx. Though the engine never saves unconfirmed tx's, the GUI does immediately after sending, so in that situation the send tx never gets updated from pending.

Updated to check confirmation state.

peachbits (changes requested):

peachbits (approved):

PR #854, opened 2024-11-13
Fix EVM websocket polling

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208525574472617

### DIFF
--- a/src/ethereum/EthereumNetwork.ts
+++ b/src/ethereum/EthereumNetwork.ts
@@ -40,7 +40,11 @@ interface EthereumNeeds {
    * is not defined, then the engine will always pull for address data; it
    * assume all needs are true.
    */
-  addressSync?: {
+  addressSync: {
+    // This should be true immediately after connecting to a network adapter.
+    needsInitialSync: boolean
+    // These are the txids that need to be checked from the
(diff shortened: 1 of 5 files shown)


src/ethereum/EthereumNetwork.ts
```diff
@@ -346,71 +396,15 @@ export class EthereumNetwork {
         // batch token balance queries.
         !isFetchTokenBalancesSupported
       ) {
-        await this.checkAndUpdate(
-          this.ethNeeds.tokenBalLastChecked[currencyCode] ?? 0,
-          BAL_POLL_MILLISECONDS,
-          async () => await this.check('fetchTokenBalance', currencyCode)
-        )
+        this.acquireTokenBalance(currencyCode)(currencyCode).catch(error => {
+          console.error(error)
```
swansontec, line 408: The old `checkAndUpdate` function would call `this.ethEngine.error('checkAndUpdate ', ...`. Shouldn't we continue to report errors in the same way?

swansontec (approved):
Approving, because the changes would hopefully be small.

PR #861, opened 2024-12-05
Deprecate Binance Beacon Chain

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208889831655306


src/binance/BinanceTools.ts
```diff
@@ -57,26 +54,29 @@ export class BinanceTools implements EdgeCurrencyTools {
...
     if (!isValid) throw new Error('Invalid BNB mnemonic')
-    const binanceKey = getPrivateKeyFromMnemonic(mnemonic)
 
-    return { binanceMnemonic: mnemonic, binanceKey }
+    const binanceKeyBytes = mnemonicToSeedSync(binanceMnemonic)
+    const master = bip32.fromSeed(binanceKeyBytes)
+    const child = master.derivePath(`44'/714'/0'/0/0`)
+    if (child.privateKey == null) {
+      throw new Error('child does not have a privateKey')
+    }
```
swansontec, line 66: How solid is the testing on this? I don't see anything wrong, but then again I don't know what the derivation path or key formats are off the top of my head. If this gives the same results as the previous code, then it's good.

swansontec (approved):

src/binance/BinanceTools.ts
peachbits: It's copied from the sdk that's being removed. There's fixtures that test it

PR #865, opened 2024-12-10
Use `fetchCorsForced` for Solana RPC nodes

samholmes:
Solana RPC nodes will throw RPC errors instead of throwing HTTP errors
so they'll fail silently and not reach the CORS fallback methods.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208916602572140

### DIFF
--- a/src/solana/SolanaEngine.ts
+++ b/src/solana/SolanaEngine.ts
@@ -39,7 +39,7 @@ import { CurrencyEngine } from '../common/CurrencyEngine'
 import { PluginEnvironment } from '../common/innerPlugin'
 import { getRandomDelayMs } from '../common/network'
 import { asyncWaterfall, promiseAny } from '../common/promiseUtils'
-import { cleanTxLogs, getFetchCors, getOtherParams } from '../common/utils'
+import { cleanTxLogs, getOtherParams } from '../common/utils'
 import { SolanaTools } from './SolanaTools'
 import {
   AccountBalance,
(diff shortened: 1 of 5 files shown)


src/solana/SolanaEngine.ts
```diff
@@ -92,7 +92,8 @@ export class SolanaEngine extends CurrencyEngine<
     this.lightMode = opts.lightMode ?? false
     this.networkInfo = env.networkInfo
     this.chainCode = tools.currencyInfo.currencyCode
-    this.fetchCors = getFetchCors(env.io)
+    this.fetch = async (uri, opts) =>
```
swansontec, line 95: `avoidCors: 'always'`. We always want to bypass CORS restrictions.

src/solana/SolanaTools.ts
```diff
@@ -219,7 +219,7 @@ export class SolanaTools implements EdgeCurrencyTools {
     const connectionConfig: ConnectionConfig = {
       commitment: this.networkInfo.commitment,
       // @ts-expect-error our fetch is close enough to the fetch api
-      fetch: this.io.fetchCors
+      fetch: this.io.fetchCorsForced
```
swansontec, line 222: This needs to be `(url, opts) => this.io.fetch(uri, { ...opts, avoidCors: 'always' })`

swansontec (changes requested):

swansontec (approved):

PR #869, opened 2024-12-13
XRP/XLM update mnemonic support

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208526037776002


src/stellar/StellarTools.ts
swansontec, line 1: While we are here, please fix `importPrivateKeys` in `StellarTools` to just:

```ts
if (...) {
   ...
  return {
    stellarKey: keypair.secret()
    stellarMnemonic: userInput
  }
} else {
  ...
  return {
    stellarKey:  userInput.replace(/ /g, '')
    stellarMnemonic: undefined
  }
}
```

That would get rid of the two local variables. Actually, looking in there, the line `userInput.replace(/ /g, '')` does nothing, since `replace` just returns a new string, but doesn't modify the existing string. Probably should fix that.

src/ripple/RippleTools.ts
```diff
@@ -91,14 +106,22 @@ export class RippleTools implements EdgeCurrencyTools {
...
-      // If that worked, return the key:
-      return { rippleKey: privateKey }
+      if (validateMnemonic(input)) {
+        // Try creating wallet
+        Wallet.fromMnemonic(input, { mnemonicEncoding: 'bip39' })
+        return {
+          rippleMnemonic: input
+        }
+      } else {
+        const privateKey = input.replace(/\s/g, '')
```
swansontec, line 117: Yes, this line of code actually does something useful. We should fix the Stellar import to work the same.

swansontec (changes requested):
All your changes look fine, but when I was reviewing this, I found some awful stuff going on inside the Stellar key import routine. Let's get that cleaned up.

swansontec (approved):

PR #875, opened 2025-01-02
Integrate Seen Tx Checkpoint API

samholmes:
- **WIP Upgrade edge-core-js local**
- **Integrate new Seen Tx Checkpoint API**

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208854722050406


src/cardano/CardanoEngine.ts
```diff
@@ -320,20 +321,17 @@ export class CardanoEngine extends CurrencyEngine<
...
+      this.transactionEvents = []
     }
   }
 
   // // ****************************************************************************
   // // Public methods
   // // ****************************************************************************
-
-  async startEngine(): Promise<void> {
+  async startEngine(opts?: EdgeCurrencyEngineStartOptions): Promise<void> {
```
swansontec, line 334: Please put back the blank line.

src/cardano/CardanoEngine.ts
```diff
@@ -342,13 +340,13 @@ export class CardanoEngine extends CurrencyEngine<
     this.addToLoop('queryTransactions', TRANSACTION_POLL_MILLISECONDS).catch(
       () => {}
     )
-    await super.startEngine()
+    await super.startEngine({ seenTxCheckpoint: this.seenTxCheckpoint })
```
swansontec, line 343: Shouldn't this be `opts`, instead of `{ seenTxCheckpoint: this.seenTxCheckpoint }`

src/polkadot/PolkadotEngine.ts
```diff
@@ -598,7 +595,7 @@ export class PolkadotEngine extends CurrencyEngine<
   async resyncBlockchain(): Promise<void> {
     await this.killEngine()
     await this.clearBlockchainCache()
-    await this.startEngine()
+    await this.startEngine({ seenTxCheckpoint: this.seenTxCheckpoint })
```
swansontec, line 598: I see this pattern in all the `resyncBlockchain` methods, where we pass in the `{ seenTxCheckpoint: this.seenTxCheckpoint}` block to `startEngine`. Could we get rid of this somehow, maybe by having the base class just keep whatever it had before?

swansontec (approved):
Approving, since the changes are tiny.

src/polkadot/PolkadotEngine.ts
samholmes: The base class is keeping the state in `this.seenTxCheckpoint`, yes. 

If `startEngine` is called without a `seenTxCheckpoint`, it will set the state in the base class to undefined. So I don't see how you propose to get rid of the need for it to be an argument to this `startEngine` invocation.

samholmes: This is defunct now that we agreed to change the core API (see core PR).

src/tezos/TezosEngine.ts
```diff
@@ -387,13 +386,13 @@ export class TezosEngine extends CurrencyEngine<
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
     this.addToLoop('checkTransactionsInnerLoop', TRANSACTION_POLL_MILLISECONDS)
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    super.startEngine(opts)
+    super.startEngine()
```
swansontec, line 389: This could be `await super.startEngine()`, and then we can delete the comment above.

swansontec (approved):
This looks good, but I found a cleanup that might be worth doing since we're here.

PR #877, opened 2025-01-07
Add SUI

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208800068201258

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- added: Add SUI
 - fixed: Ignore incorrect fees sent by the Midgard server, and instead just use 0.02 RUNE for Thorchain native transactions.
 
 ## 4.32.4 (2024-12-25)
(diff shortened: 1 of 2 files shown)


src/sui/SuiTools.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  async importPrivateKey(input: string): Promise<JsonObject> {
+    const isValid = validateMnemonic(input)
+    if (!isValid) throw new Error('Invalid mnemonic')
+
+    // test mnemonic
+    Ed25519Keypair.deriveKeypair(input)
+
+    return { [this.networkInfo.pluginMnemonicKeyName]: input }
```
swansontec, line 73: Could we do:

```ts
const wasKeys = uncleaner(asSuiPrivateKeys(pluginId))
return wasKeys({mnemonic:input })
```

And then you can delete `pluginMnemonicKeyName` from the `networkInfo`. This is literally the whole reason we define `asSuiPrivateKeys` using `asCodec`, to handle the round-tripping.

swansontec (changes requested):
Looks pretty nice.

swansontec (changes requested):
One thing got missed, but we are sooo close.

swansontec (approved):

PR #878, opened 2025-01-10
Update layer-2 currencyInfos with chain and asset names

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-core-js/pull/635

### Description

 none


swansontec (changes requested):
I have requested the core to make these two new fields mandatory for all chains. So, this PR will need to be expanded to hit all plugins across the board. What you have so far is just fine.

src/ripple/rippleInfo.ts
```diff
@@ -16,7 +16,8 @@ const networkInfo: XrpNetworkInfo = {
 
 const currencyInfo: EdgeCurrencyInfo = {
   currencyCode: 'XRP',
-  displayName: 'XRP',
+  assetDisplayName: 'XRP',
+  chainDisplayName: 'XRP',
```
swansontec, line 20: Isn't the chain itself called "Ripple"?
/me ducks for cover

src/ethereum/info/ethDevInfo.ts
```diff
@@ -129,7 +129,8 @@ const currencyInfo: EdgeCurrencyInfo = {
   currencyCode: 'ETH',
   customFeeTemplate: evmCustomFeeTemplate,
   customTokenTemplate: evmCustomTokenTemplate,
-  displayName: 'Dev Ethereum',
+  assetDisplayName: 'Dev Ethereum',
+  chainDisplayName: 'Dev Ethereum',
```
swansontec, line 133: The other Ethereum testnets, like Holesky Testnet, just use "Ethereum" as the asset name. The other non-Ethereum testnets, like Hedera Testnet, work like this one, where they have "Testnet" in the asset name.

I don't know what the correct solution is, or if there is even a problem in the first place, but it is a discrepancy.

swansontec (approved):
Approved. Feel free to merge this as-is, but I did notice two discrepancies that are worth a second look.

PR #880, opened 2025-01-11
Fix FIO obt data encryption/decryption

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208711103056196


src/fio/FioEngine.ts
```diff
@@ -235,24 +235,20 @@ export class FioEngine extends CurrencyEngine<FioTools, SafeFioWalletInfo> {
...
-          (await obtDecoder.decrypt({
-            obt_data_records: encryptedObtData,
-            more: 0
-          })) ?? { obt_data_records: [] }
-
-        return decryptedObtData.obt_data_records
+
+        const res = await fioSdk.genericAction('getObtData', {
+          limit: 100,
+          offset: 0,
```
peachbits, line 247: This implementation will only query 100 records and stop. Is that appropriate here? If you don't ultimately need fetchEncryptedObtData in that case you can remove it.

Also, while not really important here, calling fioSdk.getObtData() directly instead of genericAction gives you return types instead of any.

peachbits (changes requested):

peachbits (approved):

PR #882, opened 2025-01-15
Detect chainIds in eip681 URIs

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209002118306877


src/ethereum/EthereumTools.ts
```diff
@@ -205,6 +206,21 @@ export class EthereumTools implements EdgeCurrencyTools {
     }
     address = contractAddress
 
+    // Split the address to get the chainId according to EIP-681
+    if (address.includes('@')) {
```
paullinator, line 210: Does this allow using an Ethereum: uri with an L2 like optimism if and only if the chainId exists?

src/ethereum/EthereumTools.ts
```diff
@@ -205,6 +213,29 @@ export class EthereumTools implements EdgeCurrencyTools {
...
 
+    if (
+      parsedUri.protocol === 'ethereum' &&
+      mustValidateChainId &&
+      !address.includes('@')
+    ) {
+      throw new Error(
+        'chainId must be present when using ethereum: prefix on a non-ethereum network'
+      )
+    } else if (address.includes('@')) {
```
swansontec, line 224: This `else` isn't needed, since the previous `if` can only throw. Un-indenting this block would make it more obvious that it runs in every case, assuming we pass the error check.

swansontec (approved):
Approved, but please remove the `else`

PR #884, opened 2025-01-22
Solana query cleanups

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209199800706689

### DIFF
--- a/src/solana/SolanaEngine.ts
+++ b/src/solana/SolanaEngine.ts
@@ -38,7 +38,7 @@ import { base16, base64 } from 'rfc4648'
 import { CurrencyEngine } from '../common/CurrencyEngine'
 import { PluginEnvironment } from '../common/innerPlugin'
 import { getRandomDelayMs } from '../common/network'
-import { asyncWaterfall, promiseAny } from '../common/promiseUtils'
+import { asyncWaterfall, promiseAny, timeout } from '../common/promiseUtils'
 import { cache, cleanTxLogs, getOtherParams } from '../common/utils'
 import { SolanaTools } from './SolanaTools'
 import {
@@ -73,7 +73,7 @@ export class SolanaEngine extends CurrencyEngine<
   base58PublicKey: string
   feePerSignature: string
   getPriorityFee: () => Promise<string>
-  recentBlockhash: string
+  getRecentBlockhash: () => Promise<string>
   chainCode: string
   otherData!: SolanaWalletOtherData
   fetch: EdgeFetchFunction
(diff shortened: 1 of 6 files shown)


src/solana/SolanaEngine.ts
```diff
@@ -157,48 +156,39 @@ export class SolanaEngine extends CurrencyEngine<
...
+      )
+
+      const allTokenIds = [...Object.keys(this.allTokensMap)]
       const detectedTokenIds: string[] = []
-      for (const [i, tokenId] of allTokenIds.entries()) {
-        const tokenBal = asMaybe(asTokenBalance)(tokenBals[i])
-        // empty token addresses return an error "Invalid param: could not find account".
-        // If there was an actual error with the request it would have thrown already
-        const balance = tokenBal?.result?.value?.amount ?? '0'
+      for (const tokenId of allTokenIds) {
```
swansontec, line 190: Faster & easier to read if you get rid of the `allTokenIds` variable: `for (const tokenId of Object.keys(this.allTokensMap))`.

Also, why are we looping over every token, including disabled ones? Is this for auto-detect tokens? Otherwise, this.enabledTokenIds seems better.

src/solana/SolanaEngine.ts
```diff
@@ -726,12 +726,12 @@ export class SolanaEngine extends CurrencyEngine<
         lamports: parseInt(nativeAmount)
       })
       instructions.push(transferInstruction)
-    }
 
-    if (tokenId != null && eq(totalTxAmount, balance)) {
-      // This is a max token send so we don't need to consider the minimumAddressBalance
-    } else if (gt(add(totalTxAmount, minimumAddressBalance), balance)) {
-      throw new InsufficientFundsError({ tokenId })
+      if (eq(totalTxAmount, balance)) {
```
swansontec, line 741: So this check is only relevant to SOL, not tokens? Because before, it only applied to tokens for some reason. I guess that was the bug.

swansontec (changes requested):
I really like the new `cache` function!

src/solana/SolanaEngine.ts
peachbits: Correct it's for auto-detection

src/solana/SolanaEngine.ts
peachbits: Yeah this specific minimumAddressBalance check is only relevant to SOL transfers and was incorrectly checking token amounts

swansontec (approved):
Approved for staging

PR #886, opened 2025-01-25
Fix TON transaction handling

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209207870668232

### DIFF
--- a/src/ton/TonEngine.ts
+++ b/src/ton/TonEngine.ts
@@ -82,7 +82,7 @@ export class TonEngine extends CurrencyEngine<TonTools, SafeCommonWalletInfo> {
 
   async queryBalance(): Promise<void> {
     try {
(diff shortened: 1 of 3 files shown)


src/ton/TonEngine.ts
```diff
@@ -121,7 +121,7 @@ export class TonEngine extends CurrencyEngine<TonTools, SafeCommonWalletInfo> {
           lt: inLoopLogicalTime,
           hash: inLoopHash,
           inclusive: false,
-          archival: this.archiveTransactions
+          archival: true // must always be true, because reasons
```
swansontec, line 124: // ... because ton center reasons

PR #888, opened 2025-01-27
Add Solana broadcast retry logic

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209228669381844

### DIFF
--- a/src/solana/SolanaEngine.ts
+++ b/src/solana/SolanaEngine.ts
@@ -5,6 +5,7 @@ import {
 } from '@solana/spl-token'
 import {
   AccountInfo,
+  BlockhashWithExpiryBlockHeight,
   ComputeBudgetProgram,
   ConfirmedSignatureInfo,
   Keypair,
@@ -16,12 +17,11 @@ import {
   TokenBalance,
   TransactionInstruction,
   TransactionResponse,
-  TransactionSignature,
   VersionedTransaction,
   VersionedTransactionResponse
 } from '@solana/web3.js'
 import { add, eq, gt, gte, lt, max, mul, sub } from 'biggystring'
-import { asMaybe } from 'cleaners'
+import { asMaybe, asString } from 'cleaners'
 import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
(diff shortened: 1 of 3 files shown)


src/solana/SolanaEngine.ts
```diff
@@ -765,15 +768,19 @@ export class SolanaEngine extends CurrencyEngine<
...
 
     const unsignedTx =
       asSolanaSpendInfoOtherParams(spendInfoOtherParams)?.unsignedTx ??
       versionedTx.serialize()
 
-    const otherParams = wasSolanaTxOtherParams({ unsignedTx })
+    const otherParams = wasSolanaTxOtherParams({
+      unsignedTx,
+      blockhash: '',
+      lastValidBlockHeight: 0
```
swansontec, line 782: Since we have it already, we could return `blockhash: recentBlockhash.blockhash,
lastValidBlockHeight: recentBlockhash.lastValidBlockHeight` instead of dummy values. They aren't used, true. Still, might be safer?

src/solana/SolanaEngine.ts
```diff
@@ -838,16 +851,82 @@ export class SolanaEngine extends CurrencyEngine<
...
+      const blockheightResults = await Promise.allSettled(broadcastPromises)
+      const blockheights = blockheightResults
+        .filter(p => p.status === 'fulfilled')
+        .map(r => (r as PromiseFulfilledResult<number>).value)
+      const maxHeight = Math.max(...blockheights)
+      return maxHeight
+    }
+
+    const retryTxSubmission = async (): Promise<void> => {
+      while (true && !retryTxController.signal.aborted) {
```
swansontec, line 895: The `true && ` seems useless?

src/solana/SolanaEngine.ts
```diff
@@ -838,16 +851,82 @@ export class SolanaEngine extends CurrencyEngine<
...
+        }
+
+        // Confirmed!!
+        retryTxController.abort()
       })
-      const txid: TransactionSignature = await promiseAny(promises)
+
+      await Promise.race([retryTxSubmission(), ...confirmPromises])
+
       edgeTransaction.txid = txid
```
swansontec, line 930: Where does `txid` come from? Both `confirmPromises` and `submitTx` appear to use `const txid` locally, so I'm not sure how it flows out into this outer scope.

swansontec (approved):
Approved, but I have questions.

src/solana/SolanaEngine.ts
peachbits: yup, resolved

src/solana/SolanaEngine.ts
peachbits: It's returned from the standalone call to submitTx up above

src/solana/SolanaEngine.ts
peachbits: sure thing, done

PR #894, opened 2025-02-17
Optimize Solana transaction query

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209208659072454

### DIFF
--- a/src/common/CurrencyEngine.ts
+++ b/src/common/CurrencyEngine.ts
@@ -808,6 +808,13 @@ export class CurrencyEngine<
     return checkpointB
   }
 
+  updateTransactionEvents(): void {
+    if (this.transactionEvents.length > 0) {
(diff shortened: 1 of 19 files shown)


src/common/CurrencyEngine.ts
```diff
@@ -796,6 +796,14 @@ export class CurrencyEngine<
     return checkpointB
   }
 
+  updateTransactionEvents(): void {
+    if (this.transactionEvents.length > 0) {
+      this.walletLocalDataDirty = true
```
swansontec, line 801: The `CurrencyEngine.addTransaction` call already handles `walletLocalDataDirty = true` if there are any changes, so do we need it here?

swansontec (approved):
One question, but otherwise this makes sense.

PR #897, opened 2025-02-24
Use a `PeriodicTask` in `addToLoop`

swansontec:
Instead of directly messing with timers, we can use a periodic task to ensure that there is exactly one copy of each task running at once, and automatically start & stop them as the engine toggles on or off.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209488312133556


peachbits (approved):
Looks good to me!

PR #899, opened 2025-02-25
Add RLUSD

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209103243011508

### DIFF
--- a/src/ripple/rippleInfo.ts
+++ b/src/ripple/rippleInfo.ts
@@ -46,6 +46,21 @@ const currencyInfo: EdgeCurrencyInfo = {
 }
 
 export const builtinTokens: EdgeTokenMap = {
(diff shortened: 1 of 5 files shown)


src/ripple/rippleInfo.ts
```diff
@@ -46,6 +46,21 @@ const currencyInfo: EdgeCurrencyInfo = {
 }
 
 export const builtinTokens: EdgeTokenMap = {
```
paullinator, line 48: need to add SOLO

paullinator (changes requested):

paullinator (approved):

PR #902, opened 2025-03-06
Add handled broadcast error

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209345494542285

### DIFF
--- a/src/algorand/AlgorandEngine.ts
+++ b/src/algorand/AlgorandEngine.ts
@@ -695,6 +695,30 @@ export class AlgorandEngine extends CurrencyEngine<
       Uint8Array.from(base16.parse(encodedTx))
     )
 
+    // If we are trying to send a token, check first if the recipient has
+    // activated/opted-in the asset
+    const { tokenId, spendTargets } = edgeTransaction
+    if (
+      tokenId != null &&
+      spendTargets != null &&
+      spendTargets[0].nativeAmount !== '0'
+    ) {
+      const { assets } = await this.fetchAccountInfo(
+        spendTargets[0].publicAddress
+      )
+      if (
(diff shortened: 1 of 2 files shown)


src/algorand/AlgorandEngine.ts
```diff
@@ -729,6 +729,11 @@ export class AlgorandEngine extends CurrencyEngine<
       edgeTransaction.txid = txId
       edgeTransaction.date = Date.now() / 1000
     } catch (e: any) {
+      // Recipient did not activate what we're trying to send
```
peachbits, line 732: Relying on a string would work but it's a fragile solution. Instead, let's query for the address's information and see if the asset is enabled. You can use the fetchAccountInfo helper and move the check into signTx

peachbits (changes requested):

src/algorand/AlgorandEngine.ts
```diff
@@ -701,6 +701,22 @@ export class AlgorandEngine extends CurrencyEngine<
         rawTx.amount.toString(),
         recipient
       )
+
+      // Check if the recipient has activated/opted-in the asset
+      if (edgeTransaction.tokenId != null) {
```
peachbits, line 706: I don't think this block could ever run. The recipient is only defined for "pay" transaction types which are ALGO sends and it's undefined for "axfer" token sends. 

You'll also need to be sure this doesn't break token activation transactions which is a token send to yourself and the asset wouldn't be present in the account info response.

peachbits (changes requested):

peachbits (approved):

PR #903, opened 2025-03-06
Allow react-native-zcash / piratechain to be missing

swansontec:
If these aren't installed natively, the related plugins won't work, but everything else will.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none


peachbits (approved):
I believe some of the types exported by react-native-zcash and react-native-piratechain spends are incorrect but the changes made to accommodate them won't break anything. Worst case scenario is losing the error message thrown from native code.

PR #908, opened 2025-03-19
Integrate Edge Change Server

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209371329306828

### DIFF
--- a/src/ethereum/EthereumEngine.ts
+++ b/src/ethereum/EthereumEngine.ts
@@ -6,6 +6,7 @@ import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
   EdgeCurrencyInfo,
+  EdgeEngineSyncNetworkOptions,
   EdgeFetchFunction,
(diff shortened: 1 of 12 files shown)


src/ethereum/EthereumEngine.ts
```diff
@@ -775,6 +786,41 @@ export class EthereumEngine extends CurrencyEngine<
     await this.startEngine()
   }
 
+  async syncNetwork(opts: EdgeEngineSyncNetworkOptions): Promise<number> {
```
swansontec, line 789: We have added this new logic, but we never got rid of `this.setupAdapterSubscriptions()` over in EthereumNetwork, so we are now subscribed twice? I would have expected more deleted code.

swansontec (changes requested):

swansontec (approved):

PR #909, opened 2025-03-19
Tron - add note support to tx query

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209654951272385

### DIFF
--- a/src/tron/TronEngine.ts
+++ b/src/tron/TronEngine.ts
@@ -480,6 +480,12 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
 
     if (retArray.length < 1) return out
 
+    const { contractRet: status, fee } = retArray[0]
+
+    const success = status === 'SUCCESS'
+    const feeNativeAmount = fee.toString()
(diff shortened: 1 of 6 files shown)


src/tron/TronEngine.ts
```diff
@@ -481,6 +482,15 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
 
     const ourReceiveAddresses: string[] = []
 
+    const memos: EdgeMemo[] = []
+    if (data != null) {
+      memos.push({
+        type: 'text',
+        value: Buffer.from(base16.parse(data)).toString('ascii'),
```
swansontec, line 518: Are memos really ASCII? I would have expected UTF-8, at least. What if the raw bytes aren't valid UTF-8? Do we need a try/catch around this? Would it make sense for the `catch` block to emit a "hex" memo with the raw bytes that we can't decode?

swansontec (approved):

swansontec (approved):

PR #910, opened 2025-03-20
Replace asyncWaterfall with promiseAny for orbs connections

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none


swansontec (approved):
I don't understand how anything in this PR works. The changes make sense on their own, but I don't see how they fix anything.

peachbits:
> I don't understand how anything in this PR works. The changes make sense on their own, but I don't see how they fix anything.

Async waterfall takes too long to get through a server list and setting a shorter timeout doesn't give the nodes enough time to respond. Better to try them all instead.

PR #912, opened 2025-03-20
Use Filfox API for balance data

samholmes:
It appears the Glif API has added pricing models to their DePIN, so switching from using GLIF to Filfox as a dependency for the balance data seems reasonable.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

### DIFF
--- a/src/filecoin/Filfox.ts
+++ b/src/filecoin/Filfox.ts
@@ -97,6 +97,31 @@ export const asFilfoxTransfer = asObject({
 })
 
 //
+// Account
+//
+
+export type FilfoxAccountResult = ReturnType<typeof asFilfoxAccountResult>
+export const asFilfoxAccountResult = asObject({
+  id: asString,
+  // robust: asString,
+  // actor: asString,
+  // createHeight: asNumber,
+  // createTimestamp: asNumber,
+  // lastSeenHeight: asNumber,
+  // lastSeenTimestamp: asNumber,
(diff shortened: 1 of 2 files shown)


src/filecoin/Filfox.ts
```diff
@@ -138,6 +163,23 @@ export class Filfox {
     this.fetch = fetchFn
   }
 
+  async getAccount(address: string): Promise<FilfoxAccountResult> {
+    const url = new URL(`${this.baseUrl}/address/${address}`)
```
peachbits, line 167: The URL object is useless here

src/filecoin/Filfox.ts
```diff
@@ -96,6 +96,31 @@ export const asFilfoxTransfer = asObject({
   value: asString
 })
 
+//
+// Account
+//
+
+export type FilfoxAccountResult = ReturnType<typeof asFilfoxAccountResult>
+export const asFilfoxAccountResult = asObject({
```
peachbits, line 104: This cleaner is too strict for new unused wallets where the response object is:

{"address":"f1pzdgw5mdxqnu6aixuy4bt4ygj6lhaajb3x7tgwi"}

New wallets will get stuck syncing

peachbits (changes requested):

peachbits (approved):

PR #919, opened 2025-04-02
EVM - add dynamic minimum custom gas price

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

### DIFF
--- a/src/ethereum/EthereumEngine.ts
+++ b/src/ethereum/EthereumEngine.ts
@@ -813,7 +813,13 @@ export class EthereumEngine extends CurrencyEngine<
       spendInfo.spendTargets[0].nativeAmount = balance
 
       // Use our calcMiningFee function to calculate the fees:
-      const miningFees = calcMiningFees(this.networkInfo, spendInfo, null)
+      const currentBaseFeeWei = await this.ethNetwork.getBaseFeePerGas()
+      const miningFees = calcMiningFees(
+        this.networkInfo,
+        spendInfo,
+        null,
+        currentBaseFeeWei
+      )
 
       // If our results require a call to the RPC server to estimate the gas limit, then we
       // need to do that now:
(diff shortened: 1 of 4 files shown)


src/ethereum/EthereumNetwork.ts
```diff
@@ -223,7 +223,7 @@ export class EthereumNetwork {
     return out
   }
 
-  getBaseFeePerGas = async (): Promise<string> => {
+  getBaseFeePerGas = async (): Promise<string | undefined> => {
```
samholmes, line 226: > Fix return type on getBaseFeePerGas

But what was wrong with the return type?

src/ethereum/EthereumEngine.ts
```diff
@@ -991,10 +997,16 @@ export class EthereumEngine extends CurrencyEngine<
     }
 
     const edgeToken = tokenId != null ? this.builtinTokens[tokenId] : null
+    const networkBaseFeeWeiHex = await this.ethNetwork.getBaseFeePerGas()
+    const currentBaseFeeWei =
+      networkBaseFeeWeiHex != null
+        ? hexToDecimal(networkBaseFeeWeiHex)
+        : undefined
```
samholmes, line 1004: Why don't you need to do hexToDecimal conversion in the other instance?

samholmes (changes requested):

src/ethereum/EthereumNetwork.ts
peachbits: I thought I explained this in a commit message but I don't see it. Plugins that don't support this rpc method return undefined because asyncWaterfall always returns undefined when passed an empty array.

peachbits: The undefined return value is actually expacted and handled correctly in a few places

src/ethereum/EthereumEngine.ts
peachbits: Missed it, fixed

src/ethereum/EthereumNetwork.ts
samholmes: A more thorough fix to this issue would be to change ayncWaterfall's signature to:

``ts
export async function asyncWaterfall<T>(
  asyncFuncs: AsyncFunction[],
  timeoutMs: number = 5000
): Promise<T | undefined> {
```

And then fix all the places where there are type errors

samholmes (approved):

PR #921, opened 2025-04-16
Jon/fix/trx-tip-491

Jon-edge:
# Real-Time TRC-20 Token Transfer Fee Calculation on Tron

This document outlines how to accurately estimate TRC-20 token transfer fees on the Tron blockchain in real time. It explains the key components of fee calculation, caveats of Tron’s simulation API, and the heuristic needed to handle common discrepancies.

---

## 🧮 Fee Components

TRC-20 token transfers consume two Tron resources:

1. **Bandwidth** (bytes of transaction data)
2. **Energy** (for smart contract execution)

The total TRX fee is:

```
Total Fee =
  (Bandwidth Used - Bandwidth Remaining) × TransactionFee +
  (Energy Used - Energy Remaining) × EnergyFee +
  CreateAccountFee (if applicable) +
  MemoFee (if applicable)
```

---

### 1. Contract Energy Factor

TRON contracts may have a configured `energy_factor`, which multiplies the effective energy cost for callers if the contract owner has insufficient staked energy.

- Default: `10000` (represents 1.0× cost)
- High values (e.g., `34000`) indicate a 3.4× penalty

This value can be retrieved from:

Endpoint: `/wallet/getcontractinfo`  
Field: `contract_state.energy_factor`

### 2. Recipient Storage Status

Energy usage differs based on whether the recipient’s address already exists in the contract’s storage mapping:

| Recipient Storage State         | Energy Behavior                 |
|--------------------------------|---------------------------------|
| Already in storage             | Lower cost (storage update)     |
| Not yet in storage             | Higher cost (new storage entry) |

This behavior is tied to how EVM-like storage writes are priced ([1]).

To detect this accurately, simulate a dry-run `transfer()` call to the recipient.

---

## 🧪 Improved Detection: Dry-Run Transfer Simulation

Instead of checking token balances (which may be zero), use the following logic:

```ts
const dryRunBody = {
  owner_address: base58ToHexAddress(senderAddress),
  contract_address: base58ToHexAddress(contractAddress),
  function_selector: 'transfer(address,uint256)',
  parameter: encodeParams(['address', 'uint256'], [recipientAddress, '1']).slice(2)
}

const res = await fetch('/wallet/triggerconstantcontract', { method: 'POST', body: JSON.stringify(dryRunBody) })
const json = await res.json()
const energyUsed = json.energy_used

const recipientAlreadyInStorage = energyUsed < 50000 // heuristic threshold
```

This technique works regardless of whether the recipient has a current token balance.

---

## ⚠️ Problem with Tron’s Estimator

The Tron API provides `/wallet/triggerconstantcontract` to simulate contract calls and return `energy_used`. However, this simulation has two limitations:

1. It **does not reflect new storage slot creation** unless specifically triggered by a write path.
2. It often **underestimates energy usage by about 50%** for first-time recipients.

---

## ✅ Heuristic Fix

To align simulated estimates with actual on-chain behavior, apply a conditional correction:

```ts
const adjustedEnergy = recipientAlreadyInStorage
  ? json.energy_used
  : json.energy_used * 2.2
```

This ensures accurate estimates for both experienced and first-time recipients.

---

## 📌 Summary

| Scenario                             | Action                                      |
|--------------------------------------|---------------------------------------------|
| Recipient already in contract storage| Use `energy_used` as-is                     |
| Recipient new to storage             | Multiply `energy_used` by 2.2               |
| Contract has high energy_factor      | Already reflected in `energy_used`          |
| Node returns low estimate            | Apply fallback multiplier if needed         |

---

## 🔗 References

1. Ethereum Yellow Paper – SSTORE gas model: https://ethereum.github.io/yellowpaper/paper.pdf  
2. TIP-491: TRON Energy Factor: https://github.com/tronprotocol/tips/blob/master/tip-491.md  
3. GitHub Issue on Energy Estimation: https://github.com/tronprotocol/java-tron/issues/2982  
4. Reddit guide on USDT fee behavior: https://www.reddit.com/r/Tronix/comments/1clkav1/guide_usdt_fees_secrets/  
5. Empirical analysis of TRC-20 token transfers: user logs

### DIFF
--- a/src/tron/TronEngine.ts
+++ b/src/tron/TronEngine.ts
@@ -93,6 +93,7 @@ const ACCOUNT_POLL_MILLISECONDS = getRandomDelayMs(20000)
 const BLOCKCHAIN_POLL_MILLISECONDS = getRandomDelayMs(20000)
 const TRANSACTION_POLL_MILLISECONDS = getRandomDelayMs(20000)
 const NETWORKFEES_POLL_MILLISECONDS = getRandomDelayMs(60 * 60 * 1000) // 1 hour
+const DEFAULT_ENERGY_NO_BALANCE = 130000
 
 type TronFunction =
   | 'trx_blockNumber'
@@ -224,7 +225,9 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
         ['address[]', 'address[]'],
         [
           [address.slice(2)],
-          tokenIds.map(address => base58ToHexAddress(address).slice(2))
+          tokenIds.map(tokenAddress =>
+            base58ToHexAddress(tokenAddress).slice(2)
+          )
         ]
       )
 
@@ -937,7 +940,7 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
 
   private async getDynamicEnergyFactor(
     contractAddress: string
-  ): Promise<number> {
+  ): Promise<{ factor: number; energyPenaltyRatio: number }> {
     const body = { value: base58ToHexAddress(contractAddress) }
     const res = await this.multicastServers(
       'trx_getBalance',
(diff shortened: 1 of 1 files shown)


src/tron/TronEngine.ts
```diff
@@ -1008,31 +1065,54 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
...
+          `Using cached energy estimate for ${cacheKey}: ${this.energyEstimateCache[cacheKey]}`
+        )
       }
 
       // Get dynamic factor for the contract if available
-      energyFactor = await this.getDynamicEnergyFactor(contractAddress)
+      const energyFactorInfo = await this.getDynamicEnergyFactor(
+        contractAddress
+      )
+      energyFactor = energyFactorInfo.factor
```
peachbits, line 1101: These values are only used for logging.

peachbits (changes requested):

src/tron/TronEngine.ts
Jon-edge: Yeah, I want to make sure we maintain logs in case something goes wrong in the future, we don't have to try to somehow reproduce. I.e. when fees spike again or some other token that doesn't work with the 2.2 heuristic

src/tron/TronEngine.ts
```diff
@@ -970,22 +987,62 @@ export class TronEngine extends CurrencyEngine<TronTools, SafeTronWalletInfo> {
...
+
+          const res = await this.multicastServers(
+            'trx_getBalance',
+            '/wallet/triggerconstantcontract',
+            body
+          )
+
+          const rawBalances = asTRC20Balance(res).constant_result[0]
+          const decoded = decodeParams([], ['uint256[]'], `0x${rawBalances}`)
+          const balance = hexToDecimal(decoded[0][0]._hex)
```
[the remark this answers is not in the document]
Jon-edge: Good catch, found a new way to determine this more accurately. Implements `energy_penalty` which is specifically designed to capture storage expansion cost

peachbits (approved):

PR #925, opened 2025-05-01
Thorchain - Add TCY and testnet

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1210063554025852

### DIFF
--- a/src/cosmos/thorchainTypes.ts
+++ b/src/cosmos/thorchainTypes.ts
@@ -1,5 +1,5 @@
 import { HttpEndpoint } from '@cosmjs/stargate'
-import { asArray, asMaybe, asObject, asString } from 'cleaners'
+import { asArray, asMaybe, asObject, asOptional, asString } from 'cleaners'
 
 import { PluginEnvironment } from '../common/innerPlugin'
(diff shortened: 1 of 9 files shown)


src/cosmos/thorchainTypes.ts
```diff
@@ -48,11 +50,14 @@ export const asMidgardActionsResponse = asObject({
       metadata: asObject(
         asObject({
           memo: asString,
-          networkFees: asArray(
-            asObject({
-              amount: asString,
-              asset: asString // 'THOR.RUNE'
-            })
+          networkFees: asMaybe(
```
swansontec, line 53: Wouldn't it be `asOptional`, if it's simply missing? Otherwise we might hide arbitrary corruptions, such as `amount` being a number or something else. In that case, we would silently delete the whole fees array instead of complaining about the specific thing that's wrong inside.

swansontec (approved):
Approved with a question.

src/cosmos/thorchainTypes.ts
peachbits: yup, good catch

PR #930, opened 2025-05-08
Update TRX energy calculation logic

Jon-edge:
- Rely purely on `total_energy` or `energy_used` for the energy estimation, depending on `recipientInStorage`
- Update the `recipientInStorage` check to compare against a self spend dryrun.
- Add heuristic specific to USDT per observations in TronLink wallet and onchain transactions.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1210177602128615

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # edge-currency-accountbased
 
+- fixed: (TRX) Fees simplified to use `wallet/triggerconstantcontract` `energy_used`
+
 ## Unreleased
 
 ## 4.47.0 (2025-05-12)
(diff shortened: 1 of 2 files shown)


peachbits (changes requested):
Per IRL chat and testing node behavior. It appears we can now trust the energy_used value returned from wallet/triggerconstantcontract. Let's do this call for TRC20 transfers  and remove any additional consideration of the energy factor. Cache the response for a contract/recipient combo so makeSpend remains responsive but empty the cache after broadcasting

peachbits (approved):

PR #937, opened 2025-05-19
Zano - Add multi output support

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/react-native-zano/pull/5

### Description

 none

---
  - https://app.asana.com/0/0/1210210357632652

### DIFF
--- a/src/zano/zanoTypes.ts
+++ b/src/zano/zanoTypes.ts
@@ -1,4 +1,5 @@
 import {
+  asArray,
   asCodec,
   asNumber,
   asObject,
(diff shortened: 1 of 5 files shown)


src/zano/zanoTypes.ts
```diff
@@ -1,11 +1,13 @@
 import {
+  asArray,
   asCodec,
   asNumber,
   asObject,
   asOptional,
   asString,
   Cleaner
 } from 'cleaners'
+import { TransferParams } from 'react-native-zano'
```
swansontec, line 10: `import type`, just to be extra safe. If we import any actual code, it'll blow up the build.

swansontec (approved):
This needs one small change, but I'll approve it.

PR #938, opened 2025-05-21
Fill in MakeTxDeposit Edge transaction data when sending one asset

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1210336219030714

### DIFF
--- a/src/cosmos/engine/CosmosEngine.ts
+++ b/src/cosmos/engine/CosmosEngine.ts
@@ -23,7 +23,7 @@ import {
 } from '@cosmjs/stargate'
 import { longify } from '@cosmjs/stargate/build/queryclient'
 import { fromRfc3339WithNanoseconds, toSeconds } from '@cosmjs/tendermint-rpc'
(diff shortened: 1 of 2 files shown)


src/cosmos/engine/CosmosEngine.ts
```diff
@@ -176,38 +177,73 @@ export class CosmosEngine extends CurrencyEngine<
...
+                  ([_, token]) => {
+                    return token.currencyCode === currencyCode
+                  }
+                )
+                if (tokenObj != null) {
+                  tokenId = tokenObj[0]
+                } else {
+                  // if we don't know the asset, just let it continue as tokenId: null until we see it off the network.
+                }
+                nativeAmount = asset.amount
```
paullinator, line 222: should fallback to old implementation if you can't find the tokenId so the amount and tokenId are in the same units

paullinator (approved):

PR #952, opened 2025-06-27
Sam/eth sync regressions

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1210649610530917


src/common/utils.ts
```diff
@@ -423,3 +423,34 @@ export const createWeightedAverageCalculator = (weights: {
...
+      return await fn()
+    } catch (e) {
+      lastError = e
+      if (attempt === Number(maxRetries)) break
+      await snooze(delay)
+      delay *= factor
+      attempt++
+    }
+  }
+  throw lastError
```
swansontec, line 455: The two `maxRetries` checks seem redundant? Is the following code the same:

```ts
while (true) {
  try {
      return await fn()
    } catch (error) {
      if (attempt >= maxRetries) throw error
      await snooze(delay)
      delay *= factor
      attempt++
    }
}
```

swansontec (approved):
One question.

swansontec (approved):

PR #974, opened 2025-08-27
Add pre-built transaction support to SuiEngine

samholmes:
Enable SuiEngine to execute pre-built transactions passed via otherParams. This allows swap plugins like LI.FI to provide complete transaction data that will be executed as-is rather than being converted to simple transfers.

The implementation checks for unsignedTx or unsignedBase64 in otherParams and uses the pre-built transaction when provided, while maintaining backward compatibility for regular transfers.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1211165107420343

### DIFF
--- a/src/sui/SuiEngine.ts
+++ b/src/sui/SuiEngine.ts
@@ -25,7 +25,11 @@ import { CurrencyEngine } from '../common/CurrencyEngine'
 import { PluginEnvironment } from '../common/innerPlugin'
 import { getRandomDelayMs } from '../common/network'
 import { asMaybeContractLocation } from '../common/tokenHelpers'
-import { asSafeCommonWalletInfo, SafeCommonWalletInfo } from '../common/types'
+import {
+  asSafeCommonWalletInfo,
+  MakeTxParams,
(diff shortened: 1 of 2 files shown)


peachbits (changes requested):
Per IRL convo, let's move the new logic into otherMethods.makeTx

samholmes:
I've updated the implementation based on the feedback. The changes now:

1. Add `otherMethods.makeTx` as requested, following the pattern used by other engines (cosmos, xrp)
2. Keep the existing `makeSpend` implementation completely untouched - it still has the pre-built transaction support
3. The `makeTx` method accepts a `Uint8Array` for the unsigned transaction and optional metadata

The `otherMethods.makeTx` provides a clean interface for creating EdgeTransactions from pre-built transaction data, while the existing `makeSpend` logic remains unchanged for backward compatibility and continues to handle pre-built transactions passed via `otherParams`.

All tests pass and the implementation is ready for re-review.

jfjeifkvbk (approved):

PR #1033, opened 2026-02-25
Reduce UI lag: gate redundant callbacks in accountbased engines

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Reduces RN JS thread starvation by eliminating redundant callbacks from accountbased currency engines. Five targeted fixes:

- **Stop calling `onBlockHeightChanged`**: EVM/accountbased engines no longer call `onBlockHeightChanged`, removing ~130 expensive core-js callbacks per 10s (polygon/fantom fire ~1/sec). Confirmations are now updated directly on transactions and emitted via `onTransactions`.
- **Gate `onNewTokens` via `reportDetectedTokens`**: Adds a `detectedTokenIds` cache (persisted to `walletLocalData`) in the base class. Engines called `onNewTokens` on every poll cycle even when the token list hadn't changed. Only fires when genuinely new tokens are discovered.
- **Gate `onStakingStatusChanged` via `reportStakingStatus`**: Deduplicates staking status callbacks using JSON.stringify comparison. Tron and FIO engines were firing on every poll cycle with identical data.
- **`SyncTracker` per-engine dedup**: Adds `sendSyncStatusIfChanged` to suppress `onSyncStatusChanged` when the `totalRatio` hasn't changed. Solana alone produced 129 redundant calls per 10s at steady state.
- **Global 500ms throttle on `sendSyncStatus`**: Limits aggregate sync status volume to 2/sec across all 30+ accountbased engines during the post-login burst.

---

> [!NOTE]
> **Medium Risk**
> Touches shared engine plumbing and callback emission semantics across many chains, so regressions could affect token discovery, staking UI updates, or confirmation/sync progress reporting despite being performance-focused.
> 
> **Overview**
> Reduces JS-thread/UI lag by **deduplicating high-volume callbacks** emitted by account-based currency engines.
> 
> Adds base `CurrencyEngine` helpers `reportDetectedTokens` (with persisted `walletLocalData.detectedTokenIds`) and `reportStakingStatus` (JSON-based dedupe), and migrates multiple engines (Algorand, Cosmos, Ripple, Solana, Sui, Tron, Zano, and `EthereumNetwork`) to use them instead of firing `onNewTokens`/`onStakingStatusChanged` repeatedly.
> 
> Stops emitting the deprecated `onBlockHeightChanged` callback; `updateBlockHeight` now updates in-memory tx confirmations and emits changes via `onTransactions`. `SyncTracker` now suppresses unchanged `onSyncStatusChanged` updates and applies a global 500ms throttle for non-`totalRatio=1` statuses.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93692e0e676a502839c4973c5e915da2e0952358. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/common/SyncTracker.ts
+++ b/src/common/SyncTracker.ts
@@ -1,5 +1,8 @@
 import type { EdgeSyncStatus, EdgeTokenId } from 'edge-core-js/types'
 
+// Global throttle: max 1 sendSyncStatus per 500ms; totalRatio=1 always passes.
+let ssLastEmitTime = 0
+
 /**
  * Abstracts the ability to return a sync status,
  * since different chains track their sync status in different ways.
@@ -74,6 +77,13 @@ export function makeTokenSyncTracker(engine: SyncEngine): TokenSyncTracker {
       lastSyncStatus.totalRatio !== currentStatus.totalRatio
     ) {
       lastSyncStatus = currentStatus
+
+      if (currentStatus.totalRatio !== 1) {
+        const now = Date.now()
+        if (now - ssLastEmitTime < 500) return
+        ssLastEmitTime = now
+      }
+
       engine.sendSyncStatus(currentStatus)
     }
   }
--- a/src/common/SyncTracker.ts
+++ b/src/common/SyncTracker.ts
@@ -43,6 +43,7 @@ export function makeTokenSyncTracker(engine: SyncEngine): TokenSyncTracker {
   // Each tokenId can be a 0-1 value:
   const balanceRatios = new Map<EdgeTokenId, number>()
   const historyRatios = new Map<EdgeTokenId, number>()
+  let lastSyncStatus: EdgeSyncStatus | undefined
 
   function getSyncStatus(): EdgeSyncStatus {
     const activeTokenIds = [null, ...engine.enabledTokenIds]
@@ -66,10 +67,22 @@ export function makeTokenSyncTracker(engine: SyncEngine): TokenSyncTracker {
     return { totalRatio }
   }
 
+  function sendSyncStatusIfChanged(): void {
+    const currentStatus = getSyncStatus()
+    if (
+      lastSyncStatus == null ||
+      lastSyncStatus.totalRatio !== currentStatus.totalRatio
+    ) {
+      lastSyncStatus = currentStatus
+      engine.sendSyncStatus(currentStatus)
+    }
+  }
+
   const out: TokenSyncTracker = {
     resetSync() {
       balanceRatios.clear()
       historyRatios.clear()
+      lastSyncStatus = undefined
     },
 
     balanceComplete(tokenId) {
@@ -78,12 +91,12 @@ export function makeTokenSyncTracker(engine: SyncEngine): TokenSyncTracker {
 
     setBalanceRatios(tokenIds, ratio) {
       for (const tokenId of tokenIds) balanceRatios.set(tokenId, ratio)
-      engine.sendSyncStatus(getSyncStatus())
+      sendSyncStatusIfChanged()
     },
 
     setHistoryRatios(tokenIds, ratio) {
       for (const tokenId of tokenIds) historyRatios.set(tokenId, ratio)
-      engine.sendSyncStatus(getSyncStatus())
+      sendSyncStatusIfChanged()
     },
 
     updateBalanceRatio(tokenId, ratio) {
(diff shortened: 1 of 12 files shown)


eddy-edge (commented):
## Additional Findings

- **warning** `src/common/CurrencyEngine.ts:735`: Removal of `onBlockHeightChanged` callback. The old code dispatched `onBlockHeightChanged(blockHeight)` which core-js uses to update `walletState.height` in redux. This code no longer calls it, meaning `walletState.height` in core-js will never update for this plugin. Consumers reading `EdgeCurrencyWallet.blockHeight` will see a stale value. The `@deprecated` annotation in core-js types suggests a replacement is planned but may not be in place yet.
  - Verify that edge-core-js has an alternative path to learn the block height (e.g. via `onSyncStatusChanged`). If not, continue calling `onBlockHeightChanged` for the height update even if confirmation handling is done engine-side. Check the minimum core-js version this plugin targets.
- **warning** `src/zano/ZanoEngine.ts:89`: If `deleteWallet` fails, the error is only logged and `needsNativeStorageClear` is already reset to `false` (line 90). The resync proceeds with stale wallet files, silently defeating the purpose of `resyncBlockchain`.
  - Consider leaving `needsNativeStorageClear = true` until deletion succeeds (move the reset to after `deleteWallet` returns), or re-throw the error so the lifecycle manager's `onError` path is triggered.
- **warning** `CHANGELOG.md:3`: The `## Unreleased` section is empty, but this PR contains significant performance improvements (SyncTracker 500ms throttle, staking status dedup, token detection dedup, removal of onBlockHeightChanged) that are not documented in any changelog entry. These are the core changes implied by the branch name `paul/fixUiLag`.
  - Add changelog entries documenting the performance improvements, e.g.:
- `fixed: Throttle sync status updates to reduce UI re-renders`
- `fixed: Gate staking status callbacks to only fire on actual changes`
- `fixed: Deduplicate detected token reports`
- `changed: Remove deprecated onBlockHeightChanged callback`
- **warning** `CHANGELOG.md:22`: Grammar: `Use correct Alchemy URL's` — the apostrophe is incorrect for a plural.
  - Change to `Use correct Alchemy URLs`.
- **suggestion** `CHANGELOG.md:5`: Four version bumps (4.74.2, 4.75.0, 4.75.1, 4.75.2) are batched into a single PR.
  - Document in the PR description why these versions are batched, if this is a changelog catch-up.
- **suggestion** `CHANGELOG.md:5`: The removal of `onBlockHeightChanged` in favor of engine-side confirmation handling is a behavioral/API change that warrants at least a `changed:` changelog entry.
  - Add a `changed:` entry and consider whether a minor version bump is appropriate for semver compliance.

src/common/SyncTracker.ts
```diff
@@ -66,10 +70,29 @@ export function makeTokenSyncTracker(engine: SyncEngine): TokenSyncTracker {
...
+      lastSyncStatus == null ||
+      lastSyncStatus.totalRatio !== currentStatus.totalRatio
+    ) {
+      lastSyncStatus = currentStatus
+
+      if (currentStatus.totalRatio !== 1) {
+        const now = Date.now()
+        if (now - ssLastEmitTime < 500) return
+        ssLastEmitTime = now
+      }
```
eddy-edge: **Warning:** Silent data loss from throttle ordering: `lastSyncStatus` is updated (line 79) before the throttle check (lines 81-84). If the call is throttled via `return`, the status is recorded but never sent. Subsequent calls with the same ratio will be deduplicated against the unsent value and also never sent. Example: ratio goes 0.3→0.5 (throttled, saved but not sent) → 0.5 again (deduped, never sent). The UI would remain stuck at 0.3.

**Recommendation:** Move `lastSyncStatus = currentStatus` to just before `engine.sendSyncStatus(currentStatus)` (after the throttle gate), so only actually-emitted statuses are recorded for deduplication.

src/common/SyncTracker.ts
```diff
@@ -1,5 +1,8 @@
 import type { EdgeSyncStatus, EdgeTokenId } from 'edge-core-js/types'
 
+// Global throttle: max 1 sendSyncStatus per 500ms; totalRatio=1 always passes.
+let ssLastEmitTime = 0
```
eddy-edge: **Warning:** Global throttle shared across all wallet engines. `ssLastEmitTime` is module-level, so if wallet A emits at T=0, wallet B is blocked until T=500ms. This couples independent wallets' sync status reporting and can cause stale progress bars during multi-wallet sync (e.g. initial login with many wallets).

**Recommendation:** Move `ssLastEmitTime` into the per-tracker closure (next to `lastSyncStatus` on line 49), so each wallet engine has its own independent 500ms throttle. If a global cross-wallet limit is truly desired, document it explicitly.

PR #1044, opened 2026-04-03
Fix DOT and TON syncing

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes TON’s balance/fee/seqno/broadcast network paths to a new dRPC REST client and gates Polkadot tx syncing on a new Subscan API key, which could affect syncing/broadcast behavior if the new endpoints or keys are misconfigured.
> 
> **Overview**
> **DOT:** Adds `initOptions.subscanApiKey` and sends it as `X-API-Key` on Subscan requests; transaction syncing is now only scheduled when a key is provided, and Subscan `transfers` parsing tolerates `null` entries.
> 
> **TON:** Replaces Orbs-server usage with a new dRPC REST wrapper (`tonDrpc.ts`) and updates balance lookup, fee estimation, seqno fetching, and broadcasting to use dRPC endpoints; adds `TonNetworkInfo.drpcUrl` plus optional `initOptions.drpcApiKey` with staggered fallback between free and paid dRPC URLs, and updates deploy-txid detection by polling dRPC transaction results.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7680d601635bbb90df27f4fc1394bc7fd24be790. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1213882158399672

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- added: (DOT) Add Subscan API key support
 - changed: (TON) Replace Orbs servers with dRPC
 
 ## 4.78.0 (2026-04-06)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- changed: (TON) Replace Orbs servers with dRPC
+
 ## 4.78.0 (2026-04-06)
 
 - added: (Zano) Add makeMaxSpend otherMethod to build single-transaction multi-asset max spends.
--- a/src/polkadot/PolkadotEngine.ts
+++ b/src/polkadot/PolkadotEngine.ts
@@ -44,7 +44,6 @@ import {
   asSafePolkadotWalletInfo,
   asSubscanResponse,
   asTransactions,
-  asTransfer,
   LiberlandTransfer,
   PolkadotNetworkInfo,
   PolkadotWalletOtherData,
@@ -536,15 +535,9 @@ export class PolkadotEngine extends CurrencyEngine<
 
       // Process txs (newest first)
       transfers.forEach(tx => {
-        try {
-          this.processPolkadotTransaction(asTransfer(tx), isActiveChain)
-        } catch (e: any) {
-          const hash = tx != null && typeof tx.hash === 'string' ? tx.hash : ''
-          this.warn(`Ignoring invalid transfer ${hash}`)
-        }
+        if (tx == null) return
+        this.processPolkadotTransaction(tx, isActiveChain)
       })
-
-      if (count === this.otherData.subscanUrlMap[subscanBaseUrl].txCount) break
 
       // If we haven't reached the end, Update local txCount and progress and then query the next page
       this.otherData.subscanUrlMap[subscanBaseUrl].txCount =
--- a/src/polkadot/PolkadotEngine.ts
+++ b/src/polkadot/PolkadotEngine.ts
@@ -39,12 +39,14 @@ import { PolkadotTools } from './PolkadotTools'
 import {
   asLiberlandMeritsResponse,
   asLiberlandTransfersResponse,
+  asPolkadotInitOptions,
   asPolkadotWalletOtherData,
   asPolkapolkadotPrivateKeys,
   asSafePolkadotWalletInfo,
   asSubscanResponse,
   asTransactions,
   LiberlandTransfer,
+  PolkadotInitOptions,
   PolkadotNetworkInfo,
   PolkadotWalletOtherData,
   SafePolkadotWalletInfo,
@@ -64,6 +66,7 @@ export class PolkadotEngine extends CurrencyEngine<
   TokenSyncTracker
 > {
   engineFetch: EdgeFetchFunction
+  initOptions: PolkadotInitOptions
   networkInfo: PolkadotNetworkInfo
   otherData!: PolkadotWalletOtherData
   api!: ApiPromise
@@ -78,6 +81,7 @@ export class PolkadotEngine extends CurrencyEngine<
   ) {
     super(env, tools, walletInfo, opts, makeTokenSyncTracker)
     this.engineFetch = makeEngineFetch(env.io)
+    this.initOptions = asPolkadotInitOptions(env.initOptions)
     this.networkInfo = env.networkInfo
     this.nonce = 0
   }
@@ -130,16 +134,20 @@ export class PolkadotEngine extends CurrencyEngine<
     endpoint: string,
     body: JsonObject
   ): Promise<SubscanResponse> {
+    const headers: Record<string, string> = {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    }
+    if (this.initOptions.subscanApiKey !== '') {
+      headers['X-API-Key'] = this.initOptions.subscanApiKey
+    }
     const options = {
       method: 'POST',
-      headers: {
-        Accept: 'application/json',
-        'Content-Type': 'application/json'
-      },
+      headers,
       body: JSON.stringify(body)
     }
     const response = await this.engineFetch(baseSubscanUrl + endpoint, options)
-    if (!response.ok || response.status === 429) {
+    if (!response.ok) {
       throw new Error(`Subscan ${endpoint} failed with ${response.status}`)
     }
     const out = await response.json()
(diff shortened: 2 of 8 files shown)


src/polkadot/PolkadotEngine.ts
```diff
@@ -584,7 +585,10 @@ export class PolkadotEngine extends CurrencyEngine<
         TRANSACTION_POLL_MILLISECONDS
       )
     }
-    if (this.networkInfo.subscanBaseUrls.length > 0) {
+    if (
+      this.networkInfo.subscanBaseUrls.length > 0 &&
+      this.initOptions.subscanApiKey !== ''
```
samholmes, line 590: This condition disables transaction history syncing whenever `subscanApiKey` is not provided. The changelog says we are *adding* API-key support, but this change effectively makes history sync opt-in only for keyed deployments. Existing DOT users without `initOptions.subscanApiKey` will now report history sync complete (`setHistoryRatios(..., 1)`) with no fetched transactions.\n\nConsider still running `queryTransactions` without the header and only adding `X-API-Key` when configured, so behavior remains backward-compatible.

src/ton/TonEngine.ts
```diff
@@ -376,13 +382,19 @@ export class TonEngine extends CurrencyEngine<
...
+    const addr = this.wallet.address.toRawString()
+    const seqnoRaw = await this.tools.fetchDrpc('/runGetMethod', {
+      address: addr,
+      method: 'seqno',
+      stack: []
     })
-    const seqno: Awaited<ReturnType<typeof this.wallet['getSeqno']>> =
-      await promiseAny(seqnoFuncs)
+    const { result: seqnoResult } = asDrpcRunGetMethod(seqnoRaw)
+    if (seqnoResult.exit_code !== 0) {
```
samholmes: This hard-fails first-send flows for undeployed TON wallets. `runGetMethod(seqno)` commonly returns a non-zero exit code (for example `-13` on uninitialized/non-active accounts), but wallet seqno should be treated as `0` in that case so deployment transfer can proceed.\n\nI verified toncenter currently returns `exit_code: -13` for undeployed addresses, so this branch throws before signing and blocks activation transactions.

src/ton/TonEngine.ts
```diff
@@ -404,53 +416,46 @@ export class TonEngine extends CurrencyEngine<
...
-            edgeTransaction.date = Date.now() / 1000
-            this.otherData.contractState = 'active'
-            this.walletLocalDataDirty = true
-            return edgeTransaction
+          try {
+            const raw = await this.tools.fetchDrpc(
+              `/getTransactions?address=${addr}&limit=1`
+            )
+            const { result: txs } = asDrpcGetTransactions(raw)
+            if (txs.length > 0) {
```
samholmes: `txs.length > 0` is not a safe completion condition here. For an uninitialized wallet, there is often already at least one prior inbound funding transaction before the first outgoing send. This loop will then immediately pick that pre-existing tx and assign the wrong `txid` to the broadcasted transaction.\n\nPlease filter for the newly broadcast transaction (or at least compare against previous latest tx) before returning.

src/ton/tonDrpc.ts
```diff
@@ -0,0 +1,124 @@
...
+    throw new Error(`dRPC ${path}: ${String(json.error ?? 'request failed')}`)
+  }
+  return json
+}
+
+// ---------------------------------------------------------------------------
+// Response cleaners
+// ---------------------------------------------------------------------------
+
+const asPassthrough = (raw: unknown): unknown => raw
```
samholmes: I believe there is an `asUnknown` cleaner. Let's use that.

samholmes (changes requested):

src/polkadot/PolkadotEngine.ts
peachbits: subscan requires an apikey to access their service. It was previously limited by ip rate limits. No apikey = no transaction history.

src/ton/TonEngine.ts
peachbits: that's ok, uninitialized wallets don't have any money to spend anyway.

src/polkadot/PolkadotEngine.ts
```diff
@@ -584,7 +585,10 @@ export class PolkadotEngine extends CurrencyEngine<
...
+    ) {
       this.addToLoop('queryTransactions', TRANSACTION_POLL_MILLISECONDS)
     } else {
       this.syncTracker.setHistoryRatios(this.enabledTokenIds, 1)
```
samholmes: **Bug:** `enabledTokenIds` does not include `null` (the native token id). The `SyncTracker` internally iterates over `[null, ...engine.enabledTokenIds]`, so native history stays at 0% when Subscan is skipped.

Other engines (Algorand, Ripple, Solana, etc.) explicitly include `null`:

```suggestion
      this.syncTracker.setHistoryRatios([null, ...this.enabledTokenIds], 1)
```

OPTIONAL FOLLOW UP because this is unrelated to the PR.

src/ton/tonDrpc.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+  const entry = stack[0]
+  if (!Array.isArray(entry) || entry.length < 2) {
+    throw new Error('Unexpected seqno stack format')
+  }
+  const [type, value] = entry
+  if (type !== 'num') {
+    throw new Error(`Unexpected seqno stack type: ${String(type)}`)
+  }
+  return Number(value)
```
samholmes: **Bug:** `Number(value)` silently returns `NaN` for non-numeric strings (empty string, malformed hex, etc.) and flows into `seqno` for transaction signing.

```suggestion
  const n = Number(value)
  if (!Number.isFinite(n)) {
    throw new Error(`Invalid seqno value: ${String(value)}`)
  }
  return n
```

src/polkadot/PolkadotEngine.ts
```diff
@@ -536,16 +543,10 @@ export class PolkadotEngine extends CurrencyEngine<
...
       transfers.forEach(tx => {
-        try {
-          this.processPolkadotTransaction(asTransfer(tx), isActiveChain)
-        } catch (e: any) {
-          const hash = tx != null && typeof tx.hash === 'string' ? tx.hash : ''
-          this.warn(`Ignoring invalid transfer ${hash}`)
-        }
+        if (tx == null) return
+        this.processPolkadotTransaction(tx, isActiveChain)
       })
```
samholmes, line 548: **Nit:** The old code wrapped each transfer in try/catch so one bad row wouldn't abort the entire history pass. With `asMaybe(asTransfer)` handling parse failures, the null check is sufficient for those — but any runtime throw inside `processPolkadotTransaction` (e.g. biggystring edge case) would now propagate uncaught and abort the whole `queryTransactionsInner` call. Consider keeping a narrow try/catch here for resilience.

src/ton/TonTools.ts
```diff
@@ -100,15 +97,23 @@ export class TonTools implements EdgeCurrencyTools {
...
+  }
+
+  async fetchDrpc(path: string, body?: object): Promise<unknown> {
+    const funcs: Array<() => Promise<unknown>> = [
+      async () =>
+        await fetchDrpcRequest(this.io, this.networkInfo.drpcUrl, path, body)
+    ]
+
+    const { drpcApiKey } = this.initOptions
+    if (drpcApiKey != null) {
```
samholmes: **Nit:** `asOptional(asString)` can produce `''`. The condition `drpcApiKey != null` passes for empty strings, producing the URL `https://lb.drpc.org/rest//ton`. Consider matching the Polkadot pattern:

```suggestion
    if (drpcApiKey != null && drpcApiKey !== '') {
```

src/ton/TonEngine.ts
```diff
@@ -404,53 +416,46 @@ export class TonEngine extends CurrencyEngine<
...
+              const hashBytes = base64.parse(txs[0].transaction_id.hash)
+              edgeTransaction.txid = base16.stringify(hashBytes).toLowerCase()
+              edgeTransaction.date = Date.now() / 1000
+              this.otherData.contractState = 'active'
+              this.walletLocalDataDirty = true
+              return edgeTransaction
+            }
+          } catch (e) {
+            // retry
           }
```
samholmes, line 466: **Nit:** Every failure in the poll loop (network, HTTP, cleaner throws) is silently swallowed. After 30 failed attempts the user only sees `Transaction broadcast but unable to find txid` with no indication of the root cause. Consider logging the last error at warn level or including it in the final thrown error.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- added: (DOT) Add subscan apikey support
```
samholmes: **Nit:** `apikey` → `API key` for consistency with English conventions and the commit message.

```suggestion
- added: (DOT) Add Subscan API key support
```

samholmes (changes requested):

src/ton/tonDrpc.ts
samholmes: This one seems worthwhile.

samholmes (approved):

PR #1045, opened 2026-04-04
feat(zano): add makeMaxSpend wallet other method

samholmes:
Motivation: allow GUI migration to request one max-spend transaction across selected Zano assets instead of composing multiple spends. The method accepts selected token IDs and defaults to all assets when none are provided, so callers can opt into explicit asset control.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---

> [!NOTE]
> **Medium Risk**
> Adds a new Zano transaction-construction path that builds multi-asset max-spend transfers and changes amount parsing to enforce JS safe-integer limits, which could surface new runtime errors or edge-case balance/fee behavior.
> 
> **Overview**
> Adds a new Zano `otherMethods.makeMaxSpend` that constructs a **single multi-asset max-spend** `EdgeTransaction` by sweeping selected token balances (and optionally native ZANO minus fee) into one transfer list.
> 
> Introduces `safeParseInt` and switches Zano transfer amount conversion to use it, throwing if a string amount cannot be represented as a JS safe integer (affecting both `makeSpend` and the new max-spend path).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 130735f19254c9882de50356928570531ea028c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1213968564434713

### DIFF
--- a/src/zano/ZanoEngine.ts
+++ b/src/zano/ZanoEngine.ts
@@ -15,8 +15,11 @@ import {
   JsonObject,
   NoAmountSpecifiedError
 } from 'edge-core-js/types'
-import type { RecentTransaction, TransferParams } from 'react-native-zano'
-import { CppBridge } from 'react-native-zano'
+import type {
+  CppBridge,
+  RecentTransaction,
+  TransferParams
+} from 'react-native-zano'
 
 import { CurrencyEngine } from '../common/CurrencyEngine'
 import { PluginEnvironment } from '../common/innerPlugin'
@@ -35,6 +38,7 @@ import {
   asZanoPrivateKeys,
   asZanoTransferParams,
   asZanoWalletOtherData,
+  MakeMaxSpendParams,
   SafeZanoWalletInfo,
   ZanoNetworkInfo,
   ZanoOtherMethods,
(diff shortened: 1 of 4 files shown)


src/zano/ZanoEngine.ts
```diff
@@ -597,6 +597,78 @@ export class ZanoEngine extends CurrencyEngine<
...
   otherMethods: ZanoOtherMethods = {
     resolveName: this.resolveName.bind(this),
+    makeMaxSpend: async (
+      params: MakeMaxSpendParams
+    ): Promise<EdgeTransaction> => {
+      const { metadata } = params
+      const publicAddress = params.spendTargets[0]?.publicAddress
+      if (publicAddress == null) throw new Error('Missing publicAddress')
+
+      const tokenIdsSet = new Set<EdgeTokenId>(
```
peachbits: I think it's strange to use a undefined params.tokenIds to mean "spend everything you have". Let's make params.tokenIds required

src/zano/ZanoEngine.ts
```diff
@@ -597,6 +597,78 @@ export class ZanoEngine extends CurrencyEngine<
...
+            nativeAmount: parseInt(zanoSendAmount),
+            recipient: publicAddress
+          })
+        }
+      }
+
+      for (const tokenId of tokenIds) {
+        if (tokenId == null) continue
+        const tokenBalance = this.unlockedBalanceMap.get(tokenId) ?? '0'
+        if (eq(tokenBalance, '0')) continue
```
peachbits: Throw an error

peachbits (changes requested):

src/zano/ZanoEngine.ts
samholmes: Hm, so how does the caller max spend all tokens that have a balance? They'd need to include all tokens that are supported by the wallet. Combine the "throw error if balance is zero" to the spec, and now the caller would need to loop through all tokens and check balance for each token before running makeMaxSpend. Is this what we want? I'm okay with it, but just pointing out the extra steps the caller would need to do. Maybe we can refactor or change later (this isn't a hardened API).

peachbits (approved):

PR #1047, opened 2026-04-10
Fix missing incoming ZEC transaction notifications

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---

> [!NOTE]
> **Medium Risk**
> Adjusts new-transaction detection/notification logic for Zcash by special-casing unconfirmed incoming transactions, which could change when/if users see notifications (including potential false positives or duplicates, especially across devices).
> 
> **Overview**
> Fixes missing incoming ZEC transaction notifications by treating unconfirmed incoming transactions (`blockHeight === 0`, `confirmations === 'unconfirmed'`) as *new* even when a `seenTxCheckpoint` already exists.
> 
> To support this override, `CurrencyEngine.isTransactionNew` is loosened from `private` to `protected`, and Zcash adds an engine-specific `isTransactionNew` implementation that falls back to the base checkpoint logic for all other cases. Also adds a corresponding Unreleased changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3aab354d37d012ad28753f2eb8e790d6e6b06f00. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1213629974235942

### DIFF
--- a/src/common/CurrencyEngine.ts
+++ b/src/common/CurrencyEngine.ts
@@ -289,7 +289,8 @@ export class CurrencyEngine<
     return out
   }
 
-  private isTransactionNew(edgeTransaction: EdgeTransaction): boolean {
+  // This method is protected because it's used in the ZcashEngine to handle unconfirmed txs. It should revert to private once unconfirmed txs are properly handled in the seenTxCheckpoint API
+  protected isTransactionNew(edgeTransaction: EdgeTransaction): boolean {
     const txCheckpoint = this.getTxCheckpoint(edgeTransaction)
     if (
       // Undefined seenTxCheckpoint means we're syncing for the very first time
--- a/src/zcash/ZcashEngine.ts
+++ b/src/zcash/ZcashEngine.ts
@@ -171,6 +171,19 @@ export class ZcashEngine extends CurrencyEngine<
     })
   }
 
+  // This is a partial fix for handling unconfirmed txs in the seenTxCheckpoint API. It solves the problem for a single device, but multi-devices syncing the same account may still see incorrect notifications.
+  isTransactionNew(edgeTransaction: EdgeTransaction): boolean {
+    if (
+      edgeTransaction.blockHeight === 0 &&
+      edgeTransaction.confirmations === 'unconfirmed' &&
+      this.seenTxCheckpoint != null &&
+      !edgeTransaction.isSend
+    ) {
+      return true
+    }
+    return super.isTransactionNew(edgeTransaction)
+  }
+
   isSynced(): boolean {
     // Synchronizer status is updated regularly and should be checked before accessing the db to avoid errors
     return this.synchronizerStatus === 'SYNCED'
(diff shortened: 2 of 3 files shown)


src/zcash/ZcashEngine.ts
```diff
@@ -171,6 +171,18 @@ export class ZcashEngine extends CurrencyEngine<
     })
   }
 
+  isTransactionNew(edgeTransaction: EdgeTransaction): boolean {
+    if (
```
samholmes, line 176: Add a comment saying that this is a partial intermediary fix and is good enough for now. Mention how it solves the problem for a single device, but multi-devices syncing the same account may still see incorrect notifications.

src/common/CurrencyEngine.ts
```diff
@@ -289,7 +289,7 @@ export class CurrencyEngine<
     return out
   }
 
-  private isTransactionNew(edgeTransaction: EdgeTransaction): boolean {
+  protected isTransactionNew(edgeTransaction: EdgeTransaction): boolean {
```
samholmes, line 293: Put a caveat that protected is a tempory change and it should go back to private once we've spec'd how to handle unconfirmed txs in the seenTxCheckpoint API feature/thing.

samholmes (approved):

src/zcash/ZcashEngine.ts
```diff
@@ -171,6 +171,19 @@ export class ZcashEngine extends CurrencyEngine<
     })
   }
 
+  // This is a partial fix for handling unconfirmed txs in the seenTxCheckpoint API. It solves the problem for a single device, but multi-devices syncing the same account may still see incorrect notifications.
```
bobbythelobster, line 174: **Suggestion (cosmetic):** Consider adding the `override` keyword for explicitness. The project doesn't enforce `noImplicitOverride`, so this isn't required, but it documents intent and protects against future base-class renames:

```typescript
override isTransactionNew(edgeTransaction: EdgeTransaction): boolean {
```

bobbythelobster (approved):
## Review Summary

**Low risk.** Small, well-scoped change that correctly fixes missing ZEC incoming-transaction notifications.

### Key observations

- The override logic is sound: it only fires for unconfirmed receives (`blockHeight === 0`, `!isSend`) when a `seenTxCheckpoint` already exists, avoiding false positives during initial sync.
- **No duplicate notification risk:** `addTransaction`'s txid-based dedup prevents the same unconfirmed tx from being marked `isNew: true` on re-emit — subsequent calls hit the update branch which always sets `isNew: false`.
- **No missed notification risk:** When the tx later confirms (`blockHeight > 0`), the override falls through to `super.isTransactionNew()` which handles the confirmed case via checkpoint comparison.
- The `"0"` checkpoint from unconfirmed txs never regresses `highestSeenCheckpoint` since `selectSeenTxCheckpoint` picks the numerically larger value.
- The `private` → `protected` visibility change is the right extensibility pattern.

See inline comment for one cosmetic suggestion.

PR #1055, opened 2026-06-05
Reimplement Pirate Chain plugin over react-native-pirate-wallet

j0ntz:
### Technical Design Document

[piratechain-unified-wallet-sdk.md](https://github.com/EdgeApp/edge-currency-accountbased/blob/agent/1214721783909451/src/docs/piratechain-unified-wallet-sdk.md)

, committed on this branch. It holds the design, every decision with its rejected alternatives, the phase history, and the test evidence. Read it instead of a long description here.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Asana: https://app.asana.com/0/1215088146871429/1214721783909451

Reimplements the Pirate Chain plugin over `react-native-pirate-wallet` (npm `0.3.4`), the Pirate team's replacement for the `react-native-piratechain` module Edge wrote. In the old module the Swift and Rust layers each opened the wallet's SQLite database and the app crashed when they overlapped, which is why unrelated work has to set `piratechain: false` in the GUI's `corePlugins.ts`. The new SDK keeps all database access in Rust, and this branch retires that workaround.

The plugin's public shape and the engine's transaction mapping are unchanged. What is new:

- One device-scoped SDK wallet registry holds every ARRR wallet, keyed by wallet-id alias. The native module mints the registry credential and keeps it in the iOS Keychain or Android Keystore; it never reaches JavaScript. Each wallet's signing keys are wrapped under a second credential derived from the mnemonic; protection is enabled the first time a wallet is seen and the key is unlocked only for the span of a send, so a registry at rest can sync but cannot sign.
- A send is refused, with the reason, until the SDK reports the wallet spendable; the two retryable broadcast errors map to the same wait-and-retry message.
- Sync comes from the SDK's polling synchronizer and history from `list_transactions`. Amounts cross the wire as decimal strings, and sends go through the SDK's `send()`.
- Each wallet is pointed at Edge's own lightwalletd node with `set_lightd_endpoint`. Without it a wallet scans against the SDK's baked-in default, which no longer answers at all, and the app sits at 0% sync with no error anywhere. That node is now `https://lightwalletd2.cryptoforge.cc:443`: TLS instead of plain gRPC, and measurably faster under `0.3.2` (below).
- `getLatestNetworkHeight` takes the chain tip from a registered wallet or from `test_node`, falling back to the create-and-delete probe only when the registry is empty. Running that probe while a synchronizer is live panics the native service and aborts the app.
- A wallet takes a pool of lightwalletd endpoints rather than one, through `set_lightd_endpoint_pool`, and it now ships with an alternate (`lightwalletd1.cryptoforge.cc`) rather than empty. The SDK validates a pool and rejects one that names an unrecognized host or mixes transports; a rejected pool falls back to the single endpoint rather than failing the wallet, which both rejections were observed doing on device. The info server can still replace the list without an app release.

The plugin compiles against the package's own `index.d.ts`: `react-native-pirate-wallet` is a devDependency like the other native SDKs, the hand-written ambient declaration is gone, `create-hmac` is a declared dependency with `@types/create-hmac`, and the endpoint calls go through the typed `testLightdEndpoint`, `setLightdEndpoint` and `setLightdEndpointPool`; raw `invoke` remains for `set_tunnel` only. The bridge exposes `start` and the engine calls it after subscribing, so the first status change is never missed. `getMaxSpendable` refuses for the same reason `makeSpend` does, and a transaction whose processing throws is retried on the next poll.

Paired with EdgeApp/edge-react-gui#6021, which swaps the GUI dependency and has to land after this one publishes.

Verified on the iOS simulator, with the mocha suite at 490 passing.

A controlled cold-sync A/B settled whether to ship the `0.3.2` bump. One wallet 1,056,918 blocks behind, every other ARRR wallet archived, the same simulator and session, each side forced cold and confirmed cold by watching the block cache grow from zero, and the SDK swapped by replacing `libpirate_ffi_native.a` and relinking so the static archive was the only variable:

| SDK | node | cold time to 100% |
|---|---|---|
| `0.3.0` | `http://lightd1.pirate.black:9067` | 72 s |
| `0.3.2` | `http://lightd1.pirate.black:9067` | ~30 min projected (15.86% in 5.6 min) |
| `0.3.2` | `https://lightwalletd2.cryptoforge.cc:443` | 95 s |

The first two rows read as a 20-40x regression in `0.3.2`, which was the working verdict until the third row was measured. `0.3.2` is not slow; the pair of `0.3.2` and `lightd1.pirate.black` is. So this branch ships the bump and moves the endpoint, which also retires the plaintext transport the plugin's own comment had flagged as an exposure.

The plugin also gates spending on the SDK's `get_spendability_status`. A wallet reports `SYNCED` before its spend anchor is usable, and a send in that window used to build, get confirmed by the user, and only then fail inside the SDK, which the app rendered as "check your network connection". `makeSpend` now reads the status and throws `PendingFundsError` while the wallet cannot spend, so the send scene holds no transaction and its confirm slider stays disabled. A status the plugin cannot read falls through to the old `SYNCED`-only gate rather than blocking the send. A wallet that has not reached `SYNCED` is refused with the same `PendingFundsError`, so the amount modal and the send scene's error card show one message for every wait.

Testing showed the status is necessary and not sufficient, so `broadcastTx` restates `ERR_SYNC_FINALIZING` and `ERR_WITNESS_REPAIR_QUEUED` the same way: on 2026-08-31 a wallet reading `spendable: true, repairQueued: false, reasonCode: OK` with a validated anchor still had four broadcasts rejected by the node on anchor grounds, and the status did not move afterwards. No principal or fee was lost. Section 10 of the doc carries that back to the Pirate Chain team as the open question.

Both configurations then broadcast a funded ARRR send to the transaction-success scene: `0.3.0` at txid `d4bc2294a427e9c0f197c5ccc21d4ac2687d28fc5f4f3a718f77c757bc130585`, and the shipping configuration at txid `239ab8743afbed4f8047ed4e0b05d2aa3e39ebfa7ab9824ddfa52c72c47c4205`. Section 7 of the doc has the full matrix, the node survey, the SDK's pool-validation rules, and what stayed uncovered (automatic failover under a stalling primary, which the SDK's recognized-host rule makes untestable with a local proxy). Section 10 is a report for the Pirate Chain team.

---

> [!NOTE]
> **High Risk**
> This replaces the entire ARRR native stack and changes registry credentials, signing-key lifecycle, sync endpoints, and send/broadcast paths—core wallet and funds-handling behavior.
> 
> **Overview**
> **Pirate Chain (ARRR) moves off the crash-prone `react-native-piratechain` module onto the Pirate team's unified `react-native-pirate-wallet` (^0.3.4).** The engine and IO bridge are rebuilt around that SDK while keeping the plugin's outward API the same.
> 
> Sync and network behavior now **force Edge's TLS lightwalletd** (`lightwalletd2.cryptoforge.cc`) with an optional **failover pool** (`lightwalletdFailoverUrls`, overridable from the info server). Wallets use a **single device registry** (`configureSecureAccountStorage` / Keychain or Keystore) plus **signing protection** unlocked only during `send` via an HMAC-derived credential from the mnemonic.
> 
> **Spending** is gated on SDK spendability (`get_spendability_status`) and pre-sync checks throw **`PendingFundsError`** with wait-and-retry copy; **`broadcastTx`** maps `ERR_SYNC_FINALIZING` / `ERR_WITNESS_REPAIR_QUEUED` the same way. Sends use the SDK **`send()`** path with **string amounts**; history comes from the poller's full list with per-tx cleaning. **`start()`** runs only after subscriptions attach, and chain-tip lookup avoids mutating a live registry (empty-registry probe only).
> 
> Dependencies add **`create-hmac`**; peer/dev dependency swaps **`react-native-pirate-wallet`**. A large **`piratechain-unified-wallet-sdk.md`** design doc and **CHANGELOG** entries document the migration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 057ad6337f3f75928934704850816dc04e5fc73a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1216969318856768

### Test evidence

<table>
<tr>
<td rowspan="7" width="170" valign="top"><a href="https://github.com/EdgeApp/edge-currency-accountbased/pull/1055/commits/390ef737ede5f07ce14ff23b0aa7d1ace31a216f"><code>390ef73</code></a><br><sub>Rebuild Pirate Chain on react-native-pirate-wallet</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260729-174013-agent-proof-1216926437132721-01-arrr-my-pirate2-synced.png"></a><br><sub>arrr my pirate2 synced</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260729-174013-agent-proof-1216926437132721-02-arrr-send-dest-zaddr.png"></a><br><sub>arrr send dest zaddr</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260729-174013-agent-proof-1216926437132721-03-arrr-send-confirm.png"></a><br><sub>arrr send confirm</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260729-174013-agent-proof-1216926437132721-04-arrr-send-success.png"></a><br><sub>arrr send success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260729-174013-agent-proof-1216926437132721-05-arrr-send-txid.png"></a><br><sub>arrr send txid</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260811-202617-agent-proof-1216926437132721-01-arrr-synced.png"></a><br><sub>arrr synced</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260811-202617-agent-proof-1216926437132721-02-arrr-send-ready.png"></a><br><sub>arrr send ready</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260811-202617-agent-proof-1216926437132721-03-arrr-sync-stalled.png"></a><br><sub>arrr sync stalled</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260817-145657-agent-proof-1216926437132721-01-arrr-synced-0.3.0.png"></a><br><sub>arrr synced 0.3.0</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260817-145657-agent-proof-1216926437132721-02-arrr-send-ready.png"></a><br><sub>arrr send ready</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260817-145657-agent-proof-1216926437132721-03-arrr-send-success.png"></a><br><sub>arrr send success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260817-145657-agent-proof-1216926437132721-04-arrr-balance-after-send.png"></a><br><sub>arrr balance after send</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260825-162603-agent-proof-1216926437132721-00-arrr-synced-0.3.2-tls-pool.png"></a><br><sub>arrr synced 0.3.2 tls pool</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260825-162603-agent-proof-1216926437132721-01-arrr-synced-0.3.0.png"></a><br><sub>arrr synced 0.3.0</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260825-162603-agent-proof-1216926437132721-02-arrr-send-ready.png"></a><br><sub>arrr send ready</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260825-162603-agent-proof-1216926437132721-03-arrr-send-success.png"></a><br><sub>arrr send success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260825-162603-agent-proof-1216926437132721-04-arrr-txid.png"></a><br><sub>arrr txid</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260825-162603-agent-proof-1216926437132721-05-arrr-send-success-0.3.2-tls.png"></a><br><sub>arrr send success 0.3.2 tls</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260825-162603-agent-proof-1216926437132721-06-arrr-txid-0.3.2-tls.png"></a><br><sub>arrr txid 0.3.2 tls</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260916-005453-agent-proof-1216926437132721-21-synced-max-fee-slider-armed.png"></a><br><sub>synced max fee slider armed</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260916-005453-agent-proof-1216926437132721-22-transaction-history.png"></a><br><sub>transaction history</sub></td>
</tr>
<tr>
<td rowspan="2" width="170" valign="top"><a href="https://github.com/EdgeApp/edge-currency-accountbased/pull/1055/commits/8118f87c6179dfe9903140d03918a23e536a7d86"><code>8118f87</code></a><br><sub>Gate ARRR sends on SDK spendability</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260831-134435-agent-proof-1216926437132721-01-before-generic-error.png"></a><br><sub>before generic error</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260831-134435-agent-proof-1216926437132721-02-gate-passed-slider-armed.png"></a><br><sub>gate passed slider armed</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260831-134435-agent-proof-1216926437132721-03-not-spendable-error.png"></a><br><sub>not spendable error</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260916-005440-agent-proof-1216926437132721-20-arrr-sync-in-progress.png"></a><br><sub>arrr sync in progress</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260916-005440-agent-proof-1216926437132721-23-presync-refusal-error-card.png"></a><br><sub>presync refusal error card</sub></td>
<td width="210"></td>
</tr>
<tr>
<td rowspan="4" width="170" valign="top"><a href="https://github.com/EdgeApp/edge-currency-accountbased/pull/1055/commits/428d199efa473d2c92efc7cb517a8eef8794575b"><code>428d199</code></a><br><sub>Log the ARRR endpoint pool diagnostics</sub><br><sub>🪓 <em>forced the bridge's getSpendability in the updot-baked plugin copy to return spendable:false with reasonCode ERR_WITNESS_REPAIR_QUEUED (frames 04, 05) then ERR_SYNC_FINALIZING (frame 06); reverted, git status clean.</em></sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260902-113612-agent-proof-1216926437132721-01-keychain-registry-syncing.png"></a><br><sub>keychain registry syncing</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260902-113612-agent-proof-1216926437132721-02-gate-passed-slider-armed.png"></a><br><sub>gate passed slider armed</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260902-113612-agent-proof-1216926437132721-03-transaction-success.png"></a><br><sub>transaction success</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260902-113612-agent-proof-1216926437132721-04-txid.png"></a><br><sub>txid</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260914-175712-agent-proof-1216926437132721-01-max-fills-spendable-balance.png"></a><br><sub>max fills spendable balance</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260914-175712-agent-proof-1216926437132721-02-send-armed.png"></a><br><sub>send armed</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260914-175712-agent-proof-1216926437132721-03-transaction-success.png"></a><br><sub>transaction success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260914-183854-agent-proof-1216926437132721-04-HACKED-max-refused-witness-repair.png"></a><br><sub>🪓 max refused witness repair</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260914-183854-agent-proof-1216926437132721-05-HACKED-not-spendable-witness-repair.png"></a><br><sub>🪓 not spendable witness repair</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260914-183854-agent-proof-1216926437132721-06-HACKED-not-spendable-sync-finalizing.png"></a><br><sub>🪓 not spendable sync finalizing</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260914-193350-agent-proof-1216926437132721-07-send-armed-fresh-registry.png"></a><br><sub>send armed fresh registry</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1055/20260914-193350-agent-proof-1216926437132721-08-transaction-success-fresh-registry.png"></a><br><sub>transaction success fresh registry</sub></td>
</tr>
</table>


src/piratechain/PiratechainEngine.ts
```diff
@@ -381,42 +324,42 @@ export class PiratechainEngine extends CurrencyEngine<
...
-      }
+      const txid = await synchronizer.send(
+        [
+          {
+            addr: spendTarget.publicAddress,
+            amount: parseInt(spendAmount, 10),
+            memo
+          }
+        ],
+        parseInt(edgeTransaction.networkFee, 10)
```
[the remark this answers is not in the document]
j0ntz: Fixed in c26cdcbb: `broadcastTx` now uses `safeParseInt` for both the spend amount and the fee, matching the zano pattern.

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,179 @@
...
+      })
+      try {
+        const probeWallet = await walletSdk.getWallet(probeWalletId)
+        if (probeWallet == null) {
+          throw new Error('Missing birthday probe wallet')
+        }
+        return probeWallet.birthdayHeight
+      } finally {
+        await walletSdk.deleteWallet(probeWalletId).catch(() => undefined)
+      }
```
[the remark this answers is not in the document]
j0ntz: Fixed in c26cdcbb: the birthday probe now runs under the same registry serialization lock as `ensureWallet`, so concurrent probes (or a probe racing a restore) can no longer interleave registry mutations.

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,179 @@
...
         },
-        onUpdate(event): void {
-          emit(out, 'update', event)
+        onUpdate(snapshot): void {
+          const { progressPercent, syncStatus } = snapshot
+          emit(out, 'update', {
+            lastDownloadedHeight: syncStatus?.localHeight ?? 0,
+            networkBlockHeight: syncStatus?.targetHeight ?? 0,
+            progressPercent
+          })
```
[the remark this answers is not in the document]
j0ntz: Fixed in c26cdcbb: the io bridge now skips update events until the backend reports a positive `targetHeight`, so the sync tracker never sees zero/missing heights.

src/piratechain/PiratechainEngine.ts
```diff
@@ -164,42 +144,10 @@ export class PiratechainEngine extends CurrencyEngine<
...
-          first,
-          last
-        }
-        this.walletLocalDataDirty = true
-
-        if (blocksToHeight > 0) {
-          const historyRatio = (last - this.birthdayHeight) / blocksToHeight
-          this.syncTracker.updateTransactionRatio(historyRatio)
-        }
+      const transactions = await this.synchronizer.getTransactions()
```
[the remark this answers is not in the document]
j0ntz: Fixed in c2a610d1: the engine now tracks the height each txid was last processed at and skips stable entries, so a full-history snapshot only costs O(changed) per update instead of O(n) reprocessing. (The map clears on resync so reorgs/rescans still reprocess.)

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,187 @@
...
+    const task = ensureWalletLock.then(async () => {
+      const { birthdayHeight, mnemonic, name } = config
+      const walletSdk = getSdk()
+      const registryExists = await walletSdk.walletRegistryExists()
+      if (registryExists) {
+        const wallets = await walletSdk.listWallets()
+        const existingWallet = wallets.find(wallet => wallet.name === name)
+        if (existingWallet != null) return existingWallet.id
+      }
+      return await walletSdk.restoreWallet({ name, mnemonic, birthdayHeight })
```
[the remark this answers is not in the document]
j0ntz: Not applicable in practice: Edge wallet keys (mnemonic + birthdayHeight) are immutable once created, and the alias name is derived from the immutable Edge walletId, so a name hit can only be the same seed. A changed birthday can only arrive as a brand-new Edge wallet (new walletId → new alias → fresh restore). Resolving without code change.

j0ntz:
Simulator-tested update (force-pushed as part of the single commit):

- **App passphrase unlock**: the SDK's SQLCipher store rejects every wallet call with `Security error: App is locked` until `set_app_passphrase`/`unlock_app` runs. The io bridge now unlocks (or provisions) the passphrase store before any wallet operation. Verified on the iOS sim: registry restore, viewing-key derivation, and `zs1…` receive address all work.
- **Direct tunnel**: the SDK transports through Tor by default, which never bootstraps inside Edge — sync sat at `stage: Headers, target_height: 0` forever. The bridge now issues `set_tunnel {mode: Direct}` after unlock, matching every other Edge plugin's connectivity. Verified the sync config flips to `transport: Direct` against `http://64.23.167.130:9067` (server healthy; tip 3,990,043 via grpcurl).

**Upstream blocker found while testing**: block sync still cannot progress through this binding. `pirate-ffi-native` builds a throwaway current-thread tokio runtime per `invoke`, and `sync_control::start_sync` spawns the sync engine onto it — the task is destroyed when the call returns. The Flutter app works only because flutter_rust_bridge keeps a persistent runtime; the RN/iOS/Android SDK bindings all share this defect, and the bounded `start_background_sync` API is not exposed through the JSON dispatcher. Needs a Pirate-team fix (persistent global runtime in `pirate-ffi-native`, or expose `start_background_sync` via `invoke`). Until then balances/sends can't be exercised end-to-end.

j0ntz:
**End-to-end sync + send verified on the iOS simulator** (with a locally rebuilt pirate core carrying the one-line persistent-runtime fix):

- Funded wallet (254.547 ARRR) restored from its Edge mnemonic + birthday and fully scanned ~940k blocks to chain tip; SDK balance matched Edge's cached pre-upgrade balance to the arrrtoshi.
- Sent 3.261 ARRR wallet-to-wallet in the same account: txid `b44367658e7c92cd916eccbdee3dc8f6861a48985683037040a8ead2035c9881`, broadcast accepted by `http://64.23.167.130:9067`.

Two more findings, both addressed in the latest force-push:

1. **Wrapper send bug (upstream)**: `sdk.send()`/`signTransaction()` camelize the `build_tx` result and feed it back into `sign_tx`, which rejects it (`Invalid request JSON: missing field 'total_amount'`). The bridge now drives `build_tx → sign_tx → broadcast_tx` over the raw `invoke` channel so the pending/signed payloads round-trip untouched.
2. **Runtime defect confirmed + fix validated**: rebuilding `libpirate_ffi_native` with a persistent global tokio runtime in `WalletService::execute_blocking` (instead of a per-call current-thread runtime) makes foreground sync work through the json-invoke bindings. That's the patch to upstream to the Pirate team — without it the stock release binaries cannot sync outside their Flutter app.

src/piratechain/PiratechainEngine.ts
```diff
@@ -209,41 +169,39 @@ export class PiratechainEngine extends CurrencyEngine<
     }
   }
 
-  processTransaction(tx: ConfirmedTransaction): void {
-    let netNativeAmount = tx.value
+  processTransaction(tx: PirateTransaction): void {
+    // A negative amount is a send and already includes the network fee:
+    const netNativeAmount = String(tx.amount)
```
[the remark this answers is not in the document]
j0ntz: True in principle, but the precision loss is inherent to the SDK's transport, not this code: the wrapper JSON-parses the native envelope, so amounts arrive as JS numbers before this plugin ever sees them — `String(...)` here adds no additional loss and there is no local way to recover it. Exposure requires a single wallet holding >~90M ARRR (45% of ARRR's total supply). The real fix is upstream (string amounts in the JSON API), which we'll raise with the Pirate team alongside the runtime and sign_tx issues. Resolving without code change.

src/piratechain/rnPirateWallet.d.ts
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  export interface SynchronizerCallbacks {
+    onStatusChanged?: (event: {
+      walletId: string
+      alias: string
+      name: SynchronizerStatus
+    }) => void
+    onUpdate?: (snapshot: SynchronizerSnapshot) => void
+    onError?: (error: Error) => void
```
peachbits: `onError?: (error: Error) => void` overstates what the React Native bridge delivers. Native errors are serialized across the bridge and arrive as plain objects or strings, not real `Error` instances. The consumer at `piratechainIo.ts:204` already guards with `error instanceof Error ? error.message : String(error)`, which only makes sense if the value is not guaranteed to be an `Error`. Declaring `error: unknown` here keeps the type honest and the existing guard meaningful.

src/piratechain/PiratechainEngine.ts
```diff
@@ -150,10 +135,10 @@ export class PiratechainEngine extends CurrencyEngine<
...
     if (!this.isSynced() || this.synchronizer == null) return
     try {
-      const balances = await this.synchronizer.getBalance()
-      if (balances.totalZatoshi === '-1') return
-      this.availableZatoshi = balances.availableZatoshi
-      this.updateBalance(null, balances.totalZatoshi)
+      const balance = await this.synchronizer.getBalance()
+      // `total` includes pending; `spendable` is the confirmed balance:
+      this.availableZatoshi = String(balance.spendable)
+      this.updateBalance(null, String(balance.total))
```
peachbits, line 158: `balance.total` and `balance.spendable` arrive from the native `getBalance` RPC as JS `number` arrrtoshi values (`PirateBalance`, `rnPirateWallet.d.ts:45-49`). ARRR has 8 decimals and roughly 200M max supply, so a full balance can reach about 2e16 arrrtoshi, above `Number.MAX_SAFE_INTEGER` (about 9.007e15). For large balances the value can lose precision before `String(...)` hands it to biggystring, and `String(number)` cannot recover it. If the native module can expose these fields as decimal strings, prefer that so downstream biggystring math stays exact. Same applies to the `availableZatoshi` assignment on line 140.

peachbits (commented):
Two non-blocking suggestions on the new `react-native-pirate-wallet` integration, both at the native-bridge boundary. The plugin reimplementation reads cleanly otherwise.

src/piratechain/PiratechainEngine.ts
peachbits: human note: it should be a string, but in real world usage it's not likely that a single wallet will hold close to 50% of the supply

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,264 @@
...
 }
 
+/**
+ * The SDK encrypts its on-device databases (SQLCipher) behind an app
+ * passphrase and rejects every wallet call with "App is locked" until
+ * `set_app_passphrase`/`unlock_app` runs. Edge already gates wallet access
+ * behind its own login, so a fixed passphrase keeps at-rest encryption at
+ * parity with the previous SDK (sandbox-protected files).
+ */
+const APP_PASSPHRASE = 'edge-pirate-wallet'
```
peachbits: This is a red flag. A hardcoded passphrase makes the local registry data insecure on an unlocked device. Ideally multiple registries can exist and be scoped to an account. Another solution would be to guard the sensitive spend keys by a unique passphrase per wallet. I reached out to the piratechain team with this concern.

src/piratechain/rnPirateWallet.d.ts
j0ntz: Fixed in 9daf2a6a. SynchronizerCallbacks.onError is now typed (error: unknown), matching what the serialized RN bridge delivers; the consumer guard `error instanceof Error ? error.message : String(error)` is now meaningful.

src/piratechain/PiratechainEngine.ts
j0ntz: Fixed in 9daf2a6a. The released SDK v1.1.5 serializes amounts as decimal strings (AmountString), so PirateBalance and PirateTransaction amount fields are now string and the engine hands them to biggystring without safeParseInt, preserving precision above 2^53-1.

src/piratechain/piratechainIo.ts
j0ntz: Fixed in 9daf2a6a. The hardcoded shared passphrase is removed. v1.1.5's configureAccountStorage gives each wallet its own registry namespace, unlocked by a passphrase derived (HMAC-SHA256) from that wallet's seed. Verified on the iOS sim: three ARRR wallets each created an isolated PirateWallet/accounts/<id> registry with its own wallet_registry.db and salt.

j0ntz:
## 📸 Test evidence: e2e ARRR send (iOS sim)

**arrr my pirate2 synced**

arrr my pirate2 synced

**arrr send dest zaddr**

arrr send dest zaddr

**arrr send confirm**

arrr send confirm

**arrr send success**

arrr send success

**arrr send txid**

arrr send txid

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,274 @@
...
+          // amounts as strings so large sends keep full precision:
+          return await walletSdk.send(walletId, outputs, fee ?? null)
         },
         stop: async () => {
-          return await realSynchronizer.stop()
+          await realSynchronizer.close()
         }
       })
 
+      await realSynchronizer.start()
```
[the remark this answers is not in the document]
j0ntz: Examined and judged non-blocking in practice, with a hardening follow-up.

Why it does not drop SYNCED in the current flow: the SDK synchronizer's start() emits only the STOPPED->SYNCING transition synchronously, then defers the first poll via _schedule(0) (setTimeout). The engine attaches its yaob listeners in initSubscriptions() synchronously immediately after awaiting makeSynchronizer, so the first _refreshOnce (and thus any SYNCED statusChanged) fires on a later tick, after the listeners are in place. This is corroborated end to end: this run drove the wallet to SYNCED and broadcast a real send (txid 34ba68b0...e234e49cd) on this exact startup path.

Hardening follow-up (deferred to the post-rebase review rather than churned now, since #1055 needs a rebase onto master and a fresh review): expose the synchronizer's current status from the bridge (e.g. getStatus) and have the engine read it once right after initSubscriptions as a backstop, so a missed initial event cannot strand synchronizerStatus at STOPPED regardless of tick ordering.

j0ntz:
## 📸 Test evidence

**arrr synced**

arrr synced

**arrr send ready**

arrr send ready

**arrr sync stalled**

arrr sync stalled

src/piratechain/PiratechainEngine.ts
```diff
@@ -256,26 +231,24 @@ export class PiratechainEngine extends CurrencyEngine<
...
       // Replace this.synchronizerPromise with a fresh promise. The old promise might have already been resolved
+      const { defaultHost, defaultPort } = this.networkInfo.rpcNode
       this.synchronizerPromise = this.makeSynchronizer({
-        mnemonicSeed: piratechainPrivateKeys.mnemonic,
+        name: base16.stringify(base64.parse(this.walletId)),
+        mnemonic: piratechainPrivateKeys.mnemonic,
         birthdayHeight: piratechainPrivateKeys.birthdayHeight,
-        alias: base16.stringify(base64.parse(this.walletId)),
-        ...rpcNode
+        lightwalletdUrl: `http://${defaultHost}:${defaultPort}`
```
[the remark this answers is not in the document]
j0ntz: Real exposure, and the plaintext transport stays for now because it is the only one this SDK can complete a handshake on. Its own `test_node` probe, run from the app against the live node:

- `http://lightd1.pirate.black:9067` → `success: true`, `latest_block_height: 4085946`, `tls_enabled: false`
- `https://lightd1.pirate.black:443` → `Connection failed: gRPC connection failed: transport error`

The SDK's built-in default node (`http://64.23.167.130:9067`) is plaintext too, so before this change every wallet was already scanning in the clear, just against a node Edge does not control and which had stopped serving blocks.

I have written the risk out at the config site rather than leaving it implicit: an observer learns which block ranges the device fetches and when it broadcasts, and an active attacker can serve a forked view. The comment names the exit condition, which is a TLS-terminating lightwalletd this SDK can handshake against. That is on the list for the Pirate team alongside the broadcast timeout this run also found.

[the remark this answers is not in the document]
j0ntz: Correct, and fixed in the latest commit. Deriving the URL from `rpcNode` meant an info-server payload carrying the long-shipped `defaultPort: 443` would silently repoint sync at a port the SDK cannot speak, re-breaking exactly what this change fixes.

`networkInfo` now carries `lightwalletdUrl` as a full URL with the scheme, and the engine passes that through untouched. `rpcNode` keeps its historical shape and its `443`, so an old payload no longer has any path to the sync endpoint; a payload can only set the endpoint through the new field, where scheme, host and port travel together and cannot disagree. `asPiratechainInfoPayload` accepts `lightwalletdUrl` as an optional string.

j0ntz:
## 📸 Test evidence: react-native-pirate-wallet 0.3.0 (unified wallet v1.1.6), funded ARRR send

**arrr synced 0.3.0**

arrr synced 0.3.0

**arrr send ready**

arrr send ready

**arrr send success**

arrr send success

**arrr balance after send**

arrr balance after send

package.json
```diff
@@ -187,7 +186,7 @@
   },
   "peerDependencies": {
     "react-native-monero": "^0.3.0",
-    "react-native-piratechain": "v0.5.0",
+    "react-native-pirate-wallet": "^0.3.0",
```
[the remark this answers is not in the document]
j0ntz: Fixed in 983e6b07. `npm install --package-lock-only` regenerated the root `packages[""]` peerDependencies block, so the lockfile now records `^0.3.0` alongside package.json.

Worth noting the failure mode you describe did not actually fire: a fresh `git clone` of this branch plus `npm ci` and `npm run prepare` exited 0 at the previous head, so npm did not refuse the divergent root peer range. The lockfile was still wrong and a later `npm install` would have rewritten it, so the fix stands.

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,323 @@
...
+              return syncStatus.targetHeight
+            }
+          }
+        }
+
+        // Nothing in the registry to ask, so no synchronizer can be running
+        // either, and mutating it is safe. `create_wallet` with no birthday
+        // resolves the height from the lightwalletd tip, falling back to the
+        // SDK's static checkpoint:
+        const probeWalletId = await walletSdk.createWallet({
```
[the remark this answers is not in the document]
j0ntz: Fixed in 70ea0d97. `getLatestNetworkHeight` now takes the plugin's configured `lightwalletdUrl` and, when no registered wallet reports a tip, reads the height with `test_node` against that node instead of creating a probe wallet. `test_node` returns `latest_block_height` without registering anything, so it is also safe while synchronizers are running.

The create-and-delete probe survives only when the registry is genuinely empty, and it is now gated on `wallets.length === 0` rather than on the tip loop falling through. Your inflated-birthday concern is the reason the configured node is preferred over the SDK default even when the default is reachable.

Worth noting the SDK default is currently not reachable at all: `test_node` against `http://64.23.167.130:9067` returns `success: false` with a gRPC deadline error, while `http://lightd1.pirate.black:9067` returns `latest_block_height: 4093998`.

src/piratechain/piratechainDeviceStorage.ts
```diff
@@ -0,0 +1,68 @@
...
+    try {
+      return asDevicePassphraseFile(text).passphrase
+    } catch (error: unknown) {
+      // Unreadable contents: fall through and re-mint. The wallets in the
+      // abandoned registry restore themselves from their seeds.
+    }
+  }
+
+  const passphrase = base16.stringify(io.random(32))
+  await disklet.setText(DEVICE_PASSPHRASE_FILE, JSON.stringify({ passphrase }))
```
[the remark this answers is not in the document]
j0ntz: Valid, and accepted rather than fixed. Recording the reasoning here and in the design doc (decision 3) so it is not invisible later.

The constraint chain: `configure_wallet_storage` selects the SDK's registry namespace **device-globally**, and switching it cancels running sync and clears the registry and block caches. So Edge runs exactly one device-scoped namespace holding every ARRR wallet (decision 1). A device-scoped registry cannot be keyed to an Edge account secret, and the passphrase has to be readable before any account is unlocked for wallets to resume on launch. That rules out the account key lockbox, a password-derived key, and the seed-derived HMAC an earlier phase used.

The residual you describe is real and I am not narrowing it away: a reader of the app sandbox can unwrap the registry and spend ARRR without the Edge password, which is a weaker bar than the mnemonic enjoys. What bounds it is the OS sandbox, and destroying the file costs a rescan from birthday rather than funds.

Reopen trigger, recorded in the doc: the SDK gaining per-account or per-call storage credentials, or a way to hold the passphrase in memory for a session.

src/piratechain/PiratechainEngine.ts
```diff
@@ -164,42 +166,17 @@ export class PiratechainEngine extends CurrencyEngine<
...
-          this.syncTracker.updateTransactionRatio(historyRatio)
-        }
+      const transactions = await this.synchronizer.getTransactions()
+      for (const tx of transactions) {
+        // The SDK returns the full history each time, so only process
+        // transactions that are new or have moved (confirmed/reorged):
+        const height = tx.height ?? 0
+        if (this.processedTxHeights.get(tx.txid) === height) continue
+        this.processTransaction(tx)
+        this.processedTxHeights.set(tx.txid, height)
```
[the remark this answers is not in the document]
j0ntz: Correct, and half-fixed in 89dd8d2e. You are right that the cache detects the change and `addTransaction` then discards it: the shared engine writes a height only when it moved FORWARD (or on a same-height field change), so a reorg that unmines a transaction or re-mines it lower cannot land, and the Edge record keeps the stale height.

What changed: `queryTransactions` now recognises a backwards move and logs `Reorged height for <txid> not applied: <old> -> <new>` instead of calling `processTransaction` for an update that the base layer will throw away. The limitation is now visible in logs rather than silent, and the wasted reprocess is gone.

What did NOT change, deliberately: the underlying rule lives in `src/common/CurrencyEngine.ts` `addTransaction`, which every plugin shares. Relaxing it to accept a decrease is the real fix, but it changes reorg handling for every currency in this repo and needs its own testing pass, so it does not belong in an ARRR SDK-swap PR. Filed as a follow-up.

Residual, so it is on the record: after a reorg an ARRR transaction's confirmation height is overstated by the reorg depth until it re-confirms above the stale value, and a transaction dropped back to the mempool keeps a confirmed height. Reopen when the shared `addTransaction` gains a downgrade path.

src/piratechain/piratechainIo.ts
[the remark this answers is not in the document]
OswaldKardingson: @j0ntz the endpoint 64.23.167.130 has been retired, you can use lightd1.pirate.black, lightwalletd1.cryptoforge.cc, or 
lightwalletd2.cryptoforge.cc

j0ntz:
## 📸 Test evidence: cold-sync A/B and two funded sends

**arrr synced 0.3.2 tls pool**

arrr synced 0.3.2 tls pool

**arrr synced 0.3.0**

arrr synced 0.3.0

**arrr send ready**

arrr send ready

**arrr send success**

arrr send success

**arrr txid**

arrr txid

**arrr send success 0.3.2 tls**

arrr send success 0.3.2 tls

**arrr txid 0.3.2 tls**

arrr txid 0.3.2 tls

j0ntz:
## 📸 Test evidence: the spendability gate and the error it now produces

**before generic error**

before generic error

**gate passed slider armed**

gate passed slider armed

**not spendable error**

not spendable error

j0ntz:
## 📸 Test evidence (pirate-wallet 0.3.4: keychain registry, signing session, funded send)

**keychain registry syncing**

keychain registry syncing

**gate passed slider armed**

gate passed slider armed

**transaction success**

transaction success

**txid**

txid

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,487 @@
...
+ * than the spend itself.
+ */
+const asSpendabilityStatus = asObject({
+  spendable: asBoolean,
+  rescanRequired: asOptional(asBoolean, false),
+  repairQueued: asOptional(asBoolean, false),
+  reasonCode: asOptional(asString),
+  anchorHeight: asOptional(asNumber),
+  validatedAnchorHeight: asOptional(asNumber)
+})
```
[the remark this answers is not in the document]
j0ntz: Wrong about the API: cleaners' asOptional is implemented as raw != null ? cleaner(raw) : fallback (node_modules/cleaners/lib/cjs/cleaners.js line 26), so a JSON null anchor is treated as absent and cleans to undefined instead of throwing. Probe on this branch: asObject({ spendable: asBoolean, anchorHeight: asOptional(asNumber) })({ spendable: false, anchorHeight: null }) returns { spendable: false }. getSpendability therefore reads a finalizing status with null anchors, and checkSpendable throws PendingFundsError on spendable: false as intended.

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,487 @@
...
+          await walletSdk.unlockWalletSigning(walletId, signingCredential)
+        }
+      }
+      try {
+        await ensureSigningUnlocked()
+      } catch (error: unknown) {
+        console.warn(
+          `piratechain: wallet signing protection unavailable: ${String(error)}`
+        )
+      }
```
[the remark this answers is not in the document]
j0ntz: Pre-existing intentional design, recorded as decision 6 and the accepted risk in src/docs/piratechain-unified-wallet-sdk.md: a signing failure at start logs and lets sync proceed, because the registry credential is what sync needs and a wallet that never syncs is worse than one that cannot sign yet. The premise that restoreWallet never enables protection is not the plugin's shape: enableWalletSigningProtection runs in makeSynchronizer right after ensureWallet on every start, which is the restore path, and send() calls ensureSigningUnlocked without a guard, so a persistent enable failure refuses the spend with the SDK's own error rather than signing under the registry credential. The plugin has no send path that bypasses that wrapper. Reopen if the SDK exposes a signing call the bridge does not own.

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,487 @@
...
 
+      try {
+        await realSynchronizer.start()
+      } catch (error: unknown) {
+        // The caller never receives `out`, so nothing else can stop the native
+        // poller. Without this a retry would leave a second synchronizer
+        // running against the same registry wallet:
+        await realSynchronizer.close().catch(() => undefined)
+        throw error
+      }
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed in adb679e5 (folded into the signing-lock commit): the start() failure path in makeSynchronizer now calls lockWalletSigning(walletId) after closing the poller, so an unlocked key never outlives a synchronizer the caller never received (src/piratechain/piratechainIo.ts, the catch around realSynchronizer.start()).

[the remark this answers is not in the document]
j0ntz: The pushed head carries this fix at 925a4a6b (the branch was autosquashed after the previous reply).

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,490 @@
...
+          await realSynchronizer.close()
+          // The engine stops when the account locks or logs out, which is
+          // when the signing key must leave memory:
+          await walletSdk
+            .lockWalletSigning(walletId)
+            .catch((error: unknown) => {
+              console.warn(
+                `piratechain: failed to lock wallet signing: ${String(error)}`
+              )
+            })
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed in d398731a (folded into the signing-lock commit): stop now closes the poller inside a try/finally and locks the wallet's signing key in the finally, so a throw from close() still leaves the key locked and the error still propagates (src/piratechain/piratechainIo.ts, the stop closure).

src/piratechain/rnPirateWallet.d.ts
```diff
@@ -0,0 +1,317 @@
+/**
+ * Type declarations for `react-native-pirate-wallet`.
+ *
+ * The package is a peer dependency the GUI installs, so it is absent from
+ * this repo's own node_modules and this file carries the typings needed to
+ * compile against it. Only the surface consumed by the piratechain plugin is
+ * declared here; the shapes follow the package's `src/index.d.ts` at the
+ * version `package.json` requires.
+ */
+declare module 'react-native-pirate-wallet' {
```
peachbits: **The package ships its own types, and this file silently overrides them.**

`react-native-pirate-wallet@0.3.4` publishes `"types": "src/index.d.ts"` — 352 lines covering every surface used here. And an ambient `declare module` takes precedence over `node_modules` resolution, so installing the real package would change nothing: TypeScript keeps checking against this file. Probe against the real 0.3.4 tarball with this file's `PirateBalance` block as the ambient declaration:

```
probe.ts(2,15): error TS2305: Module '"react-native-pirate-wallet"' has no exported member 'Balance'.
```

`PirateBalance` resolves, `Balance` (the name the package actually exports) does not.

The names here are the package's names with a prefix bolted on, which is why nothing catches the drift:

| here | actually exported |
| --- | --- |
| `PirateBalance` | `Balance` |
| `PirateTransaction` | `TransactionInfo` |
| `PirateSpendabilityStatus` | `SpendabilityStatus` |
| `PirateTransactionOutput` | `TransactionOutput` |
| `PirateEndpointPoolDiagnostics` | `EndpointPoolDiagnostics` |
| `PirateFeeInfo` | `FeeInfo` |
| `PirateAccountStorageConfig` | `PirateWalletAccountStorageConfig` |

There is already one error in here: `PirateFeeInfo.defaultFee: number`, where the package declares `AmountString`. It is latent only because `getFeeInfo` is never called — exactly the failure mode this file guarantees, since a wrong shape compiles clean.

The design doc (line 184) says this file "retypes `PirateBalance`, `PirateTransaction`, and `PirateTransactionOutput` amount fields from `number` to `string`, matching v1.1.5's `AmountString`" — but 0.3.4 already declares those as `AmountString`, so the retype is transcribing a fix the package shipped rather than supplying one.

Suggest deleting this file and adding `react-native-pirate-wallet` as a devDependency (see the `package.json` comment) — the two only work together, since the file wins over the installed package either way.

peachbits (commented):
Local review of the head commit. The first two comments (`rnPirateWallet.d.ts` and `package.json`) are one coupled change — the hand-written ambient declaration wins over `node_modules`, so the devDependency only helps once the file is gone. Everything else stands alone. I skipped the `queryMutex` boolean, which cursor already raised at line 123.

package.json
```diff
@@ -187,7 +186,7 @@
   },
   "peerDependencies": {
     "react-native-monero": "^0.3.0",
-    "react-native-piratechain": "v0.5.0",
+    "react-native-pirate-wallet": "^0.3.4",
```
peachbits, line 192: `react-native-pirate-wallet` is added as a peer dependency only. The PR also drops `react-native-piratechain: 0.5.0` from `devDependencies` without replacing it, so the new SDK is never installed in this repo.

Every other native SDK here is in **both** lists — `react-native-zcash`, `react-native-monero` and `react-native-zano` are devDependencies for typechecking and peerDependencies for the consumer, and their plugins import the packages' real types (`src/zcash/zcashIo.ts`, `src/monero/moneroTypes.ts`, `src/zano/zanoTypes.ts`).

Adding the devDependency is half the fix; `src/piratechain/rnPirateWallet.d.ts` has to go at the same time or it shadows the installed package's types.

src/declare-modules.d.ts
```diff
@@ -89,3 +89,15 @@ declare module 'bs58' {
   export const encode: (buffer: Buffer | number[] | Uint8Array) => string
   export const decode: (string: string) => Buffer
 }
+
+declare module 'create-hmac' {
```
peachbits: `create-hmac` is imported at `PiratechainEngine.ts:2` but is not a declared dependency. The `overrides` entry pins a version, it does not declare one — `npm ls create-hmac` shows it arriving only transitively:

```
+-- @fioprotocol/fiosdk@1.10.3 -> @fioprotocol/fiojs -> create-hmac@1.1.7
+-- @taquito/signer@23.1.0 -> pbkdf2 -> create-hmac@1.1.7
+-- crypto-browserify@3.12.1 -> create-hmac@1.1.7 (overridden)
+-- ed25519-hd-key@1.3.0 -> create-hmac@1.1.7
...
```

It resolves today purely because npm hoists it to the top level. The day one of those drops it or nests it, the webpack bundle breaks at build time with nothing in this repo explaining why.

Two changes worth making together:
- add `create-hmac` to `dependencies` so the resolution is owned rather than inherited
- use `@types/create-hmac` (`^1.1.3`, on DefinitelyTyped) as a devDependency instead of this hand-rolled block

The hand-rolled `Hmac` here also narrows `algorithm` to `'sha256' | 'sha512'` and `digest()` to the no-arg overload, so it will need editing for any future caller that wants a different digest or `digest('hex')`.

src/piratechain/PiratechainEngine.ts
```diff
@@ -328,8 +347,39 @@ export class PiratechainEngine extends CurrencyEngine<
...
-  async makeSpend(edgeSpendInfoIn: EdgeSpendInfo): Promise<EdgeTransaction> {
+  /**
+   * The SDK reports `SYNCED` before its spend anchor is usable, and a send in
+   * that window fails inside the SDK with `ERR_SYNC_FINALIZING` only once the
+   * user has already confirmed it. `get_spendability_status` reports the
+   * window directly, so refuse the spend here instead: a `makeSpend` that
+   * throws leaves the send scene with no transaction, which is what keeps its
+   * confirm slider disabled until the wallet can actually spend.
+   */
+  async checkSpendable(): Promise<void> {
```
peachbits, line 352: `checkSpendable` gates `makeSpend` but not `getMaxSpendable`, and the two disagree before the wallet is synced.

`initData()` sets `availableZatoshi = '0'`, and `queryBalance` returns early unless `isSynced()`. So on an unsynced wallet `getMaxSpendable` computes `sub('0', defaultNetworkFee)`, which is negative, and throws `InsufficientFundsError` — a funded wallet tells the user they have no money, where `makeSpend` would correctly raise `PendingFundsError` with a wait-and-retry message.

`await this.checkSpendable()` at the top of `getMaxSpendable` makes both paths report the same reason. It also covers the spend-anchor window this method was written for: today tapping Max succeeds during `ERR_SYNC_FINALIZING` and the refusal only arrives at `makeSpend`.

src/piratechain/PiratechainEngine.ts
```diff
@@ -164,42 +172,30 @@ export class PiratechainEngine extends CurrencyEngine<
...
-          const historyRatio = (last - this.birthdayHeight) / blocksToHeight
-          this.syncTracker.updateTransactionRatio(historyRatio)
+      const transactions = await this.synchronizer.getTransactions()
+      for (const tx of transactions) {
+        // The SDK returns the full history each time, so only process
+        // transactions that are new or whose height moved:
+        const height = tx.height ?? 0
+        const seenHeight = this.processedTxHeights.get(tx.txid)
+        if (seenHeight === height) continue
+        this.processedTxHeights.set(tx.txid, height)
```
peachbits: The cache is written before the transaction is processed, so a throw in `processTransaction` loses the transaction permanently.

`processTransaction` reaches `gte(netNativeAmount, '0')`, and biggystring throws on a malformed amount. The `catch` is outside the loop, so the pass aborts — but the txid has already been recorded at its current height, so every later poll hits `seenHeight === height` and skips it. Nothing retries it short of a resync clearing the map.

Moving the `set` to after `processTransaction(tx)` returns makes a failed entry retry on the next poll instead:

```ts
if (seenHeight != null && height < seenHeight) {
  this.warn(`Reorged height for ${tx.txid} not applied: ${seenHeight} -> ${height}`)
  this.processedTxHeights.set(tx.txid, height)
  continue
}
this.processTransaction(tx)
this.processedTxHeights.set(tx.txid, height)
```

Separately, on the comment's "or whose height moved" reasoning: `addTransaction` also updates at an unchanged height when `networkFee`, `nativeAmount`, `date` or `otherParams.lastSeenTime` differ, and this skip blocks that path — a pending transaction reaches `addTransaction` once rather than on each poll. I could not find a field the SDK actually revises at a fixed height, and nothing in this engine depends on `lastSeenTime` being refreshed, so this is a note on the comment rather than a defect.

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,494 @@
...
+          await walletSdk.enableWalletSigningProtection(
+            walletId,
+            signingCredential
+          )
+        } else if (!status.unlocked) {
+          await walletSdk.unlockWalletSigning(walletId, signingCredential)
+        }
+      }
+      try {
+        await ensureSigningUnlocked()
```
peachbits: This unlock looks unnecessary, and it widens the window the rest of this design works to keep narrow.

The comment above it says sync does not depend on signing ("a locked registry can still sync ... but cannot sign"), and `send` already calls `ensureSigningUnlocked()` itself at line 456 for exactly the reason given there — another wallet's `lockAllWalletSigning` can have cleared this key since start. So the startup unlock buys nothing the send path does not re-establish, while leaving the signing key in native memory for the entire life of the engine.

Scoping it to the spend would be strictly tighter:

```ts
send: async (outputs, fee) => {
  await ensureSigningUnlocked()
  try {
    return await walletSdk.send(walletId, outputs, fee ?? null)
  } finally {
    await walletSdk.lockWalletSigning(walletId).catch(() => undefined)
  }
}
```

That also subsumes the two failure paths already fixed in this PR (the `start()` catch and the `stop` finally), since there would be nothing unlocked outside a send to leak. The CHANGELOG line "Signing keys stay locked whenever a wallet's engine is not running" could then read "except while a send is in flight".

src/piratechain/PiratechainEngine.ts
```diff
@@ -111,22 +108,33 @@ export class PiratechainEngine extends CurrencyEngine<
...
-        this.lastUpdateFromSynchronizer = undefined
-        await this.startEngine()
-      }
     })
+
+    // A status change that fired before these subscriptions existed is lost,
+    // which would strand the engine at STOPPED and block every spend. Read the
+    // status once and adopt it if no event has arrived yet:
+    synchronizer
+      .getStatus()
```
peachbits: This works — I traced the interleavings and the `!== 'STOPPED'` guard holds in both orders — but it is a sentinel standing in for a start that happens too early.

The root cause is on the io side: `makeSynchronizer` calls `realSynchronizer.start()` itself (`piratechainIo.ts:481`) and only then resolves, so the engine cannot attach its yaob listeners until after the synchronizer is already running and emitting.

Exposing `start` on the bridged handle alongside `stop` and having `syncNetwork` call it after `initSubscriptions()` removes the race rather than detecting it — and with it this round trip, the sentinel, and the "adopt it if no event has arrived yet" reasoning. The `catch` around `start()` that closes the poller and locks signing stays where it is; only the call site moves.

src/piratechain/piratechainInfo.ts
```diff
@@ -13,11 +13,28 @@ import {
 const networkInfo: PiratechainNetworkInfo = {
   rpcNode: {
     networkName: 'mainnet',
-    defaultHost: 'lightd1.pirate.black',
+    defaultHost: 'lightwalletd2.cryptoforge.cc',
```
peachbits: This field is dead now — nothing under `src/piratechain/` reads `networkInfo.rpcNode` any more (`getNewWalletBirthdayBlockheight` was its last caller and now takes `lightwalletdUrl`).

Changing `defaultHost` here makes it look live, which works against `piratechainTypes.ts`'s own "Unused by the unified SDK (endpoints live in the native core); kept for info-server payload compatibility". Worth reverting this line to `lightd1.pirate.black` so the retained-for-compatibility shape stays frozen at what shipped, and the next person editing an info-server payload does not read the fresh hostname as the one in use.

src/piratechain/piratechainIo.ts
```diff
@@ -1,84 +1,494 @@
...
+          }
+        }
+
+        // 2. Ask Edge's own node. `test_node` reports the tip without
+        // registering anything, so it is safe while synchronizers run, and it
+        // reads the configured node rather than the SDK's default (which is
+        // unreachable, and whose tip we would not want to trust anyway: a
+        // birthday taken too high leaves earlier notes unscanned).
+        if (lightwalletdUrl != null) {
+          const result = await invokeCall('test_node', {
```
peachbits: `test_node`, `set_lightd_endpoint` (line 364) and `set_lightd_endpoint_pool` (line 343) all have typed wrappers in 0.3.4 — `testLightdEndpoint`, `setLightdEndpoint`, `setLightdEndpointPool` — so `invokeCall` is only really needed for `set_tunnel`.

The cost of the raw path is visible right here: `PirateWalletSdk._call` camelizes results, while `invoke` does not, which is why `asTestNodeResult` has to clean snake_case `latest_block_height` where the typed `testLightdEndpoint` returns `NodeTestResult.latestBlockHeight`. The wrappers also carry the argument validation (`requireNonEmptyString`, `validateFailoverEndpoints`) and the `tls_pin_opt` naming this code re-derives by hand.

Not blocking, and it is moot while `rnPirateWallet.d.ts` hides those methods from the compiler — but it is one more thing that resolves itself if the real types come in as a devDependency.

peachbits (changes requested):

src/piratechain/rnPirateWallet.d.ts
j0ntz: Fixed in 5896ed9e: rnPirateWallet.d.ts is deleted and react-native-pirate-wallet 0.3.4 is a devDependency like the other native SDKs, so the plugin now compiles against the package's own index.d.ts (Balance, TransactionInfo, EndpointPoolDiagnostics, SpendabilityStatus). The poller's snapshot fields the package leaves as any (balance, syncStatus, transactions) go through cleaners before the engine reads them, and the design doc's amounts section describes the new arrangement.

package.json
j0ntz: Fixed in 5896ed9e together with the declaration file: react-native-pirate-wallet is pinned at 0.3.4 in devDependencies (exact, matching how monero and zcash are listed) and stays ^0.3.4 as a peer dependency; package-lock records the five binary optional packages.

src/declare-modules.d.ts
j0ntz: Fixed in 5896ed9e (dependency) and 98fa3414 (declaration): create-hmac is a direct dependency at 1.1.7, the exact version the existing overrides entry pins, since npm refuses a caret spec against an exact override (EOVERRIDE), and @types/create-hmac 1.1.3 replaces the hand-rolled block in declare-modules.d.ts.

src/piratechain/PiratechainEngine.ts
j0ntz: Fixed in e4e55a89: getMaxSpendable awaits checkSpendable first, so an unsynced or finalizing wallet is refused for the same reason makeSpend refuses it. Verified on the sim that a spendable wallet still fills Max with the balance less the fee and arms the slider; the refusing branch was not reachable this round because the restored wallet reported SYNCED before the send scene opened.

src/piratechain/PiratechainEngine.ts
j0ntz: Fixed in 5896ed9e as you wrote it: the seen height is recorded after processTransaction returns (and on the reorg branch after the warning), so a throw retries on the next poll instead of being skipped until a resync. Agreed on the comment note; the fixed-height field refresh has no consumer in this engine, so the skip stays.

src/piratechain/piratechainIo.ts
j0ntz: Fixed in 98fa3414, with one difference from the snippet: the bridge still enables protection the first time a wallet is seen (so a wallet that never sends has its spending keys wrapped at rest, decision 6 in the design doc) but locks it again at once, and send unlocks, sends, and locks in a finally, so the key is in native memory only for the span of a send. The lock in the start() catch is gone since nothing is unlocked at start; the lock in stop stays for a send in flight when the engine is killed. The CHANGELOG line now reads "Signing keys stay locked except while a send is in flight". Verified on the sim with a funded self-send on this build (txid 0fc26e01781020bb994164f31526a4d21c9a8434e6bdafcd6f1ec4bacb9a8abe).

src/piratechain/PiratechainEngine.ts
j0ntz: Fixed in 5896ed9e: the bridged handle exposes start, makeSynchronizer no longer starts the poller, and syncNetwork calls start() after initSubscriptions, so the first status change has a listener; the getStatus method, the adoption block and the STOPPED sentinel are gone. The close-on-failure stays inside the bridged start. The synchronizer lifecycle section of the design doc describes the new order.

src/piratechain/piratechainInfo.ts
j0ntz: Fixed in 5896ed9e: rpcNode.defaultHost is back to lightd1.pirate.black, so the retained-for-compatibility shape stays frozen at what shipped.

src/piratechain/piratechainIo.ts
j0ntz: Fixed in 5896ed9e: test_node, set_lightd_endpoint and set_lightd_endpoint_pool go through testLightdEndpoint, setLightdEndpoint and setLightdEndpointPool, and the snake_case cleaner for test_node is gone; invoke remains for set_tunnel only. The raw pool call had been sending automatic_failover: true, which the typed wrapper drops, and the service's set_lightd_endpoint_pool handler takes no such argument and derives the flag from a non-empty alternates list, so nothing changed on the wire. Verified on the sim through the pool diagnostics log: automaticFailover true, and with the primary returning 502 that evening the wallet synced over the alternate.

j0ntz:
## 📸 Test evidence (after the 2026-09-11 review fixes)

**max fills spendable balance**

max fills spendable balance

**send armed**

send armed

**transaction success**

transaction success

src/piratechain/piratechainIo.ts
```diff
@@ -1,81 +1,536 @@
...
+      const ensureSigningUnlocked = async (): Promise<void> => {
+        const status = await walletSdk.getWalletSigningStatus(walletId)
+        if (!status.protectionEnabled) {
+          await walletSdk.enableWalletSigningProtection(
+            walletId,
+            signingCredential
+          )
+        } else if (!status.unlocked) {
+          await walletSdk.unlockWalletSigning(walletId, signingCredential)
+        }
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed in b282f9c8 (folded into the signing-lock commit): ensureSigningUnlocked now reads the status enableWalletSigningProtection returns and unlocks when that reports the key locked, instead of returning after the enable; the start-time path already handled the same case.

j0ntz:
## 📸🪓 Test evidence (forced not-spendable states)

> 🪓 **Hack-forced evidence:** forced the bridge's getSpendability in the updot-baked plugin copy to return spendable:false with reasonCode ERR_WITNESS_REPAIR_QUEUED (frames 04, 05) then ERR_SYNC_FINALIZING (frame 06); reverted, git status clean. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: max refused witness repair**

🪓 HACK-FORCED: max refused witness repair

**🪓 HACK-FORCED: not spendable witness repair**

🪓 HACK-FORCED: not spendable witness repair

**🪓 HACK-FORCED: not spendable sync finalizing**

🪓 HACK-FORCED: not spendable sync finalizing

j0ntz:
## 📸 Test evidence (send on the unlock fix, protection enabled from scratch)

**send armed fresh registry**

send armed fresh registry

**transaction success fresh registry**

transaction success fresh registry

peachbits (commented):
Re-review at `359221e5`. All nine earlier threads are fixed as described; `tsc` and `eslint` pass with types now resolved from `node_modules/react-native-pirate-wallet/src/index.d.ts` and `@types/create-hmac`. Two new comments on code the fixes introduced — the first also corrects a premise in my earlier `getMaxSpendable` comment.

src/piratechain/PiratechainEngine.ts
```diff
@@ -328,8 +341,39 @@ export class PiratechainEngine extends CurrencyEngine<
...
+  /**
+   * The SDK reports `SYNCED` before its spend anchor is usable, and a send in
+   * that window fails inside the SDK with `ERR_SYNC_FINALIZING` only once the
+   * user has already confirmed it. `get_spendability_status` reports the
+   * window directly, so refuse the spend here instead: a `makeSpend` that
+   * throws leaves the send scene with no transaction, which is what keeps its
+   * confirm slider disabled until the wallet can actually spend.
+   */
+  async checkSpendable(): Promise<void> {
     if (!this.isSynced()) throw new Error('Cannot spend until wallet is synced')
```
peachbits: Before sync this throws a plain `Error`, and edge-react-gui#6021's `SendScene2` shows that two different ways. Its catch around `getMaxSpendable` / `makeSpend`:

- passes `error.message` to the amount modal, so tapping Max shows "Cannot spend until wallet is synced" inline, which is correct
- passes the error itself to the scene's `ErrorCard`, which renders the message only for an `I18nError`. A plain `Error` gets the generic card instead: "An unexpected error occurred. Please check your network connection and try again later.", with a Report Error button

So once the modal closes, a wallet that is simply syncing is presented as a network failure worth reporting. `getMaxSpendable` used to throw `InsufficientFundsError` here, which the scene maps to a specific insufficient-funds card — the wrong reason, but not a network failure. Typing an amount already hit the generic card, since this guard is on master.

Throwing `PendingFundsError` puts both surfaces on the same text, because the scene already converts it to an `I18nError` carrying the plugin's message:

```ts
if (!this.isSynced()) {
  throw new PendingFundsError('Cannot spend until the wallet finishes syncing')
}
```

Same wording as `spendabilityMessage`'s fallback, and nothing to change on the GUI side.

This also corrects my earlier `getMaxSpendable` comment, which said `makeSpend` would raise `PendingFundsError` before sync. It only does once the wallet reports `SYNCED` but cannot spend yet.

src/piratechain/piratechainIo.ts
```diff
@@ -1,81 +1,539 @@
...
+            await walletSdk.getSpendabilityStatus(walletId)
+          )
+        },
+        getTransactions: async () => {
+          // Same cache as the balance, refreshed by the poller each round:
+          return (
+            asMaybe(asTransactions)(realSynchronizer.transactions) ??
+            (await walletSdk.listTransactions(walletId, null))
+          )
         },
```
peachbits, line 507: The fallback can only hand back what the cleaner just rejected. `realSynchronizer.transactions` starts as `[]`, which cleans fine, so `asMaybe` yields `undefined` only when a populated list fails the shape. The fallback then repeats the poller's own `listTransactions(walletId, null)` and returns that copy uncleaned. The bad row still reaches `processTransaction`, and every `getTransactions` pays a second full-history fetch for as long as that row exists.

`asArray` is also all-or-nothing, so one malformed entry rejects the whole history rather than itself.

Cleaning per entry keeps the good rows and surfaces the bad one, with no RPC needed — the cache is `[]` before the first poll, and the engine only queries after the poller has filled it:

```ts
getTransactions: async () => {
  const raw: unknown[] = realSynchronizer.transactions
  const transactions = raw.flatMap(entry => {
    const tx = asMaybe(asTransactionInfo)(entry)
    return tx == null ? [] : [tx]
  })
  if (transactions.length < raw.length) {
    console.warn(
      `piratechain: dropped ${raw.length - transactions.length} malformed transactions`
    )
  }
  return transactions
},
```

with `asTransactionInfo` being the element cleaner currently inlined in `asTransactions`. The balance fallback is different and fine: `balance` starts `null`, so its RPC covers the window before the first poll.

src/piratechain/PiratechainEngine.ts
j0ntz: Fixed in "Gate ARRR sends on SDK spendability": the pre-sync branch of `checkSpendable` now throws `PendingFundsError` with "Cannot spend until the wallet finishes syncing", the same text as `spendabilityMessage`'s fallback. On the sim, a wallet mid-resync showed that text inline in the amount modal and on the send scene's Transaction Failure card, where it used to fall through to the generic network card.

src/piratechain/piratechainIo.ts
j0ntz: Fixed in "Rebuild Pirate Chain on react-native-pirate-wallet": the element cleaner is now `asTransactionInfo`, `getTransactions` cleans each cached entry, drops a malformed row with a warning that counts the dropped rows, and the `listTransactions` fallback is gone. On the sim the wallet's transaction list rendered all five of its rows through the new path.

j0ntz:
Both new threads are fixed and folded into the commits that own them: the pre-sync refusal throws `PendingFundsError` with the plugin's syncing text, and the transaction snapshot is cleaned per entry with no RPC fallback. The design doc's phase history records the round, and the sim evidence is attached below.

src/piratechain/PiratechainEngine.ts
j0ntz: Correction: this change is in fixup commit fa8ea18525db331cf38656b1fef8881867fc9bea, not folded into "Gate ARRR sends on SDK spendability" as the reply above says. The earlier push folded it by mistake while your review was open.

src/piratechain/piratechainIo.ts
j0ntz: Correction: this change is in fixup commit 74f202a56f3770b1fa15bf16ad5ccd275f01d1af, together with the design doc update, not folded into "Rebuild Pirate Chain on react-native-pirate-wallet" as the reply above says.

j0ntz:
- fa8ea18525db331cf38656b1fef8881867fc9bea: pre-sync refusal throws `PendingFundsError`
- 74f202a56f3770b1fa15bf16ad5ccd275f01d1af: per-entry transaction cleaning, plus the design doc update

peachbits (approved):

PR #1059, opened 2026-06-15
Fix TON duplicate transactions

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Fixes TON duplicate transactions ([Asana task](https://app.asana.com/0/1215088146871429/1208857053925054)).

A sent TON transaction appeared twice in the wallet history: the original send stayed Pending indefinitely while a separate confirmed "Sent Tonchain" entry showed up once the transaction landed on-chain.

Root cause: the broadcast path and the history-sync path stored the transaction id differently, so the two records never reconciled into one.

- `queryTransactions` stores `parse_tx(tx).hash`, which is the `@ton/ton` transaction hash as base64 (`tx.hash().toString('base64')`).
- `broadcastTx` instead stored a locally derived id as lowercase hex: the external-message hash for an active wallet, or a hex re-encoding of a Toncenter hash for a first-send deploy. Both differ from the base64 encoding `queryTransactions` uses, and the external-message hash differs in the underlying value too. (Toncenter's REST `transaction_id.hash` is also a different hash value than `@ton/ton`'s `tx.hash()`, so re-encoding it is not enough either.)

The pending send (from `broadcastTx`) and the confirmed send (from `queryTransactions`) therefore carried different txid strings, and core kept them as two separate entries.

Fix: after `sendBoc`, `broadcastTx` now polls the same `TonClient.getTransactions` path `queryTransactions` uses, finds the confirmed transaction whose external-in message body matches the cell we broadcast, and stores `txid = base64(tx.hash())` with `date = tx.now`. Because the id is produced by the same library call `queryTransactions` records, the two are byte-identical by construction, so the pending send reconciles into a single confirmed entry. Contract-deploy and active-wallet sends share this one path.

### Testing

Drove a real wallet-to-wallet TON send on the iOS simulator (`edge-funds`, My Toncoin to My Toncoin 2, 0.0064 TON) with the patched plugin baked into the core WebView bundle, to the Transaction Success scene. Before/after in the same wallet history: every send made before this fix produced the duplicate (a hex `unconfirmed` row plus a base64 `confirmed` row), and the send made with the fix reconciled to a single base64 `confirmed` row. Verified against the wallet's `transactionList.json` that the post-fix send is one entry whose txid matches what `queryTransactions` stores. `tsc --noEmit`, eslint, and the mocha suite pass. Proof screenshots and the supporting data table are in the test-evidence comment below.

---

> [!NOTE]
> **Medium Risk**
> Changes transaction identity and first-send broadcast polling for TON only; incorrect hashing could still split or mis-link history, but scope is limited to the TON engine.
> 
> **Overview**
> Fixes **duplicate TON send entries** (pending broadcast row plus a separate confirmed history row) by aligning how `broadcastTx` and `queryTransactions` set `EdgeTransaction.txid`.
> 
> Outgoing transactions now use the **external in-message hash** (lowercase hex from the signed message cell via `getInMessageHash`), which matches what explorers use and what history sync records for sends. **Receives** still use the on-chain transaction hash.
> 
> For an **already-deployed wallet**, `broadcastTx` sets that hash **locally** from the external message BOC with no extra lookup. On the **first send** (contract deploy), it **polls `TonClient.getTransactions`** (same path as history sync) until the deploy’s external-in body matches the signed cell, then keys on that in-message hash—replacing the old dRPC `getTransactions` deploy detection. Unused **`asDrpcGetTransactions`** cleaners are removed from `tonDrpc.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4d83f0cea27c9be6cfff9d318221e2af08e9d24. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fixed: (TON) Prevent duplicate transactions by keying a sent transaction on its external in-message hash. The broadcast and the transaction-list sync derive the same hash, so the pending send reconciles with the confirmed transaction, and the hash resolves on the block explorer. An already-deployed wallet derives this locally with no network lookup; only a wallet's very first send looks the hash up once, after the contract deploys.
+
 ## 4.84.1 (2026-06-23)
 
 - fixed: (Stellar) Resolve the "undefined is not an object (evaluating 'Horizon')" crash on login by importing stellar-sdk v13 symbols (Horizon, Keypair, Account, TransactionBuilder, etc.) as named exports. The plugin's WebView uses stellar-sdk's browser build, whose default export is undefined, so the previous default-namespace access (stellarApi.Horizon.Server) threw.
--- a/src/ton/TonEngine.ts
+++ b/src/ton/TonEngine.ts
@@ -7,6 +7,7 @@ import {
   fromNano,
   internal,
   loadMessageRelaxed,
+  Message,
   MessageRelaxed,
   SendMode,
   storeMessage,
@@ -41,7 +42,6 @@ import { snooze } from '../common/utils'
 import {
   asDrpcAddressInfo,
   asDrpcEstimateFee,
-  asDrpcGetTransactions,
   asDrpcRunGetMethod,
   asDrpcSendBoc,
   parseSeqnoFromStack
@@ -156,7 +156,11 @@ export class TonEngine extends CurrencyEngine<
 
           try {
             const parsedTx = asParsedTx(parse_tx(tx, false))
-            this.processTonTransaction(parsedTx)
+            const inMessageHash =
+              tx.inMessage != null
+                ? this.getInMessageHash(tx.inMessage)
+                : undefined
+            this.processTonTransaction(parsedTx, inMessageHash)
           } catch (e) {
             // unknown transaction type
           }
@@ -175,7 +179,20 @@ export class TonEngine extends CurrencyEngine<
     this.syncTracker.updateHistoryRatio(null, 1)
   }
 
-  processTonTransaction(tx: ParsedTx): void {
+  /**
+   * Hash of an in-message as the chain indexes it: the message cell hash, in
+   * lowercase hex. For our own sends the in-message is the external message we
+   * signed, so `broadcastTx` can derive the identical hash from the message it
+   * builds. This is the value the block explorer resolves at `/tx/<hash>`, and
+   * the value a pending send shares with its confirmed copy so the two
+   * reconcile into a single transaction.
+   */
+  getInMessageHash(message: Message): string {
+    const cell = beginCell().store(storeMessage(message)).endCell()
+    return base16.stringify(cell.hash()).toLowerCase()
+  }
+
+  processTonTransaction(tx: ParsedTx, inMessageHash?: string): void {
     const memos: EdgeMemo[] = []
     const ourReceiveAddresses = new Set<string>()
     let nativeAmount: string = '0'
@@ -224,7 +241,14 @@ export class TonEngine extends CurrencyEngine<
       memos,
       ourReceiveAddresses: [...ourReceiveAddresses],
       tokenId: null,
-      txid: tx.hash,
+      // For our own sends, key the transaction on the external in-message hash.
+      // `broadcastTx` derives the identical hash locally for an already-deployed
+      // wallet (and looks it up once for the very first send, which deploys the
+      // contract), so the pending send reconciles into this single confirmed
+      // entry. It is also the hash the block explorer resolves. Received
+      // transactions have no locally-created counterpart, so they keep the
+      // on-chain transaction hash.
+      txid: isSend && inMessageHash != null ? inMessageHash : tx.hash,
       signedTx: '',
       walletId: this.walletId
     }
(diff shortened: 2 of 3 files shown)


peachbits (changes requested):
This isn't correct. The switch to use toncenter is a bad choice because the free service is rate limited and this change leans on it during broadcast so the caller is waiting up to 30 seconds for every spend. And we don't have to do this for every spend because we can derive the tx hash for all transactions created by the wallet except for the very first transaction created by the wallet. That's why the extra transaction fetching logic was gated by contractState. I suspect something broke in this PR https://github.com/EdgeApp/edge-currency-accountbased/pull/1044

j0ntz:
@peachbits reworked again to address the block-explorer concern, and verified the exact test you described in-app.

### What changed since your review

You were right that the in-message **body** hash an earlier revision stored is not what the explorer resolves. The fix now keys a sent transaction on the external in-message **hash** (the full message cell hash, `beginCell().store(storeMessage(message)).endCell().hash()`, lowercase hex), which is precisely what toncenter indexes and what tonscan resolves at `/tx/<hash>`.

- Already-deployed wallet: derived locally in `broadcastTx` from the external message it builds (`externalBoc.hash()`), no network call.
- Very first send (deploys the contract): cannot be derived locally because the external message carries a StateInit, so `broadcastTx` polls once for the confirmed deploy and keys on its on-chain in-message hash. This is the one case that still needs the fetch, gated by `contractState`, exactly as you noted.
- `queryTransactions` records the identical hash for our own sends (`getInMessageHash(tx.inMessage)`), so pending and confirmed reconcile to one entry. Received transactions keep the on-chain transaction hash.

The dead REST `getTransactions` cleaners are removed since the deploy poll now uses the same `TonClient` path as `queryTransactions`.

### In-app test on iOS sim (your definition of done)

Drove a real send to and from a NEW wallet, with the reworked plugin loaded in the running app, then read each stored `txid` out of the wallet's on-device `transactionList.json` and confirmed every block-explorer link resolves.

- Sent **A (old, has history)** to **B (new wallet)**: `My Toncoin 2` to `My Toncoin 3`.
- After confirmation, sent **B to A**: `My Toncoin 3` to `My Toncoin 2`. This is B's first send, so it deploys B's contract (fee was higher, 0.00098 TON vs 0.00012 TON for the active send, confirming the deploy).

All 4 tx-detail explorer links are functional (the txid saved to the EdgeTransaction opens the transaction on tonscan):

| Direction | Wallet view | Stored txid | Explorer |
|---|---|---|---|
| A to B | A send | `5c4375970171b4b2ce64d5e55ca8afed101b17d036e16a24209ecb7d6f5a8ef6` | [tonscan](https://tonscan.org/tx/5c4375970171b4b2ce64d5e55ca8afed101b17d036e16a24209ecb7d6f5a8ef6) |
| A to B | B receive | `296198fc90d45df31ea6908f1c0741e3f4ef4db16c2a112e52bdde8efc171805` | [tonscan](https://tonscan.org/tx/296198fc90d45df31ea6908f1c0741e3f4ef4db16c2a112e52bdde8efc171805) |
| B to A | **B send (new wallet's first tx, deploy)** | `4c5018a99607c2a821d15f61f3b3f1ef46539de1de7a22694b86e13a2686c035` | [tonscan](https://tonscan.org/tx/4c5018a99607c2a821d15f61f3b3f1ef46539de1de7a22694b86e13a2686c035) |
| B to A | A receive | `b48fd782c69fc612a7d4f1ee32993bdd7e6048a9039736358103dbd20740ce2c` | [tonscan](https://tonscan.org/tx/b48fd782c69fc612a7d4f1ee32993bdd7e6048a9039736358103dbd20740ce2c) |

No duplicate transactions: each send added exactly one entry to the sender's history (A went 20 to 21, B went 2 to 3), so the pending send reconciled with the confirmed copy instead of appearing twice. For contrast, a hex-vs-base64 mismatch from an earlier attempt is the kind of double entry this avoids.

For completeness, the underlying hash relationship was also checked against live on-chain data: `getInMessageHash` of an external-in message (including a real deploy message that carries a StateInit) equals toncenter's `in_msg.hash`, which resolves on tonscan; the in-message **body** hash does not.

In-app send-success screenshots are attached below.

j0ntz:
## 📸 In-app test evidence (reworked fix): TON send A(old) to B(new), then B to A (deploy)

**agent proof 1208857053925054 01 ton send A to newB success**

agent proof 1208857053925054 01 ton send A to newB success

**agent proof 1208857053925054 02 txdetail txid resolves**

agent proof 1208857053925054 02 txdetail txid resolves

**agent proof 1208857053925054 03 newB deploy send success**

agent proof 1208857053925054 03 newB deploy send success

j0ntz:
## 📸 No-regression: wallet transaction lists after the in-app test — each send is a single confirmed entry, no duplicate pending tx

**agent proof 1208857053925054 04 txlist newB no duplicate**

agent proof 1208857053925054 04 txlist newB no duplicate

**agent proof 1208857053925054 05 txlist oldA no duplicate**

agent proof 1208857053925054 05 txlist oldA no duplicate

peachbits (approved):

PR #1066, opened 2026-07-02
Monero: show pending transactions, mark dropped, fix bouncing confirmations

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

Requires https://github.com/EdgeApp/react-native-monero/pull/3 (adds the `getPendingTransactions` native method this PR consumes).

### Description

Fixes the Monero receive/confirmation experience found while testing full-node sends against LWS receiving wallets. Three independent commits.

**1. Show pending transactions before their first confirmation** (`723bc2b2`)

The native transaction list sorts pending entries behind all confirmed history, and the engine's history scan stops at the newest known confirmed txid — so it never reached the pending tail, and an incoming transaction was invisible until it confirmed. On LWS (whose balance includes mempool transactions) the balance would move with no transaction row to explain it.

Read the pending set directly via the new `getPendingTransactions` on every pass, outside the cursor protocol. Only confirmed transactions may anchor the scan cursor now (a pending hash recorded as the anchor would match the same tx after it confirms and stop the scan from ever storing its block height), and a cursor already recorded that way by older code is repaired in place by re-processing the anchor row when its stored copy is still unconfirmed. Re-processed pending transactions keep their original `lastSeenTime` so unchanged entries don't emit a change event every poll.

**2. Mark pool-evicted transactions as dropped** (`6d82520d`)

Track when each pool txid was last seen pending; a txid that stays missing for 30 minutes while its stored tx still has no block height was evicted without ever confirming, so mark it dropped (`blockHeight -1` + explicit `'dropped'`, because the base engine clamps negative heights to 0 for non-terminal states). Absence alone is not a drop: a just-mined tx leaves the pool immediately but only appears confirmed once the LWS server indexes its block, so an instant diff misreports every mined tx as dropped during that window (caught on device — observed pending → dropped → confirmed). Also guards `queryTransactions` against concurrent passes.

**3. Clamp the network height monotonic** (`f4f82faa`)

Confirmation counts bounced up and down (observed 4 → 3 → 9 on device) because the base engine re-stamps every tx's count on any height change, and the native height genuinely regresses — lwsf reports the stored account scan height until its first refresh, and a load-balanced daemon can answer a block behind. Feed `updateBlockHeight` the max of the reported and stored heights so counts only advance. Reorg tradeoff (a short-lived cosmetic overcount, never a funds risk) is documented in the commit.

### Testing

`npm run test` (full mocha suite) + `tsc` + ESLint pass. Added `test/monero/MoneroEnginePendingTxs.test.ts` (7 cases) covering pending visibility during backfill, confirmed-only anchoring, poisoned-cursor healing, the grace-period drop, mined-vanish-then-confirm, and no spurious change events. Verified on the iOS simulator: a send between wallets shows a pending row immediately, counts up monotonically, and confirms — no dropped detour or bouncing counts.

---

> [!NOTE]
> **Medium Risk**
> Changes Monero sync cursor, transaction lifecycle, and persisted `otherData`, with a required native dependency bump; logic is heavily tested but mistakes could mislabel or hide transactions.
> 
> **Overview**
> Improves Monero transaction sync so mempool activity matches what users see in balance and notifications.
> 
> **Pending transactions** are ingested on every sync pass via native `getPendingTransactions` (requires **react-native-monero** `0.3.0`), separate from the confirmed-history cursor. Scan anchors and `mostRecentTxid` updates skip pending rows so a pending hash cannot freeze history; older wallets with a poisoned anchor get a one-time heal (and a sweep for stuck rows behind it). Pending rows preserve `lastSeenTime` to avoid spurious update events, and incoming receives still fire **new-transaction** notifications via an override of `isTransactionNew`.
> 
> **Dropped txs** are tracked in wallet `otherData.pendingTxSeen` with a 30-minute grace (with session/app-close resets) before marking `confirmations: 'dropped'`, avoiding false drops while LWS indexes a just-mined tx.
> 
> **Confirmation stability** ignores small network-height regressions (≤1000 blocks) when calling `updateBlockHeight`, while still accepting large regressions as corrections.
> 
> `queryTransactions` is serialized with a mutex, loads transactions on warm start, and `killEngine` drains in-flight passes before resetting state. Adds `MoneroEnginePendingTxs.test.ts` coverage.
> 
> Also bumps the npm **`long`** override to `5.3.1` for Hedera SDK compatibility (per CHANGELOG).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6eb24c09b6ce404e8330f843ea630d3b583a49b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1216221839527388

### DIFF
--- a/src/monero/MoneroEngine.ts
+++ b/src/monero/MoneroEngine.ts
@@ -26,6 +26,7 @@ import {
 import {
   cleanTxLogs,
   makeEngineFetch,
+  makeMutex,
   matchJson,
   normalizeAddress
 } from '../common/utils'
@@ -64,6 +65,25 @@ const ERROR_POLL_MS = 5000 // back off after a sync error
 // and confirmed history) before it is considered evicted from the pool:
 export const DROPPED_TX_GRACE_MS = 30 * 60 * 1000
 
+// Accept a lower reported network height only when it regresses by more than
+// this many blocks. Small dips (a load-balanced daemon a block behind, lwsf
+// reporting the stored scan height until its first refresh) are smoothed so
+// confirmation counts do not bounce, while a large regression is taken as a
+// correction so one garbage reading cannot ratchet the stored height forever:
+const HEIGHT_REGRESSION_BOUND = 1000
+
+// How often to refresh the pool-watch stamp of a tx that is still in the pool.
+// Millisecond freshness buys nothing against a 30-minute grace, and a fresh
+// stamp every pass would rewrite walletLocalData on every save loop:
+const PENDING_SEEN_REFRESH_MS = 5 * 60 * 1000
+
+// How often to refresh a still-pending tx's otherParams.lastSeenTime (seconds).
+// The base engine independently drops an unconfirmed tx once that stamp is
+// older than 24 hours, and a pool residence can outlast that (monerod keeps
+// transactions for around three days), so it must be refreshed periodically;
+// hourly keeps the refresh-triggered change events negligible:
+const LAST_SEEN_REFRESH_S = 60 * 60
+
 /**
  * Converts an Edge walletId (standard base64) into the form the native monero
  * layer expects. The native code embeds the id in a filesystem path and rejects
@@ -89,8 +109,8 @@ export class MoneroEngine extends CurrencyEngine<
   private sendKeysToNative?: (keys: MoneroPrivateKeys) => void
   private syncStartHeight: number | undefined
   private txSortOrder: 'asc' | 'desc' = 'asc'
-  private queryTransactionsRun: Promise<void> | undefined
-  private queryTransactionsQueued = false
+  private readonly queryTxMutex = makeMutex()
+  private pendingSeenReset = false
   private unsubscribeWalletEvent?: () => void
   private unsubscribeNymFetch?: () => Promise<void>
   private abortKeysWait?: () => void
@@ -401,7 +421,20 @@ export class MoneroEngine extends CurrencyEngine<
         return SYNC_POLL_MS
       }
 
-      this.updateBlockHeight(status.networkHeight)
+      // Smooth small height regressions: lwsf reports the stored account scan
+      // height until its first refresh completes, a load-balanced daemon can
+      // answer a block behind the previous poll, and the base engine re-stamps
+      // every tx's confirmation count on ANY height change, so honoring those
+      // dips makes displayed confirmations bounce. A large regression is
+      // accepted as a correction, so one garbage or wrong-chain reading cannot
+      // ratchet the stored height forever.
+      const storedHeight = this.walletLocalData.blockHeight
+      const networkHeight =
+        status.networkHeight >= storedHeight ||
+        storedHeight - status.networkHeight > HEIGHT_REGRESSION_BOUND
+          ? status.networkHeight
+          : storedHeight
+      this.updateBlockHeight(networkHeight)
 
       // Refresh the balance on every poll, not only once fully synced, so a
       // pending incoming tx or the pending change after a send is reflected
@@ -470,55 +503,50 @@ export class MoneroEngine extends CurrencyEngine<
   }
 
   private async queryTransactions(nativeWalletId: string): Promise<void> {
-    const PAGE_SIZE = 50
+    // Serialize passes with the fleet-standard mutex (Tron, Polkadot, and
+    // Algorand guard queryTransactions the same way): the sync poll and the
+    // pendingTransactionReceived event can both call this, and interleaved
+    // passes would race the scan cursor and the pending diff. A caller queued
+    // behind an in-flight pass runs its own full pass afterwards, so every
+    // caller's promise covers a pass that started at or after its call, even
+    // when an earlier pass throws.
+    return await this.queryTxMutex(async () => {
+      const PAGE_SIZE = 50
 
-    // The sync poll and the pendingTransactionReceived event can both call
-    // this; interleaved passes would race the scan cursor and the pending
-    // diff. A call that arrives mid-pass flags one coalesced re-run and awaits
-    // the in-flight run, whose promise only settles once the re-run queue has
-    // drained — so every caller's promise covers a pass that started at or
-    // after its call, and chained work (the event handler's balance refresh,
-    // syncNetwork's backfill pacing) never runs against a half-finished pass.
-    if (this.queryTransactionsRun != null) {
-      this.queryTransactionsQueued = true
-      return await this.queryTransactionsRun
-    }
+      // A queued caller can land here after killEngine (a settings change or
+      // resync wipes the state next); do not write into the fresh state.
+      if (!this.engineOn) return
 
-    const run = (async () => {
       try {
-        do {
-          this.queryTransactionsQueued = false
+        // The pending diff, cursor healing, and lastSeenTime preservation all
+        // consult the in-memory transaction list, which loadEngine only loads
+        // eagerly on a first-ever sync. Load it on warm launches too
+        // (idempotent after the first call):
+        await this.loadTransactions()
 
-          try {
-            if (this.txSortOrder === 'asc') {
-              await this.queryTransactionsAsc(nativeWalletId, PAGE_SIZE)
-            } else {
-              await this.queryTransactionsDesc(nativeWalletId, PAGE_SIZE)
-            }
+        if (this.txSortOrder === 'asc') {
+          await this.queryTransactionsAsc(nativeWalletId, PAGE_SIZE)
+        } else {
+          await this.queryTransactionsDesc(nativeWalletId, PAGE_SIZE)
+        }
 
-            // Pending transactions live outside the cursor protocol above:
-            // they sort behind all confirmed history, so neither scan reaches
-            // them. Process them on every pass so they appear before their
-            // first confirmation.
-            await this.queryPendingTransactions(nativeWalletId, PAGE_SIZE)
+        // Bail between stages when the engine stopped mid-pass, to shorten
+        // the tail of writes killEngine's drain has to wait out:
+        if (!this.engineOn) return
 
-            this.sendTransactionEvents()
-          } catch (error: unknown) {
-            // Log and keep looping: if another call arrived mid-pass, the
-            // loop still owes it a full pass — exiting here would silently
-            // skip that caller's trigger until the next sync poll. The loop
-            // stays bounded: a re-run only happens when a new call arrived
-            // during the failing pass.
-            this.log.error(`queryTransactions error: ${String(error)}`)
-          }
-        } while (this.queryTransactionsQueued)
+        // Pending transactions live outside the cursor protocol above: they
+        // sort behind all confirmed history, so neither scan reaches them.
+        // Process them on every pass so they appear before their first
+        // confirmation.
+        await this.queryPendingTransactions(nativeWalletId, PAGE_SIZE)
+      } catch (error: unknown) {
+        this.log.error(`queryTransactions error: ${String(error)}`)
       } finally {
-        this.queryTransactionsRun = undefined
+        // Flush events even when a later stage failed, so confirmed txs the
+        // scans already processed are not buffered until the next block:
+        this.sendTransactionEvents()
       }
-    })()
-
-    this.queryTransactionsRun = run
-    return await run
+    })
   }
 
   /** Look up the engine's stored copy of a transaction, if any. */
@@ -542,6 +570,22 @@ export class MoneroEngine extends CurrencyEngine<
     pageSize: number
   ): Promise<void> {
     const now = Date.now()
+
+    // Time while the engine was not running must not count toward the
+    // eviction grace: a tx can confirm or leave the pool while the app is
+    // closed, and the relaunch backfill has not caught up yet. Restart each
+    // surviving watch entry's clock once per engine session.
+    if (!this.pendingSeenReset) {
+      this.pendingSeenReset = true
+      const txids = Object.keys(this.otherData.pendingTxSeen)
+      if (txids.length > 0) {
+        const reset: { [txid: string]: number } = {}
+        for (const txid of txids) reset[txid] = now
+        this.otherData.pendingTxSeen = reset
+        this.walletLocalDataDirty = true
+      }
+    }
+
     const seen = new Set<string>()
     let page = 0
     while (true) {
@@ -554,14 +598,28 @@ export class MoneroEngine extends CurrencyEngine<
         seen.add(tx.hash)
         this.processTransaction(tx)
       }
-      if ((page + 1) * pageSize >= txPage.totalCount) break
+      // The short/empty-page check is insurance against a native totalCount
+      // that disagrees with the page contents:
+      if (
+        txPage.transactions.length < pageSize ||
+        (page + 1) * pageSize >= txPage.totalCount
+      ) {
+        break
+      }
       page++
     }
 
     // Rebuild the watch map: entries seen right now, plus absent-but-graced
     // entries. Resolved (confirmed/dropped) entries are simply not carried.
     const pendingTxSeen: { [txid: string]: number } = {}
-    for (const txid of seen) pendingTxSeen[txid] = now
+    for (const txid of seen) {
+      // Refresh a still-pending entry's stamp only periodically, so
+      // steady-state passes leave otherData byte-identical instead of
+      // rewriting walletLocalData on every save loop:
+      const prev = this.otherData.pendingTxSeen[txid]
+      pendingTxSeen[txid] =
+        prev != null && now - prev < PENDING_SEEN_REFRESH_MS ? prev : now
+    }
 
     for (const [txid, lastSeenMs] of Object.entries(
       this.otherData.pendingTxSeen
@@ -571,17 +629,26 @@ export class MoneroEngine extends CurrencyEngine<
       // Stop watching entries that resolved (confirmed or already dropped):
       if (stored == null || stored.blockHeight !== 0) continue
       // 'failed' is already a terminal state (a spend the backend reported as
-      // failed); leaving the pool is expected — don't relabel it as dropped:
+      // failed); leaving the pool is expected, so don't relabel it as dropped:
       if (stored.confirmations === 'failed') continue
-      if (now - lastSeenMs < DROPPED_TX_GRACE_MS) {
-        // Still within the grace window: keep watching.
+      if (
+        now - lastSeenMs < DROPPED_TX_GRACE_MS ||
+        // Only declare drops once the backfill has completed a full confirmed
+        // scan; during 'asc' the confirmed copy may simply not be ingested
+        // yet:
+        this.txSortOrder !== 'desc'
+      ) {
+        // Still within the grace window (or unable to judge): keep watching.
         pendingTxSeen[txid] = lastSeenMs
         continue
       }
       // Set 'dropped' explicitly: the base engine clamps negative block
-      // heights to 0 unless confirmations is already a terminal state.
+      // heights to 0 unless confirmations is already a terminal state. Spread
+      // otherParams too: addTransaction stamps lastSeenTime through it, and a
+      // shared reference would silently rewrite the stored entry's stamp.
       this.addTransaction(null, {
         ...stored,
+        otherParams: { ...stored.otherParams },
         blockHeight: -1,
         confirmations: 'dropped'
       })
@@ -651,6 +718,7 @@ export class MoneroEngine extends CurrencyEngine<
   ): Promise<void> {
     let page = 0
     let foundKnownTx = false
+    let healSweep = false
     let newestTxid: string | undefined
 
     while (!foundKnownTx) {
@@ -669,15 +737,21 @@ export class MoneroEngine extends CurrencyEngine<
       }
 
       for (const tx of txPage.transactions) {
-        if (tx.hash === this.otherData.mostRecentTxid) {
+        if (!healSweep && tx.hash === this.otherData.mostRecentTxid) {
           // Heal a stale stored copy before stopping: if the anchor tx is
           // confirmed on-chain but our copy never saw the confirmation (a
           // cursor recorded while the tx was still pending), process it once
-          // more so it stops displaying as unconfirmed.
+          // more so it stops displaying as unconfirmed. The same old-code
+          // migration can have left more stuck-pending txs behind the anchor,
+          // so when the anchor itself needed healing, walk the rest of the
+          // history once, re-processing only rows whose stored copy is still
+          // missing its height.
           if (!tx.isPending) {
             const stored = this.storedTransaction(tx.hash)
             if (stored == null || stored.blockHeight < 1) {
               this.processTransaction(tx)
+              healSweep = true
+              continue
             }
           }
           foundKnownTx = true
@@ -685,6 +759,10 @@ export class MoneroEngine extends CurrencyEngine<
         }
         // Pending rows are handled by queryPendingTransactions:
         if (tx.isPending) continue
+        if (healSweep) {
+          const stored = this.storedTransaction(tx.hash)
+          if (stored != null && stored.blockHeight >= 1) continue
+        }
         this.processTransaction(tx)
       }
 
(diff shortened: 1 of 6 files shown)


src/monero/MoneroEngine.ts
```diff
@@ -461,31 +479,105 @@ export class MoneroEngine extends CurrencyEngine<
   private async queryTransactions(nativeWalletId: string): Promise<void> {
     const PAGE_SIZE = 50
 
+    // The sync poll and the pendingTransactionReceived event can both call
+    // this; interleaved passes would race the scan cursor and the pending
+    // diff, so let the in-flight pass stand for both.
+    if (this.queryTransactionsBusy) return
+    this.queryTransactionsBusy = true
```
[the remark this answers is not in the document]
peachbits: Fixed. A call arriving while a pass is in flight now sets a re-run flag instead of returning early, so the in-flight pass repeats once when it finishes — a pending tx that arrived after the pool was scanned gets a full pass rather than waiting for the next sync poll. Added a regression test (`re-runs instead of dropping a call that arrives mid-pass`) that fires a second query mid-pass and asserts the mid-pass tx is picked up.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,117 @@ export class MoneroEngine extends CurrencyEngine<
...
+    // The sync poll and the pendingTransactionReceived event can both call
+    // this; interleaved passes would race the scan cursor and the pending
+    // diff. Rather than run concurrently, a call that arrives mid-pass sets a
+    // re-run flag and the in-flight pass repeats once when it finishes.
+    // Dropping the call outright would lose the triggering event: a tx that
+    // arrives after the current pass already scanned the pool would stay
+    // invisible until the next sync poll.
+    if (this.queryTransactionsBusy) {
+      this.queryTransactionsQueued = true
+      return
```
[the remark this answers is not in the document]
peachbits: Fixed. A coalesced caller now receives the in-flight run's promise instead of an immediately-resolved one; the run promise only settles once the re-run queue has drained, so it always covers a pass that started at or after the call — the event handler's chained balance refresh and syncNetwork's pacing no longer race a half-finished pass. The regression test now also asserts the coalesced caller's promise does not resolve before the re-run has stored the mid-pass tx.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,117 @@ export class MoneroEngine extends CurrencyEngine<
...
+        pendingTxSeen[txid] = lastSeenMs
+        continue
+      }
+      // Set 'dropped' explicitly: the base engine clamps negative block
+      // heights to 0 unless confirmations is already a terminal state.
+      this.addTransaction(null, {
+        ...stored,
+        blockHeight: -1,
+        confirmations: 'dropped'
+      })
```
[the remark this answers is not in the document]
peachbits: Fixed. The drop check now skips stored transactions whose confirmations are already 'failed' — they leave the watch set without being relabeled as dropped. Added a regression test ('does not relabel a failed tx as dropped when it leaves the pool').

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,124 @@ export class MoneroEngine extends CurrencyEngine<
...
+          // Process them on every pass so they appear before their first
+          // confirmation.
+          await this.queryPendingTransactions(nativeWalletId, PAGE_SIZE)
+
+          this.sendTransactionEvents()
+        } while (this.queryTransactionsQueued)
+      } catch (error: unknown) {
+        this.log.error(`queryTransactions error: ${String(error)}`)
+      } finally {
+        this.queryTransactionsRun = undefined
```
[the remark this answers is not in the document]
peachbits: Fixed. The error handling moved inside the do/while: a failing pass logs and the loop continues, so a call queued mid-pass still gets its full pass instead of resolving without one. The loop stays bounded — a re-run only happens when a new call arrived during the failing pass. Added a regression test ('still honors a queued call when the in-flight pass throws') where the in-flight pass throws after a mid-pass call queued, asserting the re-run still ingests the new tx.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,131 @@ export class MoneroEngine extends CurrencyEngine<
...
+  // hours to days, so the grace does not meaningfully delay them.
+  private async queryPendingTransactions(
+    nativeWalletId: string,
+    pageSize: number
+  ): Promise<void> {
+    const now = Date.now()
+    const seen = new Set<string>()
+    let page = 0
+    while (true) {
+      const txPage = await this.tools.cppBridge.getPendingTransactions(
```
j0ntz, line 592: **[confirmed, blocks clean builds]** The pinned `react-native-monero@0.1.0` does not ship `getPendingTransactions`: the installed package's `lib/src/CppBridge.d.ts` has `getAllTransactions` but no `getPendingTransactions`, the lockfile resolves the registry 0.1.0 tarball, and there is no entry in `patches/`. A clean `npm ci` fails `tsc` on this line (TS2339), and a host app shipping a native build without the method throws a TypeError here on every pass, which the catch in `queryTransactions` reduces to a per-poll log line: pending txs never appear, and confirmed txs scanned earlier in the pass wait for the next block-height change to flush. Needs [EdgeApp/react-native-monero#3](https://github.com/EdgeApp/react-native-monero/pull/3) released plus a version bump here (package.json pins 0.1.0 exact, peer range ^0.1.0) so the ordering is enforced rather than remembered.

src/monero/MoneroEngine.ts
```diff
@@ -645,7 +781,16 @@ export class MoneroEngine extends CurrencyEngine<
       edgeTransaction.confirmations = 'failed'
     }
 
-    this.addTransaction(null, edgeTransaction)
+    if (tx.isPending) {
```
j0ntz, line 854: **[confirmed]** Receives now enter the store pending-first, which suppresses the new-payment notification. First add: `blockHeight` 0 makes `getTxCheckpoint` return `'0'`, so `isTransactionNew` returns false against any synced checkpoint ([CurrencyEngine.ts:293](https://github.com/EdgeApp/edge-currency-accountbased/blob/823e5893174a75bdaaee0f490c742eb8d32e2097/src/common/CurrencyEngine.ts#L293), [866](https://github.com/EdgeApp/edge-currency-accountbased/blob/823e5893174a75bdaaee0f490c742eb8d32e2097/src/common/CurrencyEngine.ts#L866)). The later confirmation takes the existing-tx path, and `updateTransaction` hardcodes `isNew: false` ([CurrencyEngine.ts:747](https://github.com/EdgeApp/edge-currency-accountbased/blob/823e5893174a75bdaaee0f490c742eb8d32e2097/src/common/CurrencyEngine.ts#L747)). Net: a receive first observed while the engine runs (the normal open-wallet case, since the pool sees it before it mines) never emits `isNew: true`, where master emitted it on the confirmed first sighting. [ZcashEngine.ts:174](https://github.com/EdgeApp/edge-currency-accountbased/blob/823e5893174a75bdaaee0f490c742eb8d32e2097/src/zcash/ZcashEngine.ts#L174) overrides `isTransactionNew` for exactly this pending-first situation; MoneroEngine needs the same override (blockHeight 0, unconfirmed, checkpoint non-null, not a send).

src/monero/MoneroEngine.ts
```diff
@@ -645,7 +781,16 @@ export class MoneroEngine extends CurrencyEngine<
...
+      // Pending txs are re-processed on every pass. Without a lastSeenTime the
+      // base engine stamps a fresh one on each add, which reads as a change
+      // and emits a spurious update event per poll — keep the first-seen time.
+      const stored = this.storedTransaction(tx.hash)
+      const lastSeenTime = asMaybe(asNumber)(stored?.otherParams?.lastSeenTime)
+      this.addTransaction(null, edgeTransaction, lastSeenTime)
```
j0ntz, line 869: **[confirmed]** Freezing `lastSeenTime` at first-seen arms the base engine's independent 24h drop clock. `otherParams.lastSeenTime` is the base's drop timer: `updateConfirmations` flips an unconfirmed height-0 tx to `'dropped'` and zeroes `nativeAmount` once `Date.now()/1000 - lastSeenTime > DROPPED_TX_TIME_GAP` (86400 s), and it runs from `updateBlockHeight`'s every-block sweep as well as from each re-add ([CurrencyEngine.ts:56](https://github.com/EdgeApp/edge-currency-accountbased/blob/823e5893174a75bdaaee0f490c742eb8d32e2097/src/common/CurrencyEngine.ts#L56), [700](https://github.com/EdgeApp/edge-currency-accountbased/blob/823e5893174a75bdaaee0f490c742eb8d32e2097/src/common/CurrencyEngine.ts#L700), [716](https://github.com/EdgeApp/edge-currency-accountbased/blob/823e5893174a75bdaaee0f490c742eb8d32e2097/src/common/CurrencyEngine.ts#L716)). Passing the preserved first-seen value defeats the base design where each re-add restamps last-seen, so a tx still returned by `getPendingTransactions` every pass shows dropped with amount 0 once its pool residence passes 24h (the comment above says real evictions take hours to days; monerod's default pool lifetime is ~3 days), while `status.balance` still includes the funds. The 30-minute `pendingTxSeen` design is silently overridden by this 24h clock, and the new tests only exercise the 30-min path. Suggested direction: restamp `lastSeenTime` on every pool sighting and instead stop `addTransaction`'s diff clause (CurrencyEngine.ts:592) from treating a lastSeenTime-only refresh as a change, so one layer owns drop detection. Two smaller notes on the same lines: `asMaybe(asNumber)` re-rolls the existing `asMaybeOtherParamsLastSeenTime` cleaner (src/common/types.ts:201), and the `blockHeight: -1` + pre-set `'dropped'` write in the drop loop works only because three base internals happen to line up (terminal-state early return before the negative-height clamp, the 0-to-negative change branch, confirmations excluded from diffing); a first-class drop path in the base would be sturdier.

src/monero/MoneroEngine.ts
```diff
@@ -645,7 +781,16 @@ export class MoneroEngine extends CurrencyEngine<
...
+    } else {
+      this.addTransaction(null, edgeTransaction)
+    }
   }
 
   async killEngine(): Promise<void> {
```
j0ntz, line 893: **[confirmed]** `killEngine` resets `txSortOrder`/`syncTracker` but neither awaits nor clears `queryTransactionsRun`, and `nativeWalletId.stop()` does not await wallet closure (the lifecycle manager's `stop()` is fire-and-forget). `resyncBlockchain` and `changeWalletSettings` then run `clearBlockchainCache` + `startEngine` on the same instance, so a pass that was mid-bridge-call resolves afterwards and writes into the freshly wiped state with no `engineOn` gate: `mostRecentTxid`/`processedTransactionCount` (lines 640-645), `pendingTxSeen` (598-601), and re-added transactions. A stale `mostRecentTxid` anchor then makes the post-resync desc scan stop early and permanently skip the wiped older history. Fix: clear `queryTransactionsRun`/`queryTransactionsQueued` in `killEngine` and have the run loop bail when `engineOn` is false or its wallet session is stale.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,131 @@ export class MoneroEngine extends CurrencyEngine<
...
+      if (stored.confirmations === 'failed') continue
+      if (now - lastSeenMs < DROPPED_TX_GRACE_MS) {
+        // Still within the grace window: keep watching.
+        pendingTxSeen[txid] = lastSeenMs
+        continue
+      }
+      // Set 'dropped' explicitly: the base engine clamps negative block
+      // heights to 0 unless confirmations is already a terminal state.
+      this.addTransaction(null, {
+        ...stored,
```
j0ntz, line 650: **[confirmed, minor]** `{ ...stored, blockHeight: -1, confirmations: 'dropped' }` shares `otherParams` with the live stored tx. `addTransaction` writes `otherParams.lastSeenTime = now` through that alias (blockHeight -1 < 1, CurrencyEngine.ts:545-548), silently rewriting the stored entry's first-seen stamp at drop time, and its lastSeenTime diff clause then compares the object with itself, so the update only lands because of the separate 0-to-negative branch. Spread `otherParams` as well, or pass the preserved stamp explicitly.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,131 @@ export class MoneroEngine extends CurrencyEngine<
...
+        this.processTransaction(tx)
+      }
+      if ((page + 1) * pageSize >= txPage.totalCount) break
+      page++
+    }
+
+    // Rebuild the watch map: entries seen right now, plus absent-but-graced
+    // entries. Resolved (confirmed/dropped) entries are simply not carried.
+    const pendingTxSeen: { [txid: string]: number } = {}
+    for (const txid of seen) pendingTxSeen[txid] = now
```
j0ntz: **[confirmed, perf]** Stamping every seen txid with a fresh `Date.now()` guarantees the `matchJson` guard below fails on every pass while anything is pending, so `walletLocalDataDirty` is set and the save loop rewrites `walletLocalData` every 10 s for the whole pool residence. Against a 30-minute grace, millisecond freshness buys nothing: carry the previous stamp unless it is stale by more than ~60 s, or only rewrite on membership changes, so steady-state passes leave `otherData` byte-identical.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,131 @@ export class MoneroEngine extends CurrencyEngine<
...
+              await this.queryTransactionsDesc(nativeWalletId, PAGE_SIZE)
+            }
+
+            // Pending transactions live outside the cursor protocol above:
+            // they sort behind all confirmed history, so neither scan reaches
+            // them. Process them on every pass so they appear before their
+            // first confirmation.
+            await this.queryPendingTransactions(nativeWalletId, PAGE_SIZE)
+
+            this.sendTransactionEvents()
```
j0ntz: **[confirmed, minor]** A failure in the new pending stage suppresses the flush of confirmed txs the asc/desc scan already processed in the same pass; they sit buffered until the next block-height change flushes them via `updateBlockHeight`. Flushing in a `finally` keeps a transient pending-query error from delaying confirmed-tx visibility by up to a block.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,131 @@ export class MoneroEngine extends CurrencyEngine<
...
+      const txPage = await this.tools.cppBridge.getPendingTransactions(
+        nativeWalletId,
+        page,
+        pageSize
+      )
+      for (const tx of txPage.transactions) {
+        seen.add(tx.hash)
+        this.processTransaction(tx)
+      }
+      if ((page + 1) * pageSize >= txPage.totalCount) break
```
j0ntz: **[plausible]** This loop's only exit is the `totalCount` math; unlike `queryTransactionsDesc` there is no short-page guard (`transactions.length < pageSize`). The `totalCount` contract for `getPendingTransactions` lives in the unmerged dep PR and cannot be checked from this repo; if it ever reports the wallet's total tx count (same `TransactionsPage` shape as `getAllTransactions`) or a sentinel, this spins empty bridge calls on every pass, or effectively forever inside the coalesced run, wedging sync. Cheap insurance: also break on an empty or short page.

src/monero/MoneroEngine.ts
```diff
@@ -548,15 +669,30 @@ export class MoneroEngine extends CurrencyEngine<
...
 
       for (const tx of txPage.transactions) {
         if (tx.hash === this.otherData.mostRecentTxid) {
+          // Heal a stale stored copy before stopping: if the anchor tx is
+          // confirmed on-chain but our copy never saw the confirmation (a
+          // cursor recorded while the tx was still pending), process it once
+          // more so it stops displaying as unconfirmed.
+          if (!tx.isPending) {
+            const stored = this.storedTransaction(tx.hash)
+            if (stored == null || stored.blockHeight < 1) {
```
j0ntz, line 751: **[plausible, unverified: review stopped before the verifier finished]** The heal re-processes only the anchor row. The old code set `mostRecentTxid` on every processed row including pending ones, so a wallet can arrive here with the cursor anchored on the newer of two stuck-pending txs; once both confirm, the desc scan heals the anchor and breaks, and the older tx's confirmed row behind it is never re-processed. Its stored copy stays at height 0, and then the 30-min watch (if it was seen in the pool) or the base 24h clock mislabels a genuinely confirmed tx as dropped, until a manual resync. Consider letting the scan continue past the anchor once whenever the anchor itself needed healing, since the same poisoned-cursor migration may have left more stuck-pending txs behind it.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,131 @@ export class MoneroEngine extends CurrencyEngine<
...
+  }
 
-      this.sendTransactionEvents()
-    } catch (error: unknown) {
-      this.log.error(`queryTransactions error: ${String(error)}`)
+  /** Look up the engine's stored copy of a transaction, if any. */
+  private storedTransaction(txid: string): EdgeTransaction | undefined {
+    const idx = this.findTransaction(null, normalizeAddress(txid))
+    if (idx < 0) return undefined
+    return this.transactionList['']?.[idx]
```
j0ntz, line 556: **[plausible, unverified: review stopped]** `storedTransaction` reads the in-memory `transactionList`/`txIdMap`, but on a warm launch of a synced wallet `loadEngine` skips `loadTransactions` (CurrencyEngine.ts:507 loads only when `seenTxCheckpoint` is null) and nothing in the sync path loads it. If the first pass runs before the core calls `getTransactions`: (a) the drop loop sees `stored == null` for every persisted watch entry and silently discards them (line 580), so evictions that happened while the app was closed are never marked; (b) the desc heal sees `stored == null` for a healthy anchor and re-processes it into the empty list, after which the next `saveWalletLoop` merge re-feeds the entire disk history through `addTransaction`; (c) pending `lastSeenTime` preservation misses. Awaiting `this.loadTransactions()` before consulting the store would close all three.

src/monero/MoneroEngine.ts
```diff
@@ -390,7 +401,15 @@ export class MoneroEngine extends CurrencyEngine<
...
 
-      this.updateBlockHeight(status.networkHeight)
+      // Clamp the reported height monotonic. lwsf seeds daemonBlockChainHeight
+      // with the stored account scan height until its first refresh completes,
+      // and a load-balanced daemon can report a height a block behind the
+      // previous poll. The base engine re-stamps every tx's confirmation count
+      // on ANY height change, so a regressing height makes the displayed
+      // confirmations bounce downward and back up.
+      this.updateBlockHeight(
+        Math.max(status.networkHeight, this.walletLocalData.blockHeight)
```
j0ntz: **[plausible, unverified: review stopped]** The clamp makes one inflated reading permanent: `walletLocalData.blockHeight` is persisted, survives restarts and `changeUserSettings` (no cache clear there), and `Math.max` prevents any later correct height from lowering it. `getWalletStatus` output is not cleaned, and this bridge demonstrably emits 2^64 sentinels for other height fields (the test's `PENDING_HEIGHT_SENTINEL`). One garbage or wrong-chain `networkHeight` (custom server, lwsf edge case) then inflates `numConfirmations` for every tx until a full resync, where pre-PR a bad poll self-corrected on the next one. Consider bounding the ratchet (ignore only regressions smaller than a plausible reorg/lag depth) or sanity-checking `networkHeight` before feeding the max.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,131 @@ export class MoneroEngine extends CurrencyEngine<
...
+      this.otherData.pendingTxSeen
+    )) {
+      if (seen.has(txid)) continue
+      const stored = this.storedTransaction(txid)
+      // Stop watching entries that resolved (confirmed or already dropped):
+      if (stored == null || stored.blockHeight !== 0) continue
+      // 'failed' is already a terminal state (a spend the backend reported as
+      // failed); leaving the pool is expected — don't relabel it as dropped:
+      if (stored.confirmations === 'failed') continue
+      if (now - lastSeenMs < DROPPED_TX_GRACE_MS) {
```
j0ntz: **[plausible, unverified: review stopped]** The grace compares wall clock against persisted `lastSeenMs`, so time while the app is closed counts as time missing from the pool, and the comment's invariant that the confirmed scan has already updated the stored copy does not hold right after a relaunch: `txSortOrder` restarts at `'asc'`, the resume pass ingests one page, and the pending pass runs in that same pass. A tx that confirmed, or still sits in the pool ahead of the first lws refresh, while the app was closed for 30+ minutes is absent with a stored height of 0 and gets a visible dropped write plus event, then flips back when the backfill or pool refresh catches up. Consider resetting the stamps to `now` on the first pass after engine start, or skipping the drop check until a pass has run with `txSortOrder === 'desc'`.

src/monero/MoneroEngine.ts
```diff
@@ -461,31 +480,131 @@ export class MoneroEngine extends CurrencyEngine<
...
-        this.sendTransactionEvents()
-        return
+    // The sync poll and the pendingTransactionReceived event can both call
+    // this; interleaved passes would race the scan cursor and the pending
+    // diff. A call that arrives mid-pass flags one coalesced re-run and awaits
+    // the in-flight run, whose promise only settles once the re-run queue has
+    // drained — so every caller's promise covers a pass that started at or
+    // after its call, and chained work (the event handler's balance refresh,
+    // syncNetwork's backfill pacing) never runs against a half-finished pass.
+    if (this.queryTransactionsRun != null) {
```
j0ntz: **[cleanup]** This hand-rolled single-flight (run promise + queued flag + do/while in an IIFE) is the fleet's third serialization idiom: Tron, Polkadot, and Algorand all guard `queryTransactions` with `makeMutex` from src/common/utils.ts:172 (`queryTxMutex`). A mutex preserves the invariants the new tests pin (no interleaving, every caller covered by a full pass, a queued caller surviving a throwing pass) at the cost of at most one redundant pass with only two callers. If coalescing is genuinely wanted, promote it to a `makeSingleFlight` next to `makeMutex` so the primitive is owned and tested once; Zano shares this bridge architecture and will want it next.

j0ntz (commented):
Review notes from a deep pass over this PR (10 finder angles, then adversarial verification per finding). Verification was stopped partway through, so labels matter: [confirmed] passed verification or direct inspection of the base engine; [plausible, unverified] has a real-looking mechanism but its verifier never returned a verdict.

Read first: the react-native-monero pin (a clean install cannot compile this branch), the isNew notification regression, the base engine 24h auto-drop armed by the frozen lastSeenTime, and the stale queryTransactions run surviving resyncBlockchain.

14 inline comments: 7 confirmed, 1 plausible, 4 unverified, 1 cleanup, 1 convention.

j0ntz (changes requested):

src/monero/MoneroEngine.ts
peachbits: Fixed: mirrored ZcashEngine's isTransactionNew override (blockHeight 0, confirmations 'unconfirmed', checkpoint set, not a send), so a first-sighted pending receive emits isNew true. Regression test asserts the notification event.

src/monero/MoneroEngine.ts
peachbits: Correct, and the local tree only passes because the rebuilt react-native-monero is synced into node_modules. Ordering is enforced at landing: EdgeApp/react-native-monero#3 merges and publishes first, then this PR bumps the exact pin and peer range to that release before merging. Leaving this thread open to track the version bump.

src/monero/MoneroEngine.ts
peachbits: Fixed: the preserved stamp now refreshes hourly (LAST_SEEN_REFRESH_S) instead of freezing at first-seen, which keeps the base engine's 24h drop clock at bay across a multi-day pool residence without emitting a change event every poll.

src/monero/MoneroEngine.ts
peachbits: Fixed: killEngine now drains the in-flight pass after super.killEngine clears engineOn, and queryTransactions exits when engineOn is false, so a stale pass cannot write into freshly wiped state. Covered by the 'ignores passes after killEngine' test.

src/monero/MoneroEngine.ts
peachbits: Fixed: the dropped re-save spreads otherParams too, so the base engine's lastSeenTime write no longer aliases the stored entry.

src/monero/MoneroEngine.ts
peachbits: Fixed: watch stamps refresh only when older than 5 minutes (PENDING_SEEN_REFRESH_MS), so steady-state passes leave otherData byte-identical and the save loop stops rewriting walletLocalData.

src/monero/MoneroEngine.ts
peachbits: Fixed: sendTransactionEvents moved into a finally, so a pending-stage failure no longer delays the flush of confirmed txs the scans already processed.

src/monero/MoneroEngine.ts
peachbits: Fixed: the pending pager also breaks on a short or empty page, so a disagreeing native totalCount cannot spin the loop.

src/monero/MoneroEngine.ts
peachbits: Fixed: when the anchor itself needed healing, the scan now walks the remaining history once, re-processing only rows whose stored copy is missing its height, so every stuck-pending tx from the old-code migration gets repaired. Test covers a second stuck tx behind the anchor.

src/monero/MoneroEngine.ts
peachbits: Fixed: queryTransactions awaits loadTransactions() up front (idempotent after the first call), so warm launches consult the loaded store for the drop diff, cursor healing, and lastSeenTime preservation.

src/monero/MoneroEngine.ts
peachbits: Fixed: the ratchet is bounded (HEIGHT_REGRESSION_BOUND, 1000 blocks). Small dips are still smoothed, and a larger regression is accepted as a correction, so one garbage or wrong-chain reading self-corrects on the next honest poll instead of sticking until resync.

src/monero/MoneroEngine.ts
peachbits: Fixed both suggested ways: watch stamps reset to now on the first pass of each engine session, and drops are only declared once txSortOrder is 'desc'. Covered by the 'does not count time while the engine was stopped toward the drop grace' test.

src/monero/MoneroEngine.ts
peachbits: Done: replaced the hand-rolled single-flight with makeMutex per the fleet idiom. A caller queued behind an in-flight pass runs its own full pass, which preserves every invariant the tests pin (no interleaving, каждый caller covered by a full pass, queued caller survives a throwing pass), and the custom run/queued machinery is gone.

src/monero/MoneroEngine.ts
```diff
@@ -459,33 +503,163 @@ export class MoneroEngine extends CurrencyEngine<
...
+        await this.queryPendingTransactions(nativeWalletId, PAGE_SIZE)
+      } catch (error: unknown) {
+        this.log.error(`queryTransactions error: ${String(error)}`)
+      } finally {
+        // Flush events even when a later stage failed, so confirmed txs the
+        // scans already processed are not buffered until the next block:
         this.sendTransactionEvents()
-        return
       }
+    })
```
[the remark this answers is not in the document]
peachbits: Fixed: killEngine now drains the mutex first and resets the session state (txSortOrder, sync tracker, watch flags) after the in-flight pass has fully finished, so its late writes cannot clobber the resets or stick across a settings-change restart. Also added an engineOn bail between the scan and pending stages to shorten the write tail the drain waits out.

j0ntz (approved):

src/monero/MoneroEngine.ts
peachbits: Done: react-native-monero 0.3.0 is published (includes getPendingTransactions) and this branch now pins 0.3.0 exact with peer range ^0.3.0, so a clean install type-checks. EdgeApp/react-native-monero#3 merged first as planned.

PR #1060, opened 2026-06-19
Fix Stellar login crash from stellar-sdk v13 default export

j0ntz:
### CHANGELOG

Does this branch warrant an entry to the CHANGELOG?

- [x] Yes
- [ ] No

### Dependencies

none

### Description

[Asana task](https://app.asana.com/0/0/1215849737061914/f)

Fixes a crash on login for any account holding a Stellar (XLM) wallet: `undefined is not an object (evaluating 'new(C()).Horizon')`.

Root cause: `4.83.0` migrated the Stellar plugin to stellar-sdk v13. The plugin runs inside edge-core-js's plugin WebView, which loads stellar-sdk's browser build (`dist/stellar-sdk.min.js`). That build exposes no usable `default` export, so the plugin's default-namespace accesses read properties off `undefined`:

- `new stellarApi.Horizon.Server(server)` threw first, in the `StellarTools` constructor that runs on login when an XLM wallet's tools/engine load.
- Every other `stellarApi.X` access (`Keypair`, `Account`, `TransactionBuilder`, `Networks`, `Operation`, `Asset`, `Memo`, `TimeoutInfinite`) was equally broken and would have failed during key derivation, address validation, and sending.

Fix: import the v13 symbols as named exports (`Horizon`, `Keypair`, `Account`, `TransactionBuilder`, `Networks`, `Operation`, `Asset`, `Memo`, `TimeoutInfinite`) instead of off the default export. Named exports resolve in both the Node (`lib`) and browser (`dist`) builds.

Verification:
- Confirmed against the actual browser build: the old `default.Horizon` access throws the exact production error; the named-import path constructs `Horizon.Server`, a `Keypair`, and a full `TransactionBuilder` transaction successfully.
- iOS simulator: built the gui with the rebuilt plugin bundle loaded via `DEBUG_ACCOUNTBASED`, logged into an account holding four XLM wallets. All loaded on login with live balances and rates (no Horizon crash). Screenshots attached.

---

> [!NOTE]
> **Low Risk**
> Targeted import refactor with no new logic; restores expected stellar-sdk v13 usage in the browser build and affects only the Stellar plugin.
> 
> **Overview**
> Fixes **login crashes for accounts with Stellar (XLM) wallets** introduced after the stellar-sdk v13 migration: the plugin WebView loads the browser bundle, where the default export is undefined, so `stellarApi.Horizon.Server` and similar calls threw immediately when tools/engine initialized.
> 
[...]

PR #1081, opened 2026-07-31
Guard Stellar Horizon.Server against Horizon-less stellar-sdk builds

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

[Asana task](https://app.asana.com/0/1215088146871429/1217047261970600)

Hardens the Stellar plugin against the recurring XLM engine-start crash `undefined is not an object (evaluating 'new Horizon.Server')`, which pops a background alert whenever the bundled `stellar-sdk` does not expose the `Horizon` namespace.

**History.** This call site has broken twice. First (v4.83.0) from the stellar-sdk v13 migration reading a non-existent default export, fixed in #1060 by switching to named imports (`import { Horizon } from 'stellar-sdk'`). Then a recurrence where the runtime again reports `Horizon` undefined at `new Horizon.Server(...)`.

**Root cause of the recurrence.** Not a source or bundling defect. Investigation of the current published bundle shows the fix is intact and working:

- `stellar-sdk@13.3.0` `exports["."]` resolves `browser` to `dist/stellar-sdk.min.js` and `default` to `lib/index.js`; both builds expose a `Horizon` object with a `Server` constructor.
- A webpack build using this repo's exact config (`target: ['web','es5']`, browser resolution) resolves stellar-sdk to the browser dist and successfully runs `new Horizon.Server(...)`. The full `webpack` build compiles the call as `new G.Horizon.Server(...)` with `G` bound to the Horizon-bearing browser dist.
- Published-tarball bisect: `v4.84.0` is the last broken bundle (`new F.Server(` default-import form, no `Horizon.Server`); every release from `v4.84.1` through `v4.86.3` carries the fixed `new X.Horizon.Server`.
- `stellar-sdk@0.11.0` (a pre-`Horizon` version) exposes no `Horizon` namespace at all. A bundle built against it compiles `Horizon.Server` where `Horizon` is `undefined`, producing exactly the observed crash. This is a build-environment drift (wrong stellar-sdk version installed at bundle time), not a defect in the shipped code.

**The fix.** Since the source is correct and only a wrong-version build reintroduces the crash, this makes that failure loud and self-diagnosing instead of a silent recurring toast:

1. `StellarTools` constructor guard: throws a named, actionable error when `Horizon.Server` is not a constructor, replacing the opaque `undefined is not an object` WebView crash.
2. `test/stellar/stellarSdkResolution.test.ts`: asserts the named `Horizon` import provides a `Server` constructor (and `Keypair` is a constructor), so a Horizon-less stellar-sdk build fails `npm test` rather than shipping.

The guard is a no-op on a correct build, so it does not change XLM behavior when stellar-sdk resolves properly.

**Testing.** Built this branch, `updot`-baked it into `edge-react-gui`, and ran on the iOS 18.6 simulator (edge-funds). Confirmed the guarded bundle is embedded in the running app (`new G.Horizon.Server` + the guard string present in `Edge.app/edge-currency-accountbased.bundle`). Created a Stellar (XLM) wallet: the engine started with no Horizon crash, and funded XLM wallets synced fully (My Stellar 172 XLM / $29.32, My Stellar 2 1.281 XLM / $0.21) with live rate, price chart, and Horizon-served transaction history. Zero `new Horizon.Server` / `undefined is not an object` errors across the 14-minute session (the only alert observed was an unrelated benign sync-repo 404 for the freshly-created empty wallet). `npm run test` and the WebView webpack build both pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fixed: (Stellar) Guard `new Horizon.Server()` at engine start so a stellar-sdk build without the `Horizon` namespace fails with a clear, actionable error instead of an opaque `undefined is not an object` WebView crash, and add a resolution regression test so a Horizon-less stellar-sdk build is caught by the test suite rather than shipping as a recurring background toast.
+
 ## 4.86.3 (2026-07-27)
 
 - fixed: (Thorchain/Mayachain) Stop recording failed MsgDeposit actions (Midgard `type: 'failed'`) as full-amount sends. The reverted deposit never moved funds, so only the burned network fee is recorded, marked as a failed transaction — matching the existing failed-send handling.
--- a/src/stellar/StellarTools.ts
+++ b/src/stellar/StellarTools.ts
@@ -48,6 +48,19 @@ export class StellarTools implements EdgeCurrencyTools {
     // stellar-sdk v13 removed the global Network singleton; the network
     // passphrase is now passed per-transaction in StellarEngine's
     // TransactionBuilder options.
+    //
+    // Guard the Horizon namespace. `new Horizon.Server()` has crashed XLM
+    // engine start twice with an opaque `undefined is not an object` WebView
+    // error: once from the v13 default-import migration (fixed in #1060), and
+    // again when a bundle was built against a stale stellar-sdk (pre-v13
+    // versions expose no `Horizon` namespace, so the named import binds
+    // undefined). Fail loudly with an actionable message so the cause is
+    // obvious instead of surfacing as a recurring background toast.
+    if (typeof Horizon?.Server !== 'function') {
+      throw new Error(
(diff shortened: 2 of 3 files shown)


j0ntz:
## 📸 Test evidence (iOS 18.6 sim, edge-funds): XLM engine synced, no Horizon crash

**xlm wallets synced**

xlm wallets synced

**xlm balance synced**

xlm balance synced

**xlm horizon txhistory**

xlm horizon txhistory

j0ntz:
Closing unmerged: the root cause was orchestration build-environment drift, not a product defect. The agent fleet's worktrees APFS-clone node_modules from a main checkout that had fallen 64 commits behind (stellar-sdk 0.11.0, no Horizon namespace), so locally built WebView bundles baked the wrong SDK. Published bundles v4.84.1 through v4.86.3 were verified correct. Fixed on the orch side: stale-clone detection at workspace setup, a hard build gate on the marker, and scheduled main-checkout refreshes. The runtime guard and resolution test here are unnecessary in the product.

PR #1086, opened 2026-08-12
Monero: Preserve the transaction key of outgoing transactions

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Outgoing Monero transactions have shown an empty Transaction Key since Edge 4.49.0, so senders cannot prove a payment. At least one user has been stuck on a swap because of it.

The old `edge-currency-monero` implementation set `txSecret: result.tx_key` on the transaction `makeSpend` returned. The `react-native-monero` replacement does not, and `txSecret` is the only thing the core stores the key from: it writes the key into the transaction's metadata file the first time it sees the transaction, which for our own sends is the `saveTx` right after the broadcast. The sync path's existing `txSecret: tx.txKey` therefore never reaches disk for a transaction we sent ourselves, because that file already exists by then.

Two changes:

- `broadcastTx` reads the transaction key from the native wallet and puts it on the returned transaction. It has to be read there rather than in `makeSpend`: the native wallet only records the key once the transaction is committed. A key we cannot read is not worth failing an already-broadcast payment over, so a failure logs and reports no key.
- `emitStoredTxSecrets` re-reports stored transactions that carry a key, once per engine session, so the core can save a key it never stored. That covers sends made while the send path was reporting none, on any device whose Monero wallet cache still holds the key. It uses a session flag rather than a stored migration flag, so it does not depend on which package publishes first, and it is a no-op for a transaction whose key the core already has.

The core-side half, which saves a secret onto a metadata file that already exists, is [EdgeApp/edge-core-js#735](https://github.com/EdgeApp/edge-core-js/pull/735).

A key that was never persisted anywhere is gone for good: Monero picks the transaction key at random while building the transaction (`device_default::open_tx`), so it cannot be regenerated from the seed.

Asana: https://app.asana.com/0/1215088146871429/1216965258637021

---

> [!NOTE]
> **Medium Risk**
> Touches the Monero broadcast path and transaction-secret reporting used as payment proof. Broadcast still succeeds if the key cannot be read, limiting fund-movement risk.
> 
> **Overview**
> Restores **Monero transaction keys** (the sender's only proof of payment) that were dropped after the move to `react-native-monero`.
> 
> `broadcastTx` now reads the key from the native pending set after commit and attaches it as `txSecret`, so the core can persist it on the immediate `saveTx`. Failures to read the key are logged and return no secret rather than failing an already-broadcast payment.
> 
> Also adds a one-time-per-session `emitStoredTxSecrets` pass that re-reports stored txs that already carry a key, so the core can backfill secrets for sends made while the send path reported none.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab17a76c9eaaf68c6bf54ac178c32e1490de2016. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fixed: (Monero) Outgoing transactions report their transaction key again. The key is the sender's only proof of a payment, and the move to react-native-monero dropped it from the send path, so every send since then has shown an empty Transaction Key. Stored transactions also re-report their key once per session, which lets the core save a key for sends made while the send path was reporting none.
+
 ## 4.87.0 (2026-08-02)
 
 - added: (Sui) `rpcNodes`, `rpcNodesArchival`, and `maxRequestsPerSecond` to the info payload, so nodes can be changed without a client release. Transaction sweeps start on an archival node, since the walk begins at the wallet's oldest transaction and a pruned node rejects a cursor older than its retention window.
--- a/src/monero/MoneroEngine.ts
+++ b/src/monero/MoneroEngine.ts
@@ -107,6 +107,9 @@ const PENDING_SEEN_REFRESH_MS = 5 * 60 * 1000
 // hourly keeps the refresh-triggered change events negligible:
 const LAST_SEEN_REFRESH_S = 60 * 60
 
+// How many transactions to request per native transaction-list call:
+const TX_PAGE_SIZE = 50
+
 /**
  * Converts an Edge walletId (standard base64) into the form the native monero
  * layer expects. The native code embeds the id in a filesystem path and rejects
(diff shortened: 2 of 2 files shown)


j0ntz:
## 📸 Test evidence

**xmr send ready**

xmr send ready

**xmr transaction key**

xmr transaction key

j0ntz:
Closing in favor of [#1089](https://github.com/EdgeApp/edge-currency-accountbased/pull/1089), per the direction on the [Asana task](https://app.asana.com/0/1215088146871429/1216965258637021): keep the fix contained to the Monero stack, with no edge-core-js dependency.

The send-path halves are equivalent; both put the key on the transaction `broadcastTx` returns, so the core saves it at the transaction's first file write. The fork is the 4.49-window backfill. This PR re-emitted stored keys so [EdgeApp/edge-core-js#735](https://github.com/EdgeApp/edge-core-js/pull/735) could write them into the synced metadata file; #1089 mirrors them into `otherParams` on the engine's local store, the deliberate no-core-changes trade its description names. One support-facing consequence worth recording: the mirrored key is device-local and a later resync or reinstall drops it, so affected users should save their key once it reappears.

The in-app verification here (a real XMR send on the sim reaching Transaction Success, with the Transaction Key and Payment Proof rows in Advanced Details) exercised the same send-path behavior #1089 ships.

PR #1090, opened 2026-08-18
Persist the Zano wallet file once synced and run adopted wallets

j0ntz:
### Technical Design Document

[zano-migration-off-module-queue.md](https://github.com/EdgeApp/react-native-zano/blob/jon/ios-perf-zano-xmr/src/docs/zano-migration-off-module-queue.md)

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/react-native-zano/pull/17 (behavioral, not build-time: the adopt-path `run_wallet` call only matters under that PR's postponed-run mode, and is a harmless no-op on react-native-zano 0.4.0, so this PR builds and passes CI standalone)

### Description

Part of the fix for the iOS performance regression investigated in https://app.asana.com/0/1215088146871429/1217559756673909. Two changes ship here.

**1. Transaction query: mempool sweep and loop termination** (fixes the report from David: a sent transaction never appears in the receiving wallet, no incoming notification, and the sender stays pending even after the transaction has confirmations)

- `get_recent_txs_and_info2` prepends a wallet's unconfirmed transfers only when the offset is zero, so once a wallet had any confirmed history the paged catch-up could never see a transaction before it was mined. `queryTransactions` now sweeps offset zero on every poll, then pages from the persisted offset. The extra request returns at most 100 in-memory entries and is idempotent for transactions already known.
- The paging loop terminated only on the `total_transfers - last_item_index === 1` identity. `last_item_index` counts just the transfers the wallet returned, so a page whose newest entries were all filtered out (mining and defragmentation transfers are excluded by request) leaves it permanently short of that identity and the loop re-requested the same page forever, starving every later balance and transaction update for that wallet. Termination now depends on the offset actually advancing.

**2. Receive notification for swept transfers** (the other half of David's "no incoming notification")

- A transfer the mempool sweep surfaces enters the store at block height zero. `CurrencyEngine.isTransactionNew` compares a transaction's checkpoint, which is its block height, against the wallet's last seen checkpoint, so zero reads as already seen and the arrival reaches the core as a change rather than a new transaction. The confirmation that follows takes `addTransaction`'s update path, a change as well, so the GUI's receive dropdown never fired for a Zano transfer at all.
- `ZanoEngine` now overrides `isTransactionNew` the same way `MoneroEngine` and `ZcashEngine` already do: an incoming unconfirmed transaction is new once the wallet has a seen checkpoint. The same multi-device caveat those engines carry applies, since a second device tracks its own checkpoint.

**3. Wallet-file persistence and adopted wallets**

- The Zano native library only writes the wallet file when a wallet closes, and a mobile app is killed rather than closed, so sync progress was almost never persisted and every cold launch re-scanned a window that only grows (measured ~200-420% process CPU for ~6 minutes per launch on a sim account whose three Zano wallets were 26 days behind). `syncNetwork`'s synced branch now stores the wallet file through the per-wallet RPC (`invoke {"method":"store"}`): once when the wallet first reaches synced, then at most every ten minutes, and only when the wallet height advanced past the last store. It runs only while synced, when the refresh worker is idle and the per-wallet lock is free, so the call cannot block behind a scan.
- The ALREADY_EXISTS adopt path now starts the refresh worker (`run_wallet`, idempotent) for the adopted wallet. Under react-native-zano's postponed-run mode a wallet left open by an interrupted `startWallet` is not yet running, and adopting it without this call would hand back a wallet that never syncs.
- `resyncBlockchain` resets the store gates, so the rebuilt wallet's file is written at the first synced tick after the rescan instead of waiting out the throttle.

### Testing

Unit tests: `test/zano/ZanoEngineQueryTransactions.test.ts` drives `queryTransactions` against a fake wallet mirroring the two SDK behaviors the fix turns on, both read from the pinned Zano source (`wallet_rpc_server.cpp:476-505`, `wallet2.cpp:4580-4618`). Against the pre-fix code the filtered-tail case hangs (killed at 120 seconds, reproducing the infinite loop directly); with the fix, 3/3 pass.

In-app, iOS simulator (edge-funds, two Zano wallets), with this branch plus react-native-zano#17 and edge-core-js#738 linked: sent 0.063 ZANO between the two wallets and watched both sides. The receiving wallet showed the incoming transaction while it was still unconfirmed, then moved through its confirmation count; the sending wallet moved from pending through confirmations to confirmed. Both are the exact symptoms of the report. Proof frames are attached in a comment below.

The receive notification was driven separately on 2026-08-22, with `AccountCallbackManager` and `ReceiveDropdown` instrumented locally to trace every transaction event (instrumentation reverted). Before the override, two 0.063 ZANO transfers reached the receiving wallet as `transactionsChanged` only and raised no dropdown, with a 2-second screenshot loop covering the whole arrival window. After it, a third transfer produced `newTransactions` on the receiving wallet 21 seconds after broadcast, while the transfer was still unconfirmed, and the "0.0634 ZANO Received" dropdown rendered. The new fourth unit case covers the same path and fails against the pre-fix engine; the suite is 530 passing.

Wallet-file persistence was verified separately on the same sim: after each wallet reached synced its file was rewritten within the interval (1.7KB restore stubs grew to 80-158KB full wallets), and a subsequent relaunch opened all three wallets already at the chain tip with no catch-up scan.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - fixed: (Zano) Stop the transaction query from spinning forever when the newest history entries are mining or defragmentation transactions. The wallet excludes those from its answer without advancing `last_item_index`, so the loop's end-of-history test could never become true, and the wallet it happened to stopped reporting balances and transactions entirely.
 - fixed: (Zano) Show unconfirmed transactions on a wallet that already has history. The wallet reports its mempool only when asked from offset zero, so a paged catch-up starting anywhere else left an incoming transfer invisible until it was mined and a sent one pending with nothing to confirm it.
 - fixed: (Zano) Start the refresh worker when adopting an already-open wallet, so a wallet left open by an interrupted `startWallet` still syncs under react-native-zano's postponed-run mode (0.4.1).
+- fixed: (Zano) Report an unconfirmed incoming transfer as a new transaction, so the app raises its receive notification. The mempool sweep files the transfer at block height zero, which the checkpoint comparison reads as already seen, and the later confirmation takes the update path, so the arrival that the sweep exists to surface was never announced.
 
 ## 4.88.0 (2026-08-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- fixed: (Zano) Persist the wallet file once a wallet reaches synced and every ten minutes afterwards. The native library only writes the file when a wallet closes, and a mobile app is killed rather than closed, so every launch re-scanned everything since the file was last written - a catch-up window that only grew, re-paid at full multi-core CPU on each cold start.
+- fixed: (Zano) Start the refresh worker when adopting an already-open wallet, so a wallet left open by an interrupted `startWallet` still syncs under react-native-zano's postponed-run mode (0.4.1).
+
 ## 4.88.0 (2026-08-14)
 
 - added: (Zano) Verify that the native wallet address matches the address derived from the seed phrase when a wallet starts, failing the start rather than syncing a wallet whose native address is not the one shown to the user.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased
 
 - fixed: (Zano) Persist the wallet file once a wallet reaches synced and every ten minutes afterwards. The native library only writes the file when a wallet closes, and a mobile app is killed rather than closed, so every launch re-scanned everything since the file was last written - a catch-up window that only grew, re-paid at full multi-core CPU on each cold start.
+- fixed: (Zano) Stop the transaction query from spinning forever when the newest history entries are mining or defragmentation transactions. The wallet excludes those from its answer without advancing `last_item_index`, so the loop's end-of-history test could never become true, and the wallet it happened to stopped reporting balances and transactions entirely.
+- fixed: (Zano) Show unconfirmed transactions on a wallet that already has history. The wallet reports its mempool only when asked from offset zero, so a paged catch-up starting anywhere else left an incoming transfer invisible until it was mined and a sent one pending with nothing to confirm it.
 - fixed: (Zano) Start the refresh worker when adopting an already-open wallet, so a wallet left open by an interrupted `startWallet` still syncs under react-native-zano's postponed-run mode (0.4.1).
 
 ## 4.88.0 (2026-08-14)
--- a/src/zano/ZanoEngine.ts
+++ b/src/zano/ZanoEngine.ts
@@ -1,5 +1,5 @@
 import { abs, add, eq, gt, lt, mul, sub } from 'biggystring'
-import { asMaybe } from 'cleaners'
+import { asJSON, asMaybe, asObject, asValue } from 'cleaners'
 import {
   EdgeCurrencyEngine,
   EdgeCurrencyEngineOptions,
@@ -50,6 +50,20 @@ import {
   ZanoWalletOtherData
 } from './zanoTypes'
 
+/**
+ * How often a synced wallet writes its file to disk. The native library only
+ * stores on close, and a mobile app is killed rather than closed, so without
+ * periodic stores every launch re-scans from wherever the file was last
+ * written - a window that only grows.
+ */
+const WALLET_STORE_INTERVAL_MS = 10 * 60 * 1000
+
+/** The wallet RPC answers a successful store with a result object. */
+const asStoreResponse = asObject({ result: asObject({}) })
+
+/** A successful run_wallet answers with error_code OK. */
+const asRunWalletResponse = asObject({ error_code: asValue('OK') })
+
 export class ZanoEngine extends CurrencyEngine<
   ZanoTools,
   SafeZanoWalletInfo,
@@ -62,6 +76,14 @@ export class ZanoEngine extends CurrencyEngine<
   private readonly nativeId: LifecycleManager<number>
   private sendKeysToNative?: (keys: ZanoPrivateKeys) => void
   private needsNativeStorageClear: boolean = false
+  private lastStoreTime: number = 0
+  private lastStoreHeight: number = 0
+  /**
+   * Bumped whenever the store gates are reset. A store that was already in
+   * flight across a reset carries the old value and must not write its
+   * pre-reset height back into the gates.
+   */
+  private storeGeneration: number = 0
 
   constructor(
     env: PluginEnvironment<ZanoNetworkInfo>,
@@ -167,6 +189,27 @@ export class ZanoEngine extends CurrencyEngine<
           this.log(
             `initializeWallet: found existing wallet with ID ${existingWallet.wallet_id}`
           )
+
+          // From react-native-zano 0.4.1, wallets open with the refresh
+          // worker postponed, so a wallet left open by an interrupted
+          // `startWallet` may not be syncing; the postponed-run flag is
+          // process-wide and sticky, so the adopt path cannot know which way
+          // it was opened and must start the worker itself. On 0.4.0, where
+          // open auto-starts the worker, the same call is a harmless no-op.
+          // A failure here must throw: returning the id would mark the
+          // lifecycle started and hand back a wallet that never syncs,
+          // while throwing lets the next `get()` retry the whole start.
+          const runResponse = await this.tools.zano.syncCall(
+            'run_wallet',
+            existingWallet.wallet_id,
+            ''
+          )
+          if (asMaybe(asJSON(asRunWalletResponse))(runResponse) == null) {
+            throw new Error(
+              `initializeWallet: could not run the adopted wallet: ${runResponse}`
+            )
+          }
+
           return existingWallet.wallet_id
         }
       },
@@ -402,6 +445,7 @@ export class ZanoEngine extends CurrencyEngine<
       )
       await this.queryBalance()
       await this.queryTransactions()
+      await this.storeWalletFile(nativeId, status.current_wallet_height)
       return 20000
     } else {
       this.syncTracker.updateBlockRatio(
@@ -413,12 +457,61 @@ export class ZanoEngine extends CurrencyEngine<
     }
   }
 
+  /**
+   * Persists the native wallet file, so the next launch resumes from this
+   * height instead of re-scanning everything since the file was last
+   * written. Called only from the synced branch of `syncNetwork`, where the
+   * refresh worker is idle and the per-wallet lock is free; a store during
+   * the catch-up scan would block behind that lock. Throttled, and skipped
+   * entirely when the wallet height has not advanced past the last store. A
+   * failure only costs the next launch a longer catch-up, so it logs and
+   * moves on.
+   */
+  private async storeWalletFile(
+    nativeId: number,
+    walletHeight: number
+  ): Promise<void> {
+    if (walletHeight <= this.lastStoreHeight) return
+    const now = Date.now()
+    if (
+      this.lastStoreTime !== 0 &&
+      now - this.lastStoreTime < WALLET_STORE_INTERVAL_MS
+    ) {
+      return
+    }
+    const generation = this.storeGeneration
+    try {
+      const response = await this.tools.zano.invoke(
+        nativeId,
+        JSON.stringify({ method: 'store', params: {} })
+      )
+      if (asMaybe(asJSON(asStoreResponse))(response) == null) {
+        throw new Error(response)
+      }
+      // A resync between the call and its answer reset the gates for a
+      // wallet that no longer holds this height. Recording it would gate the
+      // rebuilt wallet at a tip it has not re-reached, so a kill before the
+      // chain advances would re-pay the whole rescan.
+      if (generation !== this.storeGeneration) return
+      this.lastStoreTime = now
+      this.lastStoreHeight = walletHeight
+    } catch (error: unknown) {
+      this.log.warn(`storeWalletFile failed: ${String(error)}`)
+    }
+  }
+
   // // ****************************************************************************
   // // Public methods
   // // ****************************************************************************
 
   async resyncBlockchain(): Promise<void> {
     this.needsNativeStorageClear = true
+    // The rebuilt wallet re-earns every height, so the store gates must not
+    // carry over: left in place they would skip the first store at the
+    // re-reached tip, and a kill in that window re-pays the full rescan.
+    this.lastStoreTime = 0
+    this.lastStoreHeight = 0
+    this.storeGeneration += 1
     this.nativeId.stop()
     this.unlockedBalanceMap.clear()
     await this.killEngine()
--- a/src/zano/ZanoEngine.ts
+++ b/src/zano/ZanoEngine.ts
@@ -289,30 +289,48 @@ export class ZanoEngine extends CurrencyEngine<
     const nativeId = await this.nativeId.get()
     if (nativeId == null) return
 
-    while (true) {
-      const offset = this.otherData.transactionQueryOffset
-      const transactions = await this.tools.zano.getTransactions(
-        nativeId,
-        offset
-      )
+    // Sweep the mempool first. `get_recent_txs_and_info2` prepends the
+    // wallet's unconfirmed transfers only when the offset is zero, so once a
+    // wallet has any confirmed history the paged catch-up below can never see
+    // a transaction before it is mined: an incoming transfer stays invisible
+    // to the receiver, and the sender has nothing to confirm its own.
+    const pendingPage = await this.tools.zano.getTransactions(nativeId, 0)
+    const pendingTransfers = pendingPage.transfers ?? []
+    pendingTransfers.forEach(this.processTransaction)
 
-      if (offset !== transactions.last_item_index) {
-        this.otherData.transactionQueryOffset = transactions.last_item_index
+    let offset = this.otherData.transactionQueryOffset
+    while (true) {
+      const page =
+        offset === 0
+          ? pendingPage
+          : await this.tools.zano.getTransactions(nativeId, offset)
+      const transfers = page.transfers ?? []
+      transfers.forEach(this.processTransaction)
+
+      const lastItemIndex = page.last_item_index
+      const totalTransfers = page.total_transfers
+      if (offset !== lastItemIndex) {
+        this.otherData.transactionQueryOffset = lastItemIndex
         this.walletLocalDataDirty = true
       }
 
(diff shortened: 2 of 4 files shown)


j0ntz:
## 📸 Test evidence

**after fix relaunch**

after fix relaunch

src/zano/ZanoEngine.ts
```diff
@@ -402,6 +436,7 @@ export class ZanoEngine extends CurrencyEngine<
       )
       await this.queryBalance()
       await this.queryTransactions()
+      await this.storeWalletFile(nativeId, status.current_wallet_height)
```
paullinator: **Issue:** In-flight `syncNetwork` is not cancelled by `killEngine`/`resyncBlockchain` (`makePeriodicTask.stop` does not await the running pass). `queryBalance` and `queryTransactions` each call `nativeId.get()` again, so after `stop()` they can start a new native wallet while `engineOn` is false. `queryTransactions` also writes `transactionQueryOffset` with no `engineOn` check after each `getTransactions` await, so a resync can have `clearBlockchainCache` followed by a stale pass mutating the wiped `otherData`.

**Recommendation:** Pass the already-fetched `nativeId` into `queryBalance`/`queryTransactions` and never call `get()` from those helpers. After every await in `syncNetwork` and the transaction loop, return if `!this.engineOn`. Drain in-flight query/store work in `killEngine` before clearing cache on resync.

src/zano/ZanoEngine.ts
```diff
@@ -167,6 +183,24 @@ export class ZanoEngine extends CurrencyEngine<
...
+          // may not be syncing. Starting the worker is idempotent, and a
+          // failure here must throw: returning the id would mark the
+          // lifecycle started and hand back a wallet that never syncs,
+          // while throwing lets the next `get()` retry the whole start.
+          const runResponse = await this.tools.zano.syncCall(
+            'run_wallet',
+            existingWallet.wallet_id,
+            ''
+          )
+          if (asMaybe(asRunWalletResponse)(JSON.parse(runResponse)) == null) {
```
paullinator: **Warning:** Native `run_wallet` and `store` JSON is `JSON.parse`d before `asMaybe(...)`, so invalid JSON throws `SyntaxError` instead of failing the cleaner the same way a shape mismatch does.

**Recommendation:** Use `asMaybe(asJSON(asRunWalletResponse))(runResponse)` and `asMaybe(asJSON(asStoreResponse))(response)` and drop the separate `JSON.parse`.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- fixed: (Zano) Persist the wallet file once a wallet reaches synced and every ten minutes afterwards. The native library only writes the file when a wallet closes, and a mobile app is killed rather than closed, so every launch re-scanned everything since the file was last written - a catch-up window that only grew, re-paid at full multi-core CPU on each cold start.
```
paullinator: **Warning:** Unreleased says the wallet file is persisted 'once a wallet reaches synced and every ten minutes afterwards'. `storeWalletFile` also returns immediately when `walletHeight <= lastStoreHeight`, so a wallet sitting at the same tip is not rewritten on a timer.

**Recommendation:** Qualify the Unreleased bullet the same way as the JSDoc: first store when the synced branch runs, then again only after both ten minutes and a wallet-height advance.

src/zano/ZanoEngine.ts
```diff
@@ -413,12 +448,54 @@ export class ZanoEngine extends CurrencyEngine<
...
 
+  /**
+   * Persists the native wallet file, so the next launch resumes from this
+   * height instead of re-scanning everything since the file was last
+   * written. Called only from the synced branch of `syncNetwork`, where the
+   * refresh worker is idle and the per-wallet lock is free; a store during
+   * the catch-up scan would block behind that lock. Throttled, and skipped
+   * entirely when the wallet height has not advanced past the last store. A
+   * failure only costs the next launch a longer catch-up, so it logs and
+   * moves on.
```
paullinator: **Warning:** `storeWalletFile` JSDoc says it runs 'where the refresh worker is idle and the per-wallet lock is free'. After postponed-run, `run_wallet` leaves the refresh worker running for the life of the wallet; synced means catch-up finished, not that the worker is idle. The method also does not consult `is_in_long_refresh`.

**Recommendation:** Describe the real constraint: call only from the `progress === 100 || wallet_state === 2` branch so store is not issued during the initial catch-up scan that holds the per-wallet lock.

src/zano/ZanoEngine.ts
```diff
@@ -167,6 +183,24 @@ export class ZanoEngine extends CurrencyEngine<
...
+            ''
+          )
+          if (asMaybe(asRunWalletResponse)(JSON.parse(runResponse)) == null) {
+            throw new Error(
+              `initializeWallet: could not run the adopted wallet: ${runResponse}`
+            )
+          }
+
           return existingWallet.wallet_id
         }
```
paullinator: **Warning:** ZanoEngine native start/adopt, `storeWalletFile` throttling, and `resyncBlockchain` have no unit tests. `test/zano` only covers ZanoTools, zanoMnemonic, and the checksum pin.

**Recommendation:** Add tests modeled on `test/monero/MoneroEnginePendingTxs.test.ts` covering happy `startWallet`, address mismatch (assert `closeWallet` then throw), ALREADY_EXISTS adopt with matching path, adopt address mismatch, `run_wallet` non-OK throw, store skipped while catching up / at same height / inside the interval, and store again after resync at the same tip.

src/zano/ZanoEngine.ts
```diff
@@ -167,6 +183,24 @@ export class ZanoEngine extends CurrencyEngine<
           this.log(
             `initializeWallet: found existing wallet with ID ${existingWallet.wallet_id}`
           )
+
+          // Since react-native-zano opens wallets with the refresh worker
+          // postponed, a wallet left open by an interrupted `startWallet`
+          // may not be syncing. Starting the worker is idempotent, and a
+          // failure here must throw: returning the id would mark the
+          // lifecycle started and hand back a wallet that never syncs,
+          // while throwing lets the next `get()` retry the whole start.
```
paullinator: **Suggestion:** The adopt-path comment states as fact that 'react-native-zano opens wallets with the refresh worker postponed'. This package still depends on `react-native-zano` `^0.4.0`, where open auto-starts the worker and `run_wallet` is a no-op.

**Recommendation:** Tie the comment (and the dependency) to postponed-run from 0.4.1. On 0.4.0 the same `run_wallet` call is harmless; adopt must `run_wallet` because that configure flag is process-wide and sticky.

paullinator (dismissed):
## Additional Findings

- **warning** `src/zano/ZanoEngine.ts:461`: `resyncBlockchain` now zeros `lastStoreTime`/`lastStoreHeight` (the Bugbot gap), but an in-flight `storeWalletFile` from the previous synced poll can still complete afterwards and write the old tip back into those fields. The rebuilt wallet then hits the height gate at the same tip and never stores until the chain advances, so a later kill pays the full catch-up this throttle exists to avoid.
  - Bump a store generation in `resyncBlockchain` and ignore lastStore updates unless the generation still matches after `invoke`. Alternatively drain/await the in-flight store before resetting the gates.
- **warning** `src/zano/ZanoEngine.ts:100`: `onStart` awaits a constructor-lifetime `keysPromise` with no abort. Monero on this branch uses `abortKeysWait` because `LifecycleManager.stop()` during `starting` still waits on `onStart`. If `killEngine` runs before keys are delivered, a later `nativeId.get()` (including `getDisplayPublicSeed`) waits forever on that transition, since `syncNetwork` will not send keys while `engineOn` is false.
  - Mirror Monero: race `keysPromise` against an abort reject, set the abort from `killEngine` before `nativeId.stop()`, and clear it after the wait.
- **warning** `src/zano/ZanoEngine.ts:177`: The ALREADY_EXISTS adopt path throws on an address mismatch without closing the open native wallet. `startWallet` mismatch already closes because the lifecycle manager does not run `onStop` when `onStart` throws. The adopted handle stays open, every retry hits ALREADY_EXISTS again, and the engine never starts.
  - Close `existingWallet.wallet_id` (best-effort, same as the `startWallet` mismatch path) before throwing on the address check. If `wi.address` is missing, throw a retryable error instead of treating it as a mismatch.
- **warning** `src/monero/MoneroEngine.ts:288`: The `pendingTransactionReceived` handler runs `refreshBalance` in a `.then()` after `queryTransactions` releases `queryTxMutex`. `killEngine` drains that mutex then resets session state, so `refreshBalance` can run afterwards with no `engineOn` check and publish a stale native balance onto a killed or already-restarted engine.
  - Call `refreshBalance` inside the mutex (or immediately after `queryTransactions` with an `engineOn` guard), and skip it if `!this.engineOn`.
- **warning** `src/monero/MoneroEngine.ts:843`: `queryPendingTransactions` does not re-check `engineOn` after each `getPendingTransactions` await. A stop mid-pagination leaves `seen` incomplete, then the drop pass treats missing hashes as gone from the pool. If grace has already expired and `txSortOrder === 'desc'`, that can persist `confirmations: 'dropped'` for txs that were still pending.
  - Return after every await if `!this.engineOn`, and do not rebuild `pendingTxSeen` or mark drops unless the page loop completed.
- **warning** `src/zcash/ZcashEngine.ts:461`: `rescanSettling` is cleared in `finally` before `initData()`. Progress/`update` handlers can therefore run after the rewind window is closed but before engine fields are reset. `updateBlockHeight` still runs during `rescanSettling`, so height callbacks fire for the wallet being rewound.
  - Keep `rescanSettling` true until after `initData()`, and move the `rescanSettling` guard above `updateBlockHeight`.
- **warning** `src/zcash/ZcashEngine.ts:146`: Any failure from `getIronwoodActivationHeight` is stored as `null` for the rest of the engine lifetime, so a transient native/stub error permanently disables migration detection. The catch also uses `e` instead of `error` and does not log the cause.
  - Catch as `error`, log it, and only cache `null` for expected unsupported-build failures. Leave `undefined` (or retry) for unexpected errors.
- **warning** `src/monero/MoneroEngine.ts:607`: Every advancing network height calls `updateBlockHeight`, which fires `onBlockHeightChanged` and walks the entire transaction list. Once heights catch up, `syncNetwork` still calls `queryTransactions` on every 1s poll until history backfill finishes, re-sorting the in-memory list after every new tx on the RN JS thread.
  - Stamp confirmations on each `EdgeTransaction` in `processTransaction` and skip a full-list height walk. Ingest pages without per-tx sorts (batch add, sort once), and/or pull several pages per tick.
- **warning** `422e70c4331cbc873469ce5363b5480a91474ac5:1`: Commit subject is 64 characters: “Persist the Zano wallet file once synced and run adopted wallets”. Edge commit subjects must be at most 50 characters.
  - Shorten to ≤50 characters, e.g. “Persist synced Zano wallets and run adopted ones”.
- **suggestion** `src/zano/ZanoTools.ts:121`: Passphrase-protected seed handling is duplicated in `importPrivateKey` and `derivePublicKey`: `zano.init`, `getSeedPhraseInfo`, and the same error_code/empty-address check. The two sites already drift in throw messages.
  - Extract a private helper that inits, calls `getSeedPhraseInfo`, and returns the address or throws.
- **suggestion** `src/monero/moneroTypes.ts:187`: `asAddressInfoResponse` requires many fields the engine never reads; extra required fields make LWS responses fail even when birthday height is present.
  - Keep only `start_height` (or mark unused fields optional) so the cleaner matches actual consumption.
- **suggestion** `src/zano/ZanoEngine.ts:789`: The `transfers.length === 0` guard in `makeMaxSpend` is unreachable: empty `tokenIds` already throws, and each token either pushes a transfer or throws.
  - Delete the dead check.

src/zano/ZanoEngine.ts
paullinator: **Warning:** Native `run_wallet` and `store` JSON is `JSON.parse`d before `asMaybe(...)`, so invalid JSON throws `SyntaxError` instead of failing the cleaner the same way a shape mismatch does.

**Recommendation:** Use `asMaybe(asJSON(asRunWalletResponse))(runResponse)` and `asMaybe(asJSON(asStoreResponse))(response)` and drop the separate `JSON.parse`.

src/zano/ZanoEngine.ts
paullinator: **Suggestion:** The adopt-path comment states as fact that 'react-native-zano opens wallets with the refresh worker postponed'. This package still depends on `react-native-zano` `^0.4.0`, where open auto-starts the worker and `run_wallet` is a no-op.

**Recommendation:** Tie the comment (and the dependency) to postponed-run from 0.4.1. On 0.4.0 the same `run_wallet` call is harmless; adopt must `run_wallet` because that configure flag is process-wide and sticky.

paullinator (changes requested):
Requesting changes on the remaining review items.

## Additional Findings

- **warning** `src/zano/ZanoEngine.ts:461`: `resyncBlockchain` now zeros `lastStoreTime`/`lastStoreHeight` (the Bugbot gap), but an in-flight `storeWalletFile` from the previous synced poll can still complete afterwards and write the old tip back into those fields. The rebuilt wallet then hits the height gate at the same tip and never stores until the chain advances, so a later kill pays the full catch-up this throttle exists to avoid.
  - Bump a store generation in `resyncBlockchain` and ignore lastStore updates unless the generation still matches after `invoke`. Alternatively drain/await the in-flight store before resetting the gates.
- **warning** `422e70c4331cbc873469ce5363b5480a91474ac5:1`: Commit subject is 64 characters: “Persist the Zano wallet file once synced and run adopted wallets”. Edge commit subjects must be at most 50 characters.
  - Shorten to ≤50 characters, e.g. “Persist synced Zano wallets and run adopted ones”.

src/zano/ZanoEngine.ts
```diff
@@ -413,12 +448,54 @@ export class ZanoEngine extends CurrencyEngine<
...
+   * Persists the native wallet file, so the next launch resumes from this
+   * height instead of re-scanning everything since the file was last
+   * written. Called only from the synced branch of `syncNetwork`, where the
+   * refresh worker is idle and the per-wallet lock is free; a store during
+   * the catch-up scan would block behind that lock. Throttled, and skipped
+   * entirely when the wallet height has not advanced past the last store. A
+   * failure only costs the next launch a longer catch-up, so it logs and
+   * moves on.
+   */
+  private async storeWalletFile(
```
paullinator, line 509: **Warning:** `resyncBlockchain` now zeros `lastStoreTime`/`lastStoreHeight` (the Bugbot gap), but an in-flight `storeWalletFile` from the previous synced poll can still complete afterwards and write the old tip back into those fields. The rebuilt wallet then hits the height gate at the same tip and never stores until the chain advances, so a later kill pays the full catch-up this throttle exists to avoid.

**Recommendation:** Bump a store generation in `resyncBlockchain` and ignore lastStore updates unless the generation still matches after `invoke`. Alternatively drain/await the in-flight store before resetting the gates.

paullinator (changes requested):
Moving the in-flight `storeWalletFile` warning onto the new method.

## Additional Findings

- **warning** `422e70c4331cbc873469ce5363b5480a91474ac5:1`: Commit subject is 64 characters: “Persist the Zano wallet file once synced and run adopted wallets”. Edge commit subjects must be at most 50 characters.
  - Shorten to ≤50 characters, e.g. “Persist synced Zano wallets and run adopted ones”.

src/zano/ZanoEngine.ts
j0ntz: Switched both to `asMaybe(asJSON(...))` and dropped the separate `JSON.parse`, so malformed JSON now fails the cleaner the same way a shape mismatch does instead of throwing SyntaxError past the guard.

src/zano/ZanoEngine.ts
j0ntz: Reworded: the comment now ties postponed-run to react-native-zano 0.4.1, says the flag is process-wide and sticky so the adopt path cannot know how the wallet was opened and must run it itself, and notes the call is a harmless no-op on 0.4.0 where open auto-starts the worker. Left the dependency range at ^0.4.0 for that reason: the call is safe on both, so nothing here needs 0.4.1 to function.

src/zano/ZanoEngine.ts
j0ntz: Took the store-generation option. `resyncBlockchain` bumps `storeGeneration` alongside zeroing the gates, `storeWalletFile` captures it before the invoke, and it returns without recording `lastStoreTime`/`lastStoreHeight` when the generation changed underneath. So an in-flight store from the pre-resync poll can no longer gate the rebuilt wallet at a tip it has not re-reached.

j0ntz:
Commit subject is now 48 characters: `Persist synced Zano wallets and run adopted ones`.

The three inline threads from this review round are fixed and resolved: the store/run_wallet responses go through `asMaybe(asJSON(...))`, the adopt-path comment is tied to postponed-run from 0.4.1 (the range stays `^0.4.0` because the call is a no-op there), and `resyncBlockchain` bumps a store generation that `storeWalletFile` re-checks before recording `lastStoreTime`/`lastStoreHeight`.

The wider findings in your dismissed review of the same day are not addressed here. Two of them are real and Zano-side (`onStart` awaiting `keysPromise` with no abort, and the adopt path throwing on an address mismatch without closing the handle); the Monero and Zcash engine items sit outside this PR's diff. Say if you want any of them folded in and I will take them, otherwise they are worth their own task.

j0ntz:
## 📸 Test evidence: Zano send between two wallets on iOS sim

**send success**

send success

**both wallets updated**

both wallets updated

**receiver incoming tx**

receiver incoming tx

**receiver confirming**

receiver confirming

**sender confirmed**

sender confirmed

**receiver confirmed**

receiver confirmed

j0ntz:
## 📸 Test evidence: receive notification (after fix)

**dropdown pending receive**

dropdown pending receive

**receive dropdown**

receive dropdown

j0ntz:
Closing in favour of EdgeApp/edge-currency-accountbased#1092, which supersedes this.

That branch carries all three commits from here forward with their authorship intact. `Notify on unconfirmed Zano receives` is byte-identical, `Persist synced Zano wallets and run adopted ones` differs only in a reworded comment, and `Fix Zano transaction query spin and pending txs` moves the mempool sweep after the paging loop instead of before it, trading one redundant round trip for a fresher pool snapshot at the end of a long pass.

On top of that it adds the five-minute catch-up checkpoint (the store here only ever fires for a wallet that reaches synced, which is the case that needs it least) and the HF6 payment-id handling.

The test evidence and the review history stay readable here. The design doc moved to EdgeApp/react-native-zano#19 so it outlives this branch.

PR #1092, opened 2026-08-24
Checkpoint Zano catch-up progress and deliver payment ids via integrated addresses

peachbits:
### Technical Design Document

[zano-migration-off-module-queue.md](https://github.com/EdgeApp/react-native-zano/blob/jon/zano-migration-tdd/src/docs/zano-migration-off-module-queue.md)

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

EdgeApp/react-native-zano#18 — needs a `react-native-zano` release carrying the
postponed-run support and the `TransferParams.paymentId` removal. The pin here
is still `^0.4.0`; bumping it is a merge step once that PR lands and publishes.

### Description

Addresses the performance regression QA found after the HF6 merge, plus the
HF6 payment-id correctness gap and a set of transaction-query bugs.

Related: #1090 — this branch carries forward the correct parts of that PR and
takes a different approach where that one was incorrect. Left open for now;
no action taken on it.

**Checkpoint catch-up progress every five minutes**

The Zano SDK only writes the wallet file on close, and the refresh worker holds
the per-wallet lock for the whole catch-up scan, so a plain `store` cannot land
until the scan finishes — exactly when it is no longer needed. A wallet weeks
behind that is killed mid-scan therefore persists nothing and re-pays the entire
scan on next launch.

This cycles the wallet through the lifecycle manager on an interval: the close
interrupts the refresh worker at its next block chunk, stores the partial state,
and the reopen resumes from the stored height. Gated on the height having
advanced, so a wallet stalled on a dead daemon does not churn, and guarded by a
store generation so a resync racing the reopen cannot write stale gates back.

**Deliver payment ids by folding them into the destination**

Since HF6 the tx-wide payment id is rejected by the network; ids ride inside
integrated addresses. Exchanges still hand users a plain address and an id
separately, so `makeSpend` now combines them into an integrated address on the
user's behalf when the id is the 8 hex bytes that can be encoded, and refuses
anything else rather than padding or truncating it — a transformed id would not
match the receiver's crediting ledger, turning a loud failure into a silently
lost deposit.

Receive-side parsing is unchanged, so the memo option stays visible for
transaction display. Note `createIntegratedAddress` silently *replaces* the
embedded id if handed an already-integrated address, so the resolver detects
integrated input first and only accepts a matching id.

**Persist synced Zano wallets and run adopted ones**

Stores the wallet file once synced, throttled and height-gated. The adoption
path (recovering an already-open wallet) now explicitly runs it, which the
postponed-run change makes necessary.

**Fix Zano transaction query spin and pending txs**

The paging loop could spin without terminating; it now ends on an empty page,
on no forward progress, or on reaching the end. The mempool sweep moved to after
the paging loop so pending transactions are not repeatedly clobbered by pages
of confirmed history.

**Notify on unconfirmed Zano receives**

Adds an `isTransactionNew` override mirroring `MoneroEngine` and `ZcashEngine`
so unconfirmed receives raise a notification.

### Testing

Unit tests cover the payment-id resolver, the checkpoint gating (including the
resync race), and the query loop.

Beyond that, this was run on device in a three-stage ladder on an account with
four mainnet Zano wallets created April–May 2025, each carrying a ~760k-block
catch-up window. Every stage used the same wallet-file fixture, and the last two
stages are the *same* GUI commit differing only in which packages are installed,
so the comparison isolates these changes.

With the published packages: nothing written to any wallet file across 8.5
minutes and ~72,000 blocks scanned, and the refresh worker sat in
`wallet2::refresh` → `pull_blocks` for 2269 of 2269 thread samples.

With this branch: all four wallets checkpointed at the five-minute mark
(527 bytes → 45–88 KB), staggered a few seconds apart rather than piling up, and
repeated on schedule in the next session. After a hard kill with no clean close,
the scan resumed forward from the checkpoint instead of restarting — roughly
52,000 blocks of scan work preserved per interval. Responsiveness was unchanged;
Zano-attributable mutex waits stayed at 0–1 frames per sample in both stages.

Not covered by that run: the startup freeze path, which needs an account whose
wallets are not already marked in `securityCheckedWallets`, and the legacy→HF6
wallet-file migration, which needs a fixture holding real scan history. Both are
being followed up separately.

Note for anyone debugging this later: a successful checkpoint logs at info, so
it is filtered out by the GUI's `defaultLogLevel: 'warn'` — only the failure
paths surface, which is intended. To watch checkpoints land, lower the level
rather than raising the severity of the message. Note the log *source* for an
engine is `zano-<walletSuffix>` (e.g. `zano-48`), not the bare plugin id, so a
`sources: { zano: 'info' }` override does NOT match — verified on device. Use
`defaultLogLevel: 'info'`, or key `sources` by the exact per-wallet source.

Otherwise the wallet file itself is the evidence: it grows from a few hundred
bytes to tens of KB the first time a checkpoint fires.


src/zano/ZanoEngine.ts
```diff
@@ -402,23 +497,139 @@ export class ZanoEngine extends CurrencyEngine<
...
+      return
+    }
+    if (now - this.lastCheckpointTime < CATCHUP_CHECKPOINT_INTERVAL_MS) return
+    if (walletHeight <= this.lastCheckpointHeight) return
+
+    this.log(
+      `checkpointCatchup: persisting catch-up progress at height ${walletHeight}`
+    )
+    const generation = this.storeGeneration
+    this.nativeId.stop()
```
j0ntz, line 595: **Warning:** the checkpoint closes the native handle out from under any caller that already resolved `nativeId` and is awaiting a native call on it. Before this PR `stop()` only ran on an explicit resync or kill, so no background timer could do that.

`broadcastTx` is the one that costs the user: it resolves `nativeId`, then awaits `transfer`. A wallet still catching up is exactly when a checkpoint is due, so the gap is reachable.

```mermaid
sequenceDiagram
    participant Send as broadcastTx
    participant Mgr as nativeId manager
    participant Sync as syncNetwork
    Send->>Mgr: get() returns id 7
    Send->>Send: await transfer(7, ...)
    Sync->>Mgr: checkpointCatchup calls stop()
    Mgr->>Mgr: close 7, reopen as 8
    Send->>Send: transfer(7) hits a closed handle
```

A `spendPending` flag set around `broadcastTx` that makes `checkpointCatchup` return early would close it, at the cost of deferring one checkpoint.

src/zano/ZanoEngine.ts
```diff
@@ -242,35 +296,76 @@ export class ZanoEngine extends CurrencyEngine<
...
+    // Sweep the mempool last. `get_recent_txs_and_info2` prepends the
+    // wallet's unconfirmed transfers only when the offset is zero, so once a
+    // wallet has any confirmed history the paged catch-up above can never
+    // see a transaction before it is mined: an incoming transfer stays
+    // invisible to the receiver, and the sender has nothing to confirm its
+    // own. Sweeping after the paging rather than before means the pool
+    // snapshot is as fresh as possible -- the loop can run long on a large
+    // catch-up, and a transfer that arrives mid-pass is caught this cycle
+    // instead of next.
+    const pendingPage = await this.tools.zano.getTransactions(nativeId, 0)
```
j0ntz: When the paging loop above ran at offset 0 (a new wallet, or one whose cursor never advances past it), it already fetched and processed this exact page, so the sweep repeats the same RPC and the same `processTransaction` pass on every tick. Guarding with `if (offset !== 0)`, or reusing the page the loop already holds, skips the duplicate.

src/zano/ZanoEngine.ts
```diff
@@ -242,35 +296,76 @@ export class ZanoEngine extends CurrencyEngine<
...
+      const page = await this.tools.zano.getTransactions(nativeId, offset)
+      const transfers = page.transfers ?? []
+      transfers.forEach(this.processTransaction)
 
-      if (offset !== transactions.last_item_index) {
-        this.otherData.transactionQueryOffset = transactions.last_item_index
+      const lastItemIndex = page.last_item_index
+      const totalTransfers = page.total_transfers
+      if (offset !== lastItemIndex) {
+        this.otherData.transactionQueryOffset = lastItemIndex
```
j0ntz, line 365: Not introduced here: A `resyncBlockchain` that lands while the `getTransactions` above is in flight has already emptied the cache and zeroed the cursor, and this write puts the stale high cursor back, so the cleared history is never re-fetched. The pre-PR loop had the same window. The `storeGeneration` guard you added for the store gates would close it here unchanged, if you want it while you are in this function.

src/zano/ZanoEngine.ts
```diff
@@ -402,23 +497,139 @@ export class ZanoEngine extends CurrencyEngine<
...
   // // ****************************************************************************
 
   async resyncBlockchain(): Promise<void> {
     this.needsNativeStorageClear = true
+    // The rebuilt wallet re-earns every height, so the store gates must not
+    // carry over: left in place they would skip the first store at the
+    // re-reached tip, and a kill in that window re-pays the full rescan.
+    this.lastStoreTime = 0
+    this.lastStoreHeight = 0
+    this.lastCheckpointTime = 0
```
j0ntz: These two lines are `resetCatchupCheckpoint()`. Calling it keeps the reset in one place, so a later gate field does not have to be remembered in two.

test/zano/ZanoEngineQueryTransactions.test.ts
```diff
@@ -0,0 +1,285 @@
...
+interface TestEngine {
+  engine: ZanoEngine
+  /** Every transaction event the engine handed to the core, in order. */
+  events: EdgeTransactionEvent[]
+  queryTransactions: () => Promise<void>
+  requests: number[]
+  storedTx: (txid: string) => EdgeTransaction | undefined
+}
+
+async function makeEngine(
```
j0ntz, line 119: `makeEngine` here is the same callbacks / engine-options / `PluginEnvironment` block as `ZanoEngineCheckpoint.test.ts`, differing only in the fake `tools` and `nativeId` tail. `test/fake/` already holds `FakeTools.ts` and `fakeLog.ts`, so a shared factory there would keep the next `EdgeCurrencyEngineOptions` change from breaking two near-identical copies.

j0ntz (commented):
Reviewed alongside EdgeApp/react-native-zano#18. Nothing I found blocks the merge, and the one warning below is the item worth acting on before this ships. The checkpoint mechanism and the integrated-address payment-id resolution both read correctly to me, and `resolvePaymentIdDestination` handling the already-integrated case before `createIntegratedAddress` is the right order given that function silently replaces an embedded id.

Separately, two commit subjects are over the 50-character Edge limit that Paul enforces in review: `fixed: (Zano) Checkpoint catch-up progress every five minutes` (61) and `fixed: (Zano) Deliver payment ids by folding them into the destination` (70).

j0ntz:
The design doc for this work is now in EdgeApp/react-native-zano#19, rewritten to describe what ships here rather than the first round.

Section 6 covers this repo: the store gates and the `storeGeneration` guard, `checkpointCatchup` and why a plain `store` cannot land mid-scan, the query loop's two SDK properties, the `isTransactionNew` override, and the integrated-address payment-id rules. Section 8's phase 8 records what changed against #1090, and section 9 carries the two new decisions (checkpoint by closing, refuse a payment id that will not fit).

The doc lived only on `jon/ios-perf-zano-xmr` in react-native-zano until now, and that branch goes away with #17.

j0ntz:
Pushed `e64142b7` here, which is new scope rather than a review response: the confirmation count QA reported (the app showing "9 of 10 Confirmations" on a transaction the explorer put at 7).

The cause is a unit mismatch in `syncNetwork`, and it pre-dates this branch (`master:src/zano/ZanoEngine.ts:387-390` has the same shape), so nothing already here could have taken it. The wallet status reports `current_wallet_height` from `get_top_block_height()`, which is `m_local_bc_size - 1`, a block height; `current_daemon_height` is the daemon's `getinfo` height, which counts blocks. Probing the plugin's own node at 37.27.100.59:10500: `getinfo` answered 3831817 while `getlastblockheader` put the tip block at 3831816. Taking `Math.max` of the two therefore put a synced wallet one block past the tip, and both confirmation formulas are `walletBlockHeight - txBlockHeight + 1`, so every count read one high and a transaction reached confirmed a block early. The explorer showing depth rather than confirmations is the second unit of the reported gap.

The fix converts the count to a height once, floored at zero for a disconnected daemon, and uses it for the block height and the sync-progress detail. `test/zano/ZanoEngineBlockHeight.test.ts` covers it: against the previous engine those cases fail with `expected 3000001 to equal 3000000` and `expected 2 to equal 1`.

src/zano/ZanoEngine.ts
peachbits: Done -- `spendPending` wraps the whole of `broadcastTx` (set before the nativeId resolve, cleared in a finally), and `checkpointCatchup` defers while it is set without touching its gates, so it fires on the first tick after the spend clears instead of waiting out a fresh interval. Tested, including the fires-next-tick half.

src/zano/ZanoEngine.ts
peachbits: Done -- the sweep is skipped when the loop's final page was offset zero, since that fetch already carried the unconfirmed transfers. A test pins exactly one fetch for a wallet whose cursor sits at zero.

src/zano/ZanoEngine.ts
peachbits: Took the suggestion -- `queryTransactions` captures the generation up front and bails after any fetch a resync raced, so neither the stale page nor the old cursor lands. The test models the resync inside the fetch.

src/zano/ZanoEngine.ts
peachbits: Done.

test/zano/ZanoEngineQueryTransactions.test.ts
peachbits: Done -- `test/fake/fakeZanoEngine.ts` owns the boilerplate now; both engine test files consume it (the query test passes an `onTransactions` hook, the checkpoint test attaches its recording nativeId stub).

peachbits:
Both flagged subjects reworded (42 and 45 characters).

j0ntz:
## 📸 Test evidence: confirmation count, before and after

**confirmation count**

confirmation count

**prefix count**

prefix count

j0ntz (approved):
Ensure fixups squash correctly, they're not `fixup!` tagged

PR #1095, opened 2026-08-27
Fix the Zano sync ratio and read HF6 history

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

EdgeApp/react-native-zano#20 — the final commit reads history through the
`get_recent_txs_and_info3` support added there. `react-native-zano` is
temporarily pinned to that branch's commit hash in `devDependencies` so CI
builds against it; the pin moves to the released version once
react-native-zano publishes.

### Description

Fixes three defects in the Zano sync ratio, found while verifying the
Android checkpoint work. All three shipped in 4.89.0; the GUI's
`WalletSyncCircle` animates whatever the engine sends — it neither clamps
nor ratchets — so each was user-visible.

**Commit 1 — the history ratio was computed upside down.**
`queryTransactions` reported `total_transfers / last_item_index`, the
reciprocal of the fraction fetched, so a wallet with more than one page of
history pushed unclamped ratios above 1 that shrank toward 1 as pages
loaded. A ratio past the circle's done threshold fades it out as fully
synced while the fetch is still running. The loop's break conditions
guarantee the corrected fraction stays inside (0, 1); wallets whose history
fits one page never hit the update, which is why it went unseen.

**Commit 2 — the ratio latched at 1 on startup, and saw-toothed if it did
not.**

*The latch:* the synced branch keyed on the SDK's `wallet_state`, which
reads "ready" from open until the refresh worker's first pass and again
between passes (`wallets_manager.cpp` sets it at `init_wallet_entry`; only
the worker flips it). One 1-second poll landing in that window read a
freshly restored wallet — weeks of blocks behind — as fully synced: the
blended ratio hit exactly 1, `syncComplete` latched, and every later update
was suppressed for the life of the engine. Reproduced live on an emulator:
a rebuilt wallet scanned for six minutes with no sync circle and zero
sync-status events while its first catch-up checkpoint landed on schedule.
The spurious synced pass also reset the checkpoint gates and issued a store
mid-scan. Sync is now judged by heights alone, with a zero daemon height
treated as "not connected" rather than as a tip the wallet has passed.

*The sawtooth:* the catch-up branch forwarded the SDK's `progress`, which
restarts at zero on every refresh pass (`wallet2::refresh` resets
`m_last_sync_percent` and rebases its start height on entry) — that is,
after every checkpoint reopen and every reconnect — and only measures the
remaining gap. With the latch fixed, the circle would visibly wind backward
every few minutes of a long sync. The ratio is now measured against the
height where the catch-up episode began, which is stable across restarts,
and rebases when a resync drops the wallet below it.

**Commit 3 — measure a new wallet's catch-up from its birthday.**

The episode baseline in commit 2 can be captured before the wallet's first
pull, and a freshly restored wallet reports height ~0 until that pull skips
straight to the account birthday — so the baseline could read the skip
itself as most of the work: a birthday of 75,000 against a tip of 100,000
showed ~75% the moment the skip landed. The work to be done is
birthday-to-tip: that moment is 0%, and reaching 80,000 is 20% (which is
also what the SDK's own meter reported before checkpointing, since it
rebases its session start to the birthday).

The baseline now chases the wallet height while the SDK's session progress
reads zero and freezes at the first tick of nonzero progress. The birthday
skip happens entirely inside the progress-zero window — the SDK's meter
starts at the birthday too — so the frozen baseline lands at the birthday,
within 1% of the gap (session progress is an integer percent). Checkpoint
reopens reset session progress, but the frozen flag holds, so the baseline
never rebases mid-episode. Resumed wallets are unaffected: nothing to skip,
and the ratio keeps measuring the remaining work.

**Commit 4 — read history from the HF6 endpoint, with payment id memos.**

Rides the react-native-zano switch to `get_recent_txs_and_info3`. The old
endpoint reported only the deprecated transaction-wide payment id — an
empty string for everything sent since HF6, which also attached a blank
`Payment ID` memo to every Zano transaction — and its legacy serializer
refuses entries carrying more than one distinct per-output id. Amounts are
now summed across the per-payment-id groups (one Edge transaction per
asset, exactly as before), each distinct non-empty id becomes a hex
`Payment ID` memo, and the empty-id group — where a sender's own spent
inputs and change always land — produces none. A sent transaction keeps
the memos it was saved with at broadcast: its own history entry cannot
carry the recipient's id, and the rebuild used to replace the stored
record wholesale, erasing the id within seconds of the first mempool
sweep.

Verified with a live on-chain spend between two wallets carrying payment
id `a1b2c3d4e5f60718`: the receiver's entry delivers the id verbatim as a
hex group before and after confirmation and it renders as the deposit's
`Payment ID` in transaction details, extracted purely from the chain; the
sender's entry nets its spend in the empty-id group with the recipient's
id structurally absent, and the saved memo survives every history rebuild
including the confirmation update.

**Verification.** Eight new unit tests: the history fraction climbs and
never passes 1; a fresh wallet's ready state is not read as synced; the
ratio is monotonic across restarts that reset SDK progress; a new wallet
measures from its birthday (heights 1 → 75,000 → 80,000 → 87,500 assert
ratios 0, 0, 0.2, 0.5); synced is reported exactly when the wallet reaches
the tip and a later block burst baselines a fresh episode; a resync
rebases; a disconnected daemon is not synced; a deposit's intrinsic
payment id becomes its memo while the empty-id group produces none. Full suite green. The latch
was demonstrated on-device before the fix; a post-fix on-device trace is
pending QA hardware (the test emulator's network stack gave out).

---

> [!NOTE]
> **Medium Risk**
> Changes Zano sync UX logic and transaction memo/history parsing on a privacy coin engine; behavior is heavily unit-tested but affects long syncs and HF6 payment-id display.
> 
> **Overview**
> Fixes **Zano sync progress** reporting and switches history to the **HF6** API, with **`react-native-zano` ^0.5.1**.
> 
> **Sync ratio:** Catch-up progress no longer trusts the SDK’s transient “ready” state or session `progress` (which reset on checkpoint/reconnect). **Synced** is decided from **wallet vs daemon height**; catch-up uses an **episode baseline** that chases the wallet through the birthday skip, stays monotonic across checkpoint restarts, and resets via `resetCatchupState`. **History paging** now reports `fetched / total` instead of the inverted fraction that could exceed 1 and hide the sync circle mid-fetch.
> 
> **History / memos:** Amounts and **payment IDs** come from `subtransfers_by_pid` (via `get_recent_txs_and_info3` in the native module). Non-empty per-output IDs become hex **`paymentId`** memos; empty groups no longer add bogus IDs. **Outgoing** txs **preserve broadcast-time memos** when history rebuilds would drop the payment id (`findSavedMemos`).
> 
> **Tests:** New `ZanoEngineSyncRatio.test.ts` plus coverage for payment-id memos and history ratio behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 879319d74b714ef1cfa1d0d7fcc3cfcc9fb93820. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/zano/ZanoEngine.ts
```diff
@@ -528,8 +564,41 @@ export class ZanoEngine extends CurrencyEngine<
...
+      // reopen, and on reconnects -- and only measures the remaining gap,
+      // so it saw-toothed the GUI's circle backward every few minutes. The
+      // height where this episode began is stable across all of that. A
+      // resync drops the wallet below the baseline, so rebase when it does:
+      if (
+        this.catchupStartHeight < 0 ||
+        this.catchupStartHeight > status.current_wallet_height
+      ) {
+        this.catchupStartHeight = status.current_wallet_height
+        this.catchupBaselineFrozen = false
```
j0ntz, line 624: **Warning:** a checkpoint reopen can rebase and unfreeze the baseline mid-episode, which the comment just below says cannot happen.

That comment ends "Checkpoint reopens reset session progress to zero, but the frozen flag holds, so the baseline never rebases mid-episode." It holds only while the wallet height never dips. `checkpointCatchup`'s own docstring prices its restart at "one re-fetched block chunk", so the reopened wallet can report a `current_wallet_height` below where it was before the close. If that lands below a frozen baseline, this branch fires, lowering the baseline and clearing `catchupBaselineFrozen`.

The unfreeze is the expensive half. Session progress is also zero right after a reopen, so the chase below re-engages and absorbs every block scanned until progress ticks back to 1 percent of the remaining gap, reporting ratio 0 the whole time. On a slow chain that is thousands of blocks, which is the case `SYNC_STATUS_INTERVAL_MS` in `WeightedSyncTracker` was written for.

The window is narrow: it needs the interval's scanning to have advanced less than one re-fetched chunk. Keeping `catchupBaselineFrozen` set when the dip is small, or distinguishing a resync (height collapses toward the birthday) from a chunk-sized dip, would close it.

src/zano/ZanoEngine.ts
```diff
@@ -528,8 +564,41 @@ export class ZanoEngine extends CurrencyEngine<
...
+          this.catchupBaselineFrozen = true
+        } else if (status.current_wallet_height > this.catchupStartHeight) {
+          this.catchupStartHeight = status.current_wallet_height
+        }
+      }
+      const scanned = status.current_wallet_height - this.catchupStartHeight
+      const target = daemonBlockHeight - this.catchupStartHeight
       this.syncTracker.updateBlockRatio(
-        status.progress / 100,
+        target > 0 ? scanned / target : 0,
```
j0ntz: **Suggestion:** `MoneroEngine` already has this baseline, with a clamp this version drops.

`src/monero/MoneroEngine.ts:616-665` captures a `syncStartHeight` on the first catching-up tick and computes `(syncedHeight - syncStartHeight) / (networkHeight - syncStartHeight)`, guarded the same way on a positive range, but wrapped in `Math.max(0, ...)` with the reason inline: "Clamp to 0 so a reorg (syncedHeight < syncStartHeight) can't feed a negative ratio into the weighted sync tracker."

Here non-negativity rests entirely on the rebase check above running before this line on every tick. Nothing states that ordering, so an edit that moves the computation up, or adds an early return between them, feeds a negative ratio straight into the tracker. `Math.max(0, ...)` costs nothing and makes the invariant local.

The two engines now solve the same problem twice, with Zano adding the birthday-skip freeze on top. Extracting a shared `src/common/` helper is worth doing once a third SDK needs it, not necessarily in this PR.

src/zano/ZanoEngine.ts
```diff
@@ -117,6 +117,29 @@ export class ZanoEngine extends CurrencyEngine<
...
+  private catchupStartHeight: number = -1
+
+  /**
+   * Whether `catchupStartHeight` has stopped chasing the wallet height for
+   * this episode. Set on the first tick where the SDK reports nonzero
+   * session progress -- real scanning, as opposed to the birthday skip,
+   * which happens entirely inside "progress 0" territory because the
+   * SDK's own meter starts at the birthday too.
+   */
+  private catchupBaselineFrozen: boolean = false
```
j0ntz, line 151: **Suggestion:** two lifecycle gaps in this episode state, both minor.

1. `resyncBlockchain` resets `lastStoreTime`, `lastStoreHeight`, and the checkpoint gates for the reason its comment gives ("The rebuilt wallet re-earns every height, so the store gates must not carry over"), but not these two fields. The rebase check usually covers it, since the rebuilt wallet restarts at or below the old baseline. It stops covering it when the rebuild does not happen: `deleteWallet` in the lifecycle manager's `onStart` logs its error rather than surfacing it, so the wallet can reopen at the pre-resync height with a stale frozen baseline still driving the ratio. Adding these two to that reset block makes it unconditional and matches the comment already there.

2. Nothing persists these fields, so an app relaunch mid-catch-up baselines at the resumed height and the ratio restarts near zero even though the scan is genuinely half done. Not a regression, since `status.progress` was session-scoped too, but `otherData` already persists `transactionQueryOffset`, and the CHANGELOG's birthday example ("a birthday of 75,000 against a tip of 100,000 starts at 0%") only reads true on the first launch. Fine as a follow-up.

j0ntz (commented):
The inverted history ratio and the heights-only synced test both look right to me. The notes below are all on the catch-up baseline's lifecycle.

src/zano/ZanoEngine.ts
peachbits: Good catch, and the comment was the wrong half. Fixed in 6e4224dd: only a drop of more than `CATCHUP_EPISODE_MAX_DIP` starts a new episode, so a chunk-sized dip keeps both the baseline and the freeze.

I took the threshold from the SDK rather than inventing one. `wallet2` re-scans a reorg in place via `detach_blockchain` up to `WALLET_CONCISE_MODE_MAX_REORG_BLOCKS` (`CURRENCY_BLOCKS_PER_DAY * 7`, and `DIFFICULTY_TOTAL_TARGET` is 60s, so 10,080 blocks); past that it sets `wallet_reset_needed` and abandons the chain, which is exactly the case that deserves a new episode. A checkpoint reopen re-fetches at most one chunk, far inside it.

Worth adding: I could not find a path where a checkpoint reopen actually lowers the height. The store waits on the per-wallet lock the worker holds for the chunk, so it persists a completed chunk boundary at or above anything we polled, and the reopen loads that. The dip that does reach this branch is a reorg. Either way the branch now handles it.

src/zano/ZanoEngine.ts
peachbits: Added the clamp, with a comment pointing at the same reason MoneroEngine gives. Agreed the invariant should be local rather than depending on the rebase check running first.

Left the shared helper alone for now, per your own suggestion -- Monero measures from a plain first-tick baseline while Zano adds the birthday chase and the episode threshold on top, so the shapes have not converged yet. A third SDK is the right trigger.

src/zano/ZanoEngine.ts
peachbits: (1) Fixed. `resetCatchupCheckpoint` is now `resetCatchupState` and clears the baseline alongside the gates, so the synced path and `resyncBlockchain` both get all four fields and a new field cannot be added to one caller and forgotten in the other. That closes the silent-`deleteWallet` case you describe, where the wallet reopens at its pre-resync height.

(2) Agreed, and left as the follow-up you suggest. Persisting the baseline in `otherData` needs a little thought about what it means after a gap -- a wallet resumed hours later has a genuinely different amount of work to do than when the episode started -- so I would rather size that separately than bolt it onto this PR. You are right that the CHANGELOG example only reads true on the first launch; the entry now leads with the birthday case rather than implying it survives a relaunch.

j0ntz: Link/create followup task

j0ntz: https://app.asana.com/1/9976422036640/project/1213880789473005/task/1217559756673909?focus=true

j0ntz (approved):
Just fix the double dashes `--` in the committed code. Can just do a single dash `-`

src/zano/ZanoEngine.ts
```diff
@@ -429,7 +461,15 @@ export class ZanoEngine extends CurrencyEngine<
...
       }
-    } else {
+    }
+    for (const paymentId of paymentIds) {
+      memos.push({
+        memoName: 'Payment ID',
+        type: 'hex',
+        value: paymentId
+      })
+    }
```
j0ntz, line 471: **Suggestion:** `'Payment ID'` should be `'paymentId'`, and `'Comment'` on line 436 should be `'comment'`.

The GUI keys memos by the `memoOptions` name, not by a display string. `TransactionDetailsScene.tsx:625` renders `getMemoTitle(memo.memoName)`, which is `memoTitles[memoName] ?? memoName`, and that map in `util/memoUtils.ts` is keyed `comment`, `paymentId`, `destination tag`, `memo`, `note`, `op_return`. Neither `'Payment ID'` nor `'Comment'` is a key, so both fall through to the raw string. It looks right in English only because `memo_payment_id_title` is itself "Payment ID"; every other locale shows English on Zano transactions.

Using the `memoOptions` names also settles the divergence this commit's carry-forward otherwise creates. `findSavedMemos` returns the broadcast-time array, whose names come from `makeSpend` reading `currencyInfo.memoOptions`, so they are already `paymentId` and `comment` (`zanoInfo.ts:186-187`). As written, a send that takes the carry-forward renders localized while a receive next to it in the same wallet renders English.

The `Comment` half pre-dates this PR, so this is not a regression it introduced. The Payment ID emission is being rewritten here anyway, and the fix is the same two words.

j0ntz (commented):
Re-reviewed at 2677014. The three earlier findings are all addressed, and the baseline rewrite reads correct to me: the episode threshold, the clamp, and `resetCatchupState` covering all four fields from both callers.

One note below, plus one with no line to hang it on: the mempool-sweep comment at `ZanoEngine.ts:407` still names `get_recent_txs_and_info2`, the endpoint this PR retires.

j0ntz (approved):

src/zano/ZanoEngine.ts
```diff
@@ -491,6 +542,22 @@ export class ZanoEngine extends CurrencyEngine<
     }
   }
 
+  /**
```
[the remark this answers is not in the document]
peachbits: Valid finding, and correctly diagnosed on both halves: `addTransaction`'s update condition compares blockHeight, networkFee, nativeAmount, `otherParams.lastSeenTime` and date, so a memo-only difference takes the "no update" branch, and `queryTransactions` only walks forward from `transactionQueryOffset`. Not fixing in this PR.

Two corrections to the blast radius. The mempool sweep re-fetches offset 0 every tick, and that page carries the oldest ~100 confirmed entries alongside the unconfirmed ones, so for any wallet under ~100 transactions the older history *is* re-presented every tick - it just does not update, for the memo-diff reason above. So the accurate statement is "replayed but not updated" rather than "never replayed", with the same end state.

And the blank `Payment ID` memo only exists on rows written by a previous build; it came from the old `payment_id != null` guard against a field that was always present as `""`. Nothing processed by this build gets one. So this is stale stored rows, not new breakage - though the PR description's claim that the empty memo "is gone" is true going forward only, which is the fair core of this.

Declining the fix deliberately: widening the shared `addTransaction` comparison to include memos would change update semantics for every currency in the repo to serve one chain's backfill, and a Zano-only migration that rewinds `transactionQueryOffset` would re-pay a full history walk for a cosmetic gain. The residue is an empty memo row on pre-upgrade transactions and missing intrinsic ids on pre-upgrade deposits, both of which self-correct on any resync.

PR #1098, opened 2026-09-09
Read Solana version 1 transactions

j0ntz:
### Description

[Asana task](https://app.asana.com/0/0/1218333177692561/f)

📄 [Solana version 1 transactions: history keeps syncing past a v1 row](https://github.com/EdgeApp/edge-currency-accountbased/blob/jon/solana-v1-transactions/src/docs/solana-v1-transactions.md)

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Test evidence

<table>
<tr>
<td rowspan="1" width="170" valign="top"><a href="https://github.com/EdgeApp/edge-currency-accountbased/pull/1098/commits/f1622c630ee976c2bc420bdd7b9b95012a4bff98"><code>f1622c6</code></a><br><sub>Fix flaky builtinTokens and FTM fee tests</sub><br><sub>🪓 <em>local uncommitted accb harness pointed Solana at Helius devnet with a v1-only address and wrapped the archive connections: a node answering null ahead of the real one (09), then -32015 on one node and an outage on the next for the newest row (10); reverted, git status clean</em></sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1098/20260916-041309-agent-proof-1218333177692561-09-HACKED-lagging-node-null-history.png"></a><br><sub>🪓 lagging node null history</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-currency-accountbased/pr-1098/20260916-041309-agent-proof-1218333177692561-10-HACKED-unsupported-then-outage-history.png"></a><br><sub>🪓 unsupported then outage history</sub></td>
<td width="210"></td>
</tr>
</table>


j0ntz:
## 📸🪓 Test evidence

> 🪓 **Hack-forced evidence:** pinned the Solana plugin's RPC nodes to https://api.devnet.solana.com and hard-coded SolanaEngine's watched address to the devnet account 3gpqjMYx8S1Pqx7cnG1oQpTwxoSD3MZVyiQooHkAaUCf, whose entire history is version 1 transactions; both edits reverted, git status clean. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: solana v1 history**

🪓 HACK-FORCED: solana v1 history

**🪓 HACK-FORCED: v1 transaction detail**

🪓 HACK-FORCED: v1 transaction detail

**🪓 HACK-FORCED: v1 txid**

🪓 HACK-FORCED: v1 txid

src/solana/SolanaEngine.ts
```diff
@@ -614,12 +664,20 @@ export class SolanaEngine extends CurrencyEngine<
...
+          })
+        } catch (error: unknown) {
+          // A transaction we cannot interpret costs one history row. Advancing
+          // the watermark past it keeps the rest of the history flowing.
+          this.error(
+            `Could not process transaction ${tx.transaction.signatures[0]}: `,
+            error instanceof Error ? error : new Error(String(error))
+          )
+        }
+        this.otherData.newestTxid[safeTokenId] = tx.transaction.signatures[0]
```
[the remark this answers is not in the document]
j0ntz: Fixed. `fetchTransactionChunk` now returns an `unavailable` set holding the signatures no node would answer for, and the first one freezes `newestTxid` for the rest of the pass, so a later chunk's success can no longer carry the next sync's `until` past rows we never got. A row that fails to parse still advances the watermark, since retrying a message the client cannot deserialize is the frozen wallet this PR removes, and a null result from a node is an answer rather than an outage, so it stays out of the set. Two new cases in `test/solana/solanaV1Transaction.test.ts` cover the distinction.

j0ntz:
## 📸🪓 Test evidence (after the watermark fix)

> 🪓 **Hack-forced evidence:** pinned the Solana plugin's RPC nodes to https://api.devnet.solana.com and hard-coded SolanaEngine's watched address to the devnet account 3gpqjMYx8S1Pqx7cnG1oQpTwxoSD3MZVyiQooHkAaUCf, whose entire history is version 1 transactions; both edits reverted, git status clean. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: history after watermark fix**

🪓 HACK-FORCED: history after watermark fix

src/solana/SolanaEngine.ts
```diff
@@ -568,39 +638,45 @@ export class SolanaEngine extends CurrencyEngine<
...
-      > = await asyncStaggeredRace(funcs)
 
       // Process the transactions from oldest to newest
-      for (let i = 0; i < txResponse.length; i++) {
+      for (let i = 0; i < transactions.length; i++) {
         numProcessedTx++
-        if (txResponse[i].meta?.err != null) continue // ignore these
+        const tx = transactions[i]
+        if (tx == null) {
+          if (unavailable.has(transactionRequest[i])) sawRetryableGap = true
```
peachbits, line 672: `unavailable` treats every `getTransaction` throw as transient. Decision 7.2 rests on "an RPC that would not answer is transient", which holds for timeouts and 5xx but not for -32015, the error this PR exists to fix.

When Agave 4.4 ships version 2 transactions, `maxSupportedTransactionVersion: 1` produces -32015 again. The batch throws, the per-signature retry throws on every archive node, the signature lands in `unavailable`, `sawRetryableGap` latches, and `newestTxid` stops advancing. `asSolanaWalletOtherData` carries only `newestTxid`, so nothing counts attempts and nothing clears the latch on a later pass. That is the section 1 wall again, one version up.

Classifying the deterministic RPC errors separately before adding to `unavailable` would keep the freeze for real outages while letting a permanently unservable signature advance the way a parse failure already does.

src/solana/SolanaEngine.ts
```diff
@@ -614,12 +690,22 @@ export class SolanaEngine extends CurrencyEngine<
             memos.push(match[1])
           }
         }
-        const amounts = this.parseTxAmounts(txResponse[i], pubkey)
-        amounts.forEach(amount => {
-          this.processSolanaTransaction(txResponse[i], amount, timestamp, memos)
-        })
-        this.otherData.newestTxid[safeTokenId] =
-          txResponse[i].transaction.signatures[0]
+        try {
```
peachbits, line 708: The `try` spans both `parseTxAmounts` and the `forEach` that records each amount. A throw from `processSolanaTransaction` on the second of several amounts leaves the first already in `transactionList`, and line 706 still advances the watermark past the row.

Decision 7.2 accepts losing a whole row to a parse failure, but a half-recorded row is worse than a skipped one, because nothing downstream can tell it is incomplete and it never comes back without a full resync. Narrowing the `try` to `parseTxAmounts`, or building every `EdgeTransaction` before recording any of them, keeps the row all-or-nothing.

test/solana/solanaV1Transaction.test.ts
```diff
@@ -0,0 +1,228 @@
...
+      tx,
+      amount,
+      v1TransactionFixture.blockTime,
+      []
+    )
+  })
+  return engine.captured
+}
+
+describe('Solana version 1 transactions', function () {
```
peachbits, line 185: These five cases cover `fetchTransactionChunk` well, but they call it through `SolanaEngine.prototype.fetchTransactionChunk.call` and stop there. `newestTxid` appears nowhere under `test/`, so `queryTransactionsInner` and the `sawRetryableGap` freeze are untested.

That freeze is the riskiest behavior this PR adds, since it converts one bad row into a stalled wallet. Two cases would cover it: an unavailable signature in a chunk leaves `newestTxid` unchanged, and a clean chunk advances it.

peachbits (changes requested):
All three findings concern the new watermark freeze. The first two are ways `newestTxid` can stall or advance incorrectly, and the third is that the mechanism itself has no test coverage.

The `maxSupportedTransactionVersion` bump looks right and is well covered by the new fixture.

src/solana/SolanaEngine.ts
j0ntz: Fixed. `isUnsupportedTransactionVersion` reads the code off the `SolanaJSONRPCError` and keeps -32015 out of `unavailable`, so a version the client cannot read now advances the watermark like a parse failure, while timeouts, rate limits and unhealthy nodes still freeze it. I classified only -32015: -32011, -32009 and -32001 depend on one node's configuration or retention, and another archival node in the race may still hold the transaction, so treating them as permanent would bring back the silent loss the freeze prevents (decision 7.5 in the design doc).

src/solana/SolanaEngine.ts
j0ntz: Fixed. `processSolanaTransaction` is now the pure `makeSolanaTransaction`: the `try` builds every amount of the transaction, and `addTransaction` runs only after all of them have built, so a throw part way through records nothing. I kept the whole build inside the `try` rather than narrowing it to `parseTxAmounts`, since a throw while building would otherwise escape the loop and end the pass.

test/solana/solanaV1Transaction.test.ts
j0ntz: Added. The stub now answers `getSignaturesForAddress`, and four cases drive `queryTransactionsInner` through a fake engine: a clean chunk advances `newestTxid`, an outage before a good row holds it at empty while the row still records, -32015 moves past it, and a row whose second amount fails to build records nothing. Each one fails when its fix is reverted.

j0ntz:
All three findings are fixed: -32015 no longer freezes the watermark, a transaction's rows are recorded all-or-nothing, and `queryTransactionsInner` has watermark coverage. Decisions 7.2, 7.5 and 7.6 and the phase history in the design doc carry the reasoning.

j0ntz:
## 📸🪓 Test evidence (after the review fixes)

> 🪓 **Hack-forced evidence:** pinned the Solana plugin's RPC nodes to https://api.devnet.solana.com and hard-coded SolanaEngine's watched address to the devnet account 3gpqjMYx8S1Pqx7cnG1oQpTwxoSD3MZVyiQooHkAaUCf, whose entire history is version 1 transactions; both edits reverted, git status clean. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: history after review fixes**

🪓 HACK-FORCED: history after review fixes

**🪓 HACK-FORCED: v1 detail after review fixes**

🪓 HACK-FORCED: v1 detail after review fixes

src/solana/SolanaEngine.ts
```diff
@@ -568,39 +652,45 @@ export class SolanaEngine extends CurrencyEngine<
...
           const funcs = this.tools.archiveConnections.map(
             connection => async () => {
-              return await connection.getBlockTime(txResponse[i].slot)
+              return await connection.getBlockTime(tx.slot)
             }
           )
           const blocktimeRaw = await asyncStaggeredRace(funcs)
           const blocktimeClean = asMaybe(asBlocktime)(blocktimeRaw)
-          if (blocktimeClean == null) continue
+          if (blocktimeClean == null) {
```
peachbits: `asBlocktime` is `asRpcResponse(asNumber)`, so it expects `{ result: number }`, but `connection.getBlockTime` resolves to a bare `number | null`. `asMaybe` therefore returns null on every call, including one where a node answered with a valid timestamp, and `sawRetryableGap` latches unconditionally whenever this fallback runs.

With -32015 now handled, this is the only remaining path that can freeze the watermark permanently. The mismatch is deterministic, `asSolanaWalletOtherData` holds no attempt counter, and nothing clears the latch on a later pass, so any row lacking `blockTime` on both the signature list and the transaction body stalls history there on every sync.

Section 4 of the design doc describes this freeze as firing when "a `getBlockTime` lookup no node answered". That is the intended behavior, but not what the code does.

Bugbot reported the cleaner mismatch and that thread is marked resolved, but the code is unchanged at `46afe05f`. Reachability is narrow, since `getSignaturesForAddress` usually carries `blockTime`, but the failure is total when reached. The new `Solana history watermark` block is the natural place for a fifth case.

src/docs/solana-v1-transactions.md
```diff
@@ -0,0 +1,270 @@
...
+
+**Resilience.** The chunk fetch moves into `fetchTransactionChunk`, which keeps the batch as the fast path and retries the chunk one signature at a time when the batch throws. A signature that still fails becomes `null` in the returned array rather than an exception. It lands in the result's `unavailable` set only when asking again could change the answer, which means a node outage: a timeout, a rate limit, an unhealthy node. Two failures stay out of the set because every node gives the same answer on every sync:
+
+- A `null` result, which is a node saying the transaction is not there.
+- -32015, an unsupported transaction version. That is a property of our request rather than of the node, so it recurs until the client learns the next version. `isUnsupportedTransactionVersion` reads the code off web3.js's `SolanaJSONRPCError` and compares it to `SolanaJSONRPCErrorCode.JSON_RPC_SERVER_ERROR_UNSUPPORTED_TRANSACTION_VERSION` ([7.5](#75-treat-only-an-unsupported-version-as-a-permanent-fetch-failure)).
+
+[`src/solana/SolanaEngine.ts`](https://github.com/EdgeApp/edge-currency-accountbased/blob/4795a57427d1925218987917abe1d767320a1245/src/solana/SolanaEngine.ts)
```
peachbits: This permalink pins to `4795a574`, the commit before the phase 2 fixes. The prose above describes `makeSolanaTransaction` and `isUnsupportedTransactionVersion`, neither of which exists at that sha, so following the link shows `processSolanaTransaction` and an unconditional `unavailable.add(signature)`. Worth repointing at the current head when the fingerprint is regenerated.

peachbits (changes requested):
The three findings from my last review are fixed and verified: `isUnsupportedTransactionVersion` keeps -32015 out of `unavailable`, `makeSolanaTransaction` makes the record atomic, and the new `Solana history watermark` block covers both. I reverted the -32015 guard locally and cases 6 and 10 failed as documented, so those are real regression tests. `tsc --noEmit` exits 0 and all 588 tests pass on 1.99.0.

One permanent-freeze path remains, on a thread that was resolved without the code changing.

src/solana/SolanaEngine.ts
j0ntz: Fixed. The raced result is cleaned with `asNumber` now, and the three answers are read apart: a number is the timestamp, a `null` drops the row without latching, and only a rejection (every node threw) sets `sawRetryableGap`. The `try` around the race also closes a path that predates this branch, where a lookup every node failed rejected out of `queryTransactionsInner` and ended the whole pass instead of one row. `asBlocktime` is deleted, since nothing else used it. Cases 12 to 14 cover the split; restoring the envelope cleaner fails 12 and 14.

src/docs/solana-v1-transactions.md
j0ntz: Repointed. The citation now pins the branch's first commit, which carries both phase 2 fixes and the code the prose describes. I used that commit rather than the head so the link survives the fixup churn while the review is open.

j0ntz:
Both findings are fixed. The remaining permanent-freeze path is closed: `getBlockTime` is cleaned as a bare number, a `null` answer drops the row without latching, and only a rejection freezes the watermark. The design doc's SolanaEngine citation is repointed at a commit that carries the phase 2 fixes.

j0ntz:
## 📸🪓 Test evidence

> 🪓 **Hack-forced evidence:** blocktime was hard-set to null in queryTransactionsInner so every row takes the getBlockTime fallback, and the plugin was pointed at devnet with the v1 address hardcoded; all four edits reverted, both worktrees clean. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: blocktime fallback history**

🪓 HACK-FORCED: blocktime fallback history

**🪓 HACK-FORCED: blocktime fallback balance**

🪓 HACK-FORCED: blocktime fallback balance

peachbits (approved):

src/solana/SolanaEngine.ts
```diff
@@ -568,41 +651,59 @@ export class SolanaEngine extends CurrencyEngine<
...
 
       // Process the transactions from oldest to newest
-      for (let i = 0; i < txResponse.length; i++) {
+      for (let i = 0; i < transactions.length; i++) {
         numProcessedTx++
-        if (txResponse[i].meta?.err != null) continue // ignore these
+        const tx = transactions[i]
+        if (tx == null) {
+          if (unavailable.has(transactionRequest[i])) sawRetryableGap = true
+          continue
```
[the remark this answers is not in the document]
j0ntz: Fixed in 2c299eb1: every permanent skip (-32015, a reverted transaction, a slot with no block time) now writes the row's signature into newestTxid unless a retryable gap already froze it, so a trailing unreadable row is not refetched. Case 13 covers -32015 as the newest row and case 17 the no-block-time row.

src/solana/SolanaEngine.ts
```diff
@@ -523,6 +542,70 @@ export class SolanaEngine extends CurrencyEngine<
...
+
+    try {
+      const funcs = this.tools.archiveConnections.map(
+        connection => async () => {
+          return await connection.getTransactions(signatures, options)
+        }
+      )
+      const transactions: Array<VersionedTransactionResponse | null> =
+        await asyncStaggeredRace(funcs)
+      return { transactions, unavailable: new Set() }
```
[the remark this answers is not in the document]
j0ntz: Fixed in 2c299eb1: each node's getTransaction retry throws on null, so the race asks the next node, and a row no node serves lands in unavailable and freezes the watermark. Batch rows that came back null go through the same retry. Cases 7, 8 and 14 cover it.

src/solana/SolanaEngine.ts
```diff
@@ -523,6 +542,70 @@ export class SolanaEngine extends CurrencyEngine<
...
+        const tx: VersionedTransactionResponse | null =
+          await asyncStaggeredRace(funcs)
+        transactions.push(tx)
+      } catch (error: unknown) {
+        this.error(
+          `getTransaction failed for ${signature}: `,
+          error instanceof Error ? error : new Error(String(error))
+        )
+        transactions.push(null)
+        if (!isUnsupportedTransactionVersion(error)) unavailable.add(signature)
```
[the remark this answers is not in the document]
j0ntz: Fixed in 2c299eb1: each node's retry returns -32015 as an answer instead of throwing it, so an outage on a later node cannot become the race's rejection and hide it. Case 9 runs a -32015 node followed by an unhealthy one.

PR #1099, opened 2026-09-09
Robinhood Chain: internal transactions from hosted Blockscout

j0ntz:
### Technical Design Document

[robinhood-chain.md](https://github.com/EdgeApp/edge-currency-accountbased/blob/jon/hood-stuck-sync/src/docs/robinhood-chain.md)

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

[EdgeApp/edge-react-gui#6208](https://github.com/EdgeApp/edge-react-gui/pull/6208) declares the new `blockscoutApiKey` in the app's EVM env config. Not a build dependency: this branch stands alone, and the key reaches the plugin through `env.json` either way.

### Description

QA reported every Robinhood Chain wallet stuck at exactly 50% sync on iOS and Android, on develop and staging.

50% is `makeTokenSyncTracker` arithmetic: a wallet with no enabled tokens splits its ratio evenly between balance and history. Balance completed; history never did. History never completed because every native-asset pass came back `partial`, and a partial pass is withheld from the synced ids. Every pass was partial because two sources failed in series:

- Alchemy rejects the `internal` category on Robinhood Chain (`-32602 The 'internal' category is not supported for this network`), so `includesInternal` is false and `mergeInternalTxs` runs on every native-asset sync.
- That merge asks the chain's own Blockscout instance, which had begun answering a Cloudflare managed challenge to the browser user agent the adapter sends.

**The frozen ratio is not the defect.** It is the engine reporting, accurately, that this chain's history is incomplete. Internal transfers are how ETH arrives through a contract call here, so a wallet that claimed 100% would be hiding missing money movements rather than a missing progress bar, and every other evmscan-backed chain withholds the pass on the same failure. The repair has to be a source that answers, not a reinterpreted ratio.

**`Pass the Blockscout challenge on Robinhood Chain`** sends the `sec-ch-ua` client hint alongside the browser user agent `BlockscoutAdapter` already sends, which that instance began requiring. It buys access and nothing more: the same instance now answers `x-ratelimit-limit: 10` per window rather than the 300 per minute it used to document, so one wallet's sync spends the whole budget in a single pass. A 43-minute device run carrying only this fix logged zero successful internal merges against 84 direct 429s.

**`Guard the partial-sync contract with a test`** pins the four behaviors of a partial pass in `test/ethereum/network/partialTxSync.test.ts`: it applies rows it has never seen, leaves a known confirmed transaction alone, keeps `lastTransactionQueryHeight` where it was, and withholds the asset from the synced ids. The symptom looks enough like a stall that the contract is worth pinning.

**`Add Blockscout's hosted API as a history source`** is the repair. `api.blockscout.com` serves chain 4663's `txlistinternal` under a key. It multiplexes its chains behind one origin, so its route names the chain as `/v2/api?chain_id=`, where Etherscan v2 spells it `chainid` and a single-chain instance names nothing; the choice among the three moves out of `fetchGetEtherscan` into `makeEvmScanUrl`, a pure function with its own test. The key is a new `blockscoutApiKey` init option rather than a value inside `evmScanApiKey`, because that option holds a chain's Etherscan keys and one is picked at random per request. A keyless request is answered 402, which the adapter classifies as a rate limit and would retry for seven seconds on every call, so the constructor drops that server when the option is absent and a build without the key behaves as it did before this branch. A chain that listed only the hosted API would be left with no server on such a build, so the constructor then nulls `fetchBlockheight`, `fetchTxs` and `fetchInternalTxs` the way `AlchemyAdapter` does. Review pinned what that buys: kept in, an empty server list answers zero rows, which `fetchInternalTxs` would mark `includesInternal`, and `fetchBlockheight` would throw on the `undefined` an empty waterfall resolves to. Dropped, the wallet still reports 100%, since `mergeInternalTxs` returns at its zero-adapter guard before the `partial` flag that withholds the asset is set. Such a chain would sync complete-looking on a keyless build either way, and the code comment and the design doc now say so.

`src/docs/robinhood-chain.md` carries the outage, the budgets on both servers, and the resulting engine contract as Phase 9.

### Testing

`~/.cursor/skills/verify-repo.sh . --base origin/master` passes: npm install, prepare, eslint on the changed files, and `npm run test`. New unit tests: `test/ethereum/networkAdapters/evmScanUrl.test.ts` covers all three server shapes, `test/ethereum/networkAdapters/blockscoutAdapter.test.ts` covers the keyless server drop and the adapter leaving the waterfall when no server is left, and `partialTxSync.test.ts` fails on its first case if a partial pass is made to count as synced, so it guards the contract this PR deliberately preserves.

Driven in the app on the iOS simulator against `edge-funds`, which holds three Robinhood Chain wallets, with this branch linked in through `updot` and the PRO key in the worktree `env.json` only. The engine log carries the merged pass for all three wallets, 60 times over a 12-minute window:

```
robinhood-I1: robinhood processEthereumNetworkUpdate tokenTxs robinhood-mainnet.g.alchemy.com + https://api.blockscout.com won
```

That `+ <server>` form is the successful internal merge: Alchemy's external rows folded with Blockscout's internal rows inside one pass. All three wallets then read 100%.

Internal rows reach the transaction list, not just the merge. `My Robinhood Chain 2` after a Resync shows a `Received Ethereum +Ξ 0.0114973` dated Jul 15 2026 3:41 AM. That transfer exists only in `txlistinternal` (block 10328689); the address has no incoming row in `txlist` at all, so the only path to that entry is the merged Blockscout rows.

The designed degradation was also exercised, unplanned: two wallets took a transient HTTP 500 from the hosted API on the first pass, logged `fetchInternalTxs failed, applying external rows only`, stayed partial with the query window open, and merged cleanly on the next pass a minute later.

---

> [!NOTE]
> **Medium Risk**
> Changes shared EVM scan URL construction and Blockscout adapter behavior used beyond Robinhood; incorrect key handling or partial-sync semantics could affect wallet history completeness or sync progress reporting.
> 
> **Overview**
> Fixes **Robinhood Chain wallets stuck at 50% sync** by giving internal transaction history a reliable Blockscout source and hardening how Blockscout requests are built and authenticated.
> 
> Adds a new **`blockscoutApiKey`** EVM init option (separate from `evmScanApiKey`) and wires **`api.blockscout.com`** into Robinhood’s Blockscout adapter alongside the chain instance. **`makeEvmScanUrl`** centralizes the three scan URL shapes (Etherscan v2 `chainid`, hosted Blockscout `/v2/api?chain_id=`, self-hosted `/api`), and **`BlockscoutAdapter`** drops the hosted server when no key is present (avoiding 402/retry loops) and sends **`sec-ch-ua`** with the browser user agent for Cloudflare on the chain’s instance.
> 
> **`EvmScanAdapter`** now uses `makeEvmScanUrl` for all scan fetches; fetch methods can be nulled so an adapter with no servers leaves the waterfall cleanly. New tests cover URL building, the keyless server drop, and the existing **partial** history-pass contract (withhold sync, keep query window open) in **`partialTxSync.test.ts`**. Docs and CHANGELOG record Phase 9 and the dependency on **`edge-react-gui`** for `ROBINHOOD_INIT.blockscoutApiKey`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf54b03d78340b5fbbefe51c8b5db8a75f406eef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


j0ntz:
## 📸🪓 Test evidence

> 🪓 **Hack-forced evidence:** appended the live wallet.syncRatio to the wallet name in WalletListCurrencyRow so the ratio is readable on screen; the ratios themselves are real, unforced engine output. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: robinhood stuck at 50 without fix**

🪓 HACK-FORCED: robinhood stuck at 50 without fix

**🪓 HACK-FORCED: robinhood reaches 100 with fix**

🪓 HACK-FORCED: robinhood reaches 100 with fix

j0ntz:
## 📸🪓 Test evidence

> 🪓 **Hack-forced evidence:** WalletListCurrencyRow was edited locally to append the live wallet.syncRatio to each wallet name as [sync N%], because the shipped UI renders the ratio only as an SVG ring with no readable value; reverted, git status clean. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: robinhood wallets 100 percent**

🪓 HACK-FORCED: robinhood wallets 100 percent

**internal only receive in history**

internal only receive in history

src/ethereum/networkAdapters/BlockscoutAdapter.ts
```diff
@@ -55,7 +76,35 @@ export class BlockscoutAdapter extends EvmScanAdapter<BlockscoutAdapterConfig> {
...
+        ...config,
+        servers: config.servers.filter(server => {
+          if (!server.includes(BLOCKSCOUT_PRO_HOST)) return true
+          ethEngine.warn(
+            `Blockscout server dropped: no blockscoutApiKey for ${server}`
+          )
+          return false
+        })
+      }
+    }
```
peachbits, line 106: **Warning** — when this filter empties `servers`, a missing internal-tx source becomes a silent success rather than a dropped adapter.

Robinhood keeps `robinhoodchain.blockscout.com`, so it does not happen today. But the new `BlockscoutAdapterConfig` doc presents `https://api.blockscout.com` as a server a chain can list, and a chain listing only that on a keyless build lands here with `servers: []`. Then:

- `qualifyNetworkAdapters` still keeps the adapter — it filters on `adapter[method] != null` (`EthereumNetwork.ts:656`), and `fetchInternalTxs` is still a function.
- `getAllTxsEthscan` hits its `this.config.servers.length === 0` HACK and returns `{ result: [] }` (`EvmScanAdapter.ts:377`).
- So `fetchInternalTxs` *resolves* with zero rows and `includesInternal: true`. `mergeInternalTxs` sets `tuple.includesInternal = true` and never sets `partial`, `lastTransactionQueryHeight` advances, and the wallet reports 100% synced with the internal transfers missing — the reading Phase 9 says was rejected the first time.
- `fetchBlockheight` has no such guard: `serialServers` → `asyncWaterfall([])` resolves `undefined`, and the destructure on line 110 throws a `TypeError`.

`AlchemyAdapter` covers exactly this at `AlchemyAdapter.ts:112-115`, nulling its methods when the list empties "instead of failing every sync with a confusing error". Doing the same for `fetchInternalTxs`/`fetchTxs`/`fetchBlockheight` (widened to `| null`) would make this the shape the comment claims it is. Class fields initialize right after `super()` returns, so a constructor-body assignment sticks, same as Alchemy's.

Worth a test as well: the keyless drop decides whether a build reaches the hosted API at all, and it is the only new logic in this PR without one.

test/ethereum/network/partialTxSync.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    walletLocalData: {
+      blockHeight: 400,
+      highestTxBlockHeight: 0,
+      lastTransactionDate: {},
+      lastTransactionQueryHeight: {}
+    }
+  }
+
+  const engine = {
+    ...state,
```
peachbits: **Suggestion** — `...state` copies the primitives by value, so `state.syncComplete` and `engine.syncComplete` are separate objects' fields from here on, while `walletLocalData`, `transactionList` and `txIdMap` stay shared references.

Every current test only mutates the objects, so it works. But a later test that sets `state.syncComplete = true` to reach the `addressSync` branch would silently have no effect and pass for the wrong reason — in a file whose whole job is to keep this contract from being answered wrongly again. A getter (`get syncComplete() { return state.syncComplete }`) or spreading only the object fields removes the trap.

peachbits (commented):
Checked out `jon/hood-stuck-sync`: `tsc` clean, `eslint` clean on all changed files, and the 9 new tests pass (585 vs 576 on the base). The one failing spec, `FTM Network Fees`, fails identically on `v4.91.0`, so it is not from this PR.

peachbits (changes requested):

src/ethereum/networkAdapters/BlockscoutAdapter.ts
j0ntz: Fixed in 3090aed1: when the keyless drop empties `servers`, the constructor now nulls `fetchBlockheight`, `fetchTxs` and `fetchInternalTxs`, the `AlchemyAdapter` shape, so the adapter leaves the waterfall instead of reporting an empty internal history as complete. That required widening `EvmScanAdapter`'s two fields to `| null`, and `test/ethereum/networkAdapters/blockscoutAdapter.test.ts` covers the drop, the empty-key shapes, the emptied list, the shared config staying untouched, and the keyed case.

test/ethereum/network/partialTxSync.test.ts
j0ntz: Fixed in 0fe20d5c: the fake engine now reads `syncComplete`, `transactionList`, `txIdMap` and `walletLocalData` through getters on `state`, so a test that sets `state.syncComplete` is changing what the engine sees.

j0ntz:
## 📸🪓 Test evidence (after fix)

> 🪓 **Hack-forced evidence:** Wallet rows locally append the live wallet.syncRatio as [sync N%] (both frames); frame 04 also lists only api.blockscout.com in robinhoodInfo and blanks blockscoutApiKey. All reverted. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: robinhood wallets 100 percent final head**

🪓 HACK-FORCED: robinhood wallets 100 percent final head

**🪓 HACK-FORCED: keyless hosted only adapter dropped 100 percent**

🪓 HACK-FORCED: keyless hosted only adapter dropped 100 percent

src/ethereum/networkAdapters/BlockscoutAdapter.ts
```diff
@@ -55,58 +76,104 @@ export class BlockscoutAdapter extends EvmScanAdapter<BlockscoutAdapterConfig> {
...
+    // With no server left the adapter leaves the waterfall entirely, the way
+    // `AlchemyAdapter` does. Kept in, `getAllTxsEthscan` would answer an empty
+    // server list with zero rows, so `fetchInternalTxs` would report a
+    // complete internal history it never fetched, and `fetchBlockheight`
+    // would throw on the `undefined` an empty waterfall resolves to.
+    if (this.config.servers.length === 0) {
+      this.fetchBlockheight = null
+      this.fetchInternalTxs = null
+      this.fetchTxs = null
+    }
```
peachbits, line 124: **Suggestion** — the comment overstates what leaving the waterfall buys.

The fix itself is right, and I confirmed the nulls stick: both the bogus `includesInternal: true` and the `fetchBlockheight` `TypeError` are gone.

But with the adapter dropped, `mergeInternalTxs` returns at its `qualifyNetworkAdapters('fetchInternalTxs').length === 0` guard (`EthereumNetwork.ts:383`) — *before* the `try`, so `tuple.partial` is never set. The pass counts as complete, `syncedTokenIds` gets `null`, and the wallet reports 100%. That is what evidence frame 04 is showing: not "internal history complete", but "no source configured, so nothing to withhold for". Same user-visible outcome as the path this block removes.

So what the drop actually prevents is the false `includesInternal: true` and the throw — not a complete-looking history. Worth saying that here and in the matching sentence in `robinhood-chain.md`.

Not reachable for Robinhood: `robinhoodchain.blockscout.com` never matches the filter, so `servers` never empties. Flagging it because the doc has that instance at zero successful merges over a 43-minute run, which makes deleting it from `robinhoodInfo` the obvious next cleanup — and the moment it goes, the chain is hosted-only and a keyless build silently claims a complete history. Whoever does that should know the drop does not cover it.

peachbits (approved):
Approving at `3090aed1`. Both earlier findings are properly fixed: the empty-server drop now nulls all three methods (and `blockscoutAdapter.test.ts` pins it, including the field-initializer ordering), and the fake engine reads through getters. `tsc` clean, `eslint` clean, 590 passing (up 5); the `FTM Network Fees` failure is the same one that fails on `v4.91.0`.

One comment-accuracy suggestion below — worth folding into a fixup, but not worth holding the merge for.

src/ethereum/networkAdapters/BlockscoutAdapter.ts
j0ntz: Fixed in d877db40, both here and in the matching `robinhood-chain.md` sentence. The comment now says the drop removes the false `includesInternal` claim and the `fetchBlockheight` throw, and states outright that the sync still reaches 100% because `mergeInternalTxs` returns at its zero-adapter guard before `partial` is ever set. The hosted-only case is called out in both places as one the drop does not cover.

j0ntz:
The comment-accuracy suggestion is folded into d877db40, which also re-stamps the design doc. Nothing else on the approval needed a change.

PR #1107, opened 2026-09-18
Fix Solana sends that land but report a failure

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

[Asana task](https://app.asana.com/0/0/1218612847313833/f)

QA saw every SOL and SPL send on staging fail with "Signature ... has expired: block height exceeded." Each of those transactions landed anyway. All three signatures in the QA screenshots are finalized on-chain with no error.

`solana.api.pocket.network` answers `getBlockHeight` at `confirmed` commitment with the slot, which is about 22 million above the real block height. `broadcastTx` ran `confirmTransaction` on every RPC node, and whichever node settled first decided the outcome, even when it settled with an error. On the Pocket connection, web3.js compared that slot to `lastValidBlockHeight` and threw its expiry error within a second, before Helius could confirm. The resubmission loop took the highest height any node reported, so it could end the send the same way. Pocket's gateway code notes that some of its Solana providers return the slot for this call (added in pokt-network/path#523). Production (accb 4.90.1) has the same code and node list. web3.js 1.98.4 fails the same way, so the 1.99.0 bump is not the cause.

Changes:

- Drop the Pocket node from Solana `rpcNodes`. Its other answers are also stale, and its websocket endpoint rejects connections.
- `broadcastTx` asks every node at each stage and takes the first success. A node's failure only counts once every node has failed or timed out. The timeouts are 10s per node for sends and status lookups, and 90s for confirmations, which is longer than the blockhash stays valid.
- The resubmission loop uses the lowest height any node reports. It only stops resubmitting and no longer fails the send.
- When every node fails to confirm, `getSignatureStatuses` with `searchTransactionHistory` checks every node for the transaction before a failure is reported.
- Any on-chain error fails the send. Before this, only string errors did, so an `InstructionError` object counted as success.
- `broadcastTx` throws when there are no RPC connections instead of hanging on an empty `promiseAny`.

After this, Solana's `rpcNodes` are `api.mainnet-beta.solana.com` (public and rate-limited) and Helius. This PR does not add a replacement node.

Testing:

- New `test/solana/solanaBroadcast.test.ts` has 7 cases. 5 of them fail against the old `broadcastTx`, including one that reproduces the QA error from a node reporting the slot as its height.
- `npm test`, eslint, and `tsc --noEmit` pass.

Not yet tested with a real send on a device. The 90s confirmation timeout has no unit test because the repo has no fake timers.

---

> [!NOTE]
> **Medium Risk**
> Changes Solana send confirmation and failure reporting on a critical user path; behavior is more conservative (fewer false failures, stricter on-chain error surfacing) but multi-node timing logic is complex.
> 
> **Overview**
> Fixes Solana sends that **landed on-chain** but the wallet reported as failed with **"block height exceeded"**, caused by racing RPC nodes (including one that returned slot as block height) and treating the first failing node as final.
> 
> **`broadcastTx`** now treats each stage (submit, confirm, status) as **multi-node with first success wins**: per-node timeouts (10s send/status, 90s confirm), resubmission driven by the **minimum** reported block height (so inflated heights no longer stop early), a **`getSignatureStatuses`** fallback when every `confirmTransaction` fails, proper failure when on-chain errors are **objects** (e.g. `InstructionError`), empty RPC list throws instead of hanging, and **`finally`** aborts background resubmit/confirm work.
> 
> Removes **`solana.api.pocket.network`** from default **`rpcNodes`**. Adds **`test/solana/solanaBroadcast.test.ts`** with seven scenarios covering slot-as-height, slow confirm, silent nodes, status fallback, real expiry, instruction errors, and no connections.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f42b2f150e16878a3b015081730909d1820026c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/solana/SolanaEngine.ts
```diff
@@ -1116,89 +1126,158 @@ export class SolanaEngine extends CurrencyEngine<
...
-        if (height < lastValidBlockHeight) {
-          await submitTx()
-        } else {
-          confirmationController.abort()
-          throw new Error('transaction expired')
-        }
+        if (height != null && height > lastValidBlockHeight) return
+        // A node already accepted the transaction, so a failed resubmission
+        // does not fail the send:
+        await submitTx().catch(() => {})
```
j0ntz, line 1190: **Suggestion:** this swallows every resubmission failure silently ([no-empty-catch](https://github.com/EdgeApp/edge-dev-agents/blob/main/.cursor/rules/typescript-standards.mdc)). Ignoring the rejection is right, but if all nodes start rejecting the resubmits (bad signature, rate limit) nothing reaches the logs. Logging at most once per broadcast keeps the 400ms loop from spamming:

```ts
await submitTx().catch((error: unknown) => {
  if (!loggedResubmitError) this.warn('Solana resubmit failed: ', error)
  loggedResubmitError = true
})
```

src/solana/SolanaEngine.ts
```diff
@@ -1116,89 +1126,158 @@ export class SolanaEngine extends CurrencyEngine<
...
-        // Confirmed!!
-        retryTxController.abort()
+        return status
       })
+      return await promiseAny(statusPromises)
+    }
+
+    try {
+      const txid = await submitTx()
+      retryTxSubmission().catch(() => {})
```
j0ntz, line 1244: **Suggestion:** `retryTxSubmission` cannot reject as written (`checkBlockheight` uses `allSettled`, `submitTx` is caught), so this empty handler exists only to satisfy `no-floating-promises`. A later edit that makes it throw would vanish here. `.catch((error: unknown) => this.warn('Solana resubmit loop failed: ', error))` costs nothing and keeps that visible.

test/solana/solanaBroadcast.test.ts
```diff
@@ -0,0 +1,181 @@
...
+  return node
+}
+
+const broadcast = async (nodes: FakeNode[]): Promise<EdgeTransaction> => {
+  const engine = {
+    networkInfo: { stakedConnectionRpcNodes: [] },
+    tools: { connections: nodes, makeConnections: () => [] },
+    warn: () => undefined
+  }
+  const edgeTransaction: any = {
```
j0ntz, line 84: **Suggestion:** `any` with no explanation ([no-any](https://github.com/EdgeApp/edge-dev-agents/blob/main/.cursor/rules/typescript-standards.mdc)). `Partial<EdgeTransaction>` cast at the call site, or a one-line comment saying why the fake skips the required fields, covers it.

j0ntz (commented):
Conventions pass only (no deep correctness review).

Commit subject `Confirm Solana sends on the first node that succeeds` is 52 characters; the limit is 50 ([commit-subject-50](https://github.com/EdgeApp/edge-dev-agents/blob/main/.cursor/rules/review-standards.mdc)). `Confirm Solana sends on first node to succeed` is 45.

src/solana/SolanaEngine.ts
```diff
@@ -1116,89 +1126,158 @@ export class SolanaEngine extends CurrencyEngine<
...
-          connection.getBlockHeight('confirmed'),
-          1000
+    // The lowest height any node reports, so a node reporting too high cannot
+    // end the resubmissions early. A lagging node only adds a few extra
+    // resubmissions.
+    const checkBlockheight = async (): Promise<number | undefined> => {
+      const blockheightResults = await Promise.allSettled(
+        rpcConnections.map(
+          async connection =>
+            await timeout(connection.getBlockHeight('confirmed'), 1000)
```
j0ntz, line 1169: **Suggestion:** now that the node timeouts are named constants (`NODE_REQUEST_TIMEOUT_MS`, `NODE_CONFIRM_TIMEOUT_MS`), this `1000` is the one left bare. It predates the PR, but a `NODE_HEIGHT_TIMEOUT_MS` next to the others, with a line on why it is so much tighter (it runs every ~400ms in the resubmit loop), keeps all three tunable in one place.

j0ntz (changes requested):

j0ntz (approved):

